PR #1, opened 2016-04-02
Separator dots in post numbers

rsaarelm:
You can't actually type a number longer than three digits in the
current CLI without separator dots, so better have them in the docs
as well.


galenwp:
Sorry, we left this open while we reorganized everything.  Now I think this doc has been completely worked.

PR #8, opened 2016-05-25
explain terminology, add details to setup instructions

asgardiator:
* Capitalization: "Urbit" for software, "urbit" for a planet/comet/etc
* Add clarity to some instructions (prefix shell commands with $, state 'ctrl-x' to avoid confusion with typing `^X` into REPL
* Tighten some language


galenwp:
This looks good to me, but I'd be curious to know @cgyarvin's opinion about the language changes in the beginning.

asgardiator:
Great. The language changes add flavor, but may be technically inaccurate. The flavor may also be excessive. In either case, improvement is welcomed.

galenwp:
Okay, I didn't look at this closely enough the first time around.  I think you're right that this language could use some improvement, but I'm not sure if this is working.  

Can I PR your branch?  Maybe that's the right way to do this, rather than going edit-by-edit.

asgardiator:
*clears throat* I also accept PRs.

PR #9, opened 2016-05-29
Nock interpreter in Groovy

kohanyirobert:
I've put together a simple interpreter for the sake of learning a bit of Nock and to understand Hoon eventually. I've added a new implementation page. The actual code has some tests in it, feel free to review it. Suggestions are more than welcome.

One question: can we link/import the actual code file from the origin repo directly? W/o copy-pasta? (I'm not sure if it's doable.)

Edit: I've noticed the `next: true` thingy in the doc files (and that it was missing from from `hoon.md`). Not sure what that is. Hope I've got it right.

### DIFF
--- a/docs/nock/implementations/groovy.md
+++ b/docs/nock/implementations/groovy.md
@@ -9,6 +9,17 @@ sort: 8
 From [Kohányi Róbert](https://github.com/kohanyirobert/gnock/blob/master/gnock.groovy):
 
 ```
+def i(def a) {
+  a.class in [
+    byte, Byte,
+    char, Character,
+    short, Short,
+    int, Integer,
+    long, Long,
+    BigInteger
+  ] && a >= 0
+}
+
 def n(def a) {
   if (a in List) {
     if (a.size() == 1) {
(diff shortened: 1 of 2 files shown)


galenwp:
This is great!

I wish we had a way to import the code directly from the repo.  Can think of a few ways to do that, but none of them are super easy.

The `next:true` thing just puts a little 'next' link at the bottom of the page.  It's part of the YAML that the `:tree` uses to render the page.  Looking at these pages, I honestly don't know why they're sorted the way they are.  Anyway, I think you want to put this in *before* hoon — so do you mind changing the `sort: 7` in `hoon.md` to `sort:8`?

kohanyirobert:
Yeah, I've kind of figured out the meaning of `next` and `sort` while I was reading through the docs today. I saw that you just updated the docs quite a bit. Good stuff. Very interesting. Anyway, I've updated the Markdown files, so `hoon.md` will be the last in line.

galenwp:
Thanks!  This looks good to me.

Glad you like the new docs.

PR #12, opened 2016-06-02
Nock interpreter in Scheme

kohanyirobert:
After writing a Nock interpreter in Groovy #9 I've decided to rewrite it in Scheme. Groovy does not handle tail recursion really well and is bit slow. I've rewritten the Groovy code to use continuation-passing style tail recursion, but it didn't help too much - `StackOverflowError`s all over the place. (I've updated `groovy.md` to reflect the changes.)

The Scheme version benefits from the continuation-passing style recursion. A deep recursion basically runs forever (it does pollutes the stack/heap - don't know which one - but much slower). Anyway, I didn't know too much about CPS before, so it was worth digging into it a bit. The code is a direct rewrite of the Groovy one, no novelty here. Also, I haven't used Scheme since reading the The Little Schemer ages ago, so it might look ugly :)

Sorry for only contributing only useless things, but it was fun.

*goes back figuring out how to implement decrement in nock*

### DIFF
--- a/docs/nock/implementations/scheme.md
+++ b/docs/nock/implementations/scheme.md
@@ -1,0 +1,128 @@
+---
+navhome: /docs
+title: Scheme
+sort: 8
+next: true
+---
+
+# Groovy 
+
+From [Kohányi Róbert](https://github.com/kohanyirobert/snock/blob/master/snock.ss):
+
+```
+(import (rnrs (6)))
+
+(define (i a) a)
+
+(define (n a r)
+  (cond
+    ((list? a)
+     (let ((l (length a)))
+       (cond
+         ((equal? l 0) (raise 1))
+         ((equal? l 1) (r (car a)))
+         (else
+           (let ((t (cond
+                      ((equal? l 2) (cadr a))
+                      (else (cdr a)))))
+             (n (car a)
+                (lambda (p) (n t
+                               (lambda (q) (r (cons p q)))))))))))
+    ((fixnum? a) (r a))
+    (else (raise 2))))
+
+(define (wut a)
+  (cond
+    ((fixnum? a) 1)
+    (else 0)))
+
+(define (lus a)
+  (cond
+    ((equal? (wut a) 0) (raise 3))
+    (else (+ 1 a))))
+
+(define (tis a)
+  (cond
+    ((equal? (wut a) 1) (raise 4))
+    ((equal? (car a) (cdr a)) 0)
+    (else 1)))
+
+(define (fas a r)
+  (cond
+    ((equal? (wut a) 1) (raise 5))
+    (else
+      (let ((h (car a))
+          (t (cdr a)))
+      (cond
+        ((not (fixnum? h)) (raise 6))
+        ((equal? h 0) (raise 7))
+        ((equal? h 1) (r t))
+        ((fixnum? t) (raise 8))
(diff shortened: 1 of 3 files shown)


docs/nock/implementations/scheme.md
```diff
@@ -0,0 +1,128 @@
+---
+navhome: /docs
+title: Scheme
+sort: 8
+next: true
+---
+
+# Groovy
```
galenwp: I think you mean `# Scheme`

galenwp:
@philipcmonk is our resident lisp guy — so I'd defer to him on the code.  I think you accidentally copied the Groovy title into the Scheme file.

philipcmonk:
I like it.  Once upon a time I wrote a nock interpreter in Racket.  Similar in a lot of ways, but yours seems cleaner.

https://github.com/philipcmonk/racketnock

kohanyirobert:
@philipcmonk I don't know if it's cleaner or not, but my multiple usage of `(list (list ... )` seems a bit off, however I don't know enough Scheme/LISP in general to know about a better way.

philipcmonk:
Yeah, mine does the same thing, except instead of operating on lists, I use cells (improper lists) at every level.  You need to create a lot of multilevel cell/list structures because nock is a combinator calculus.  I'm not completely sure how you seem to be able to do it with proper lists; does your version give proper lists as its result?  Unless the result is an atom, I mean.

kohanyirobert:
I've started off with Groovy which only have lists. I've wanted to be able to write cells without watching out for proper pair/bracket balance. So I've written a function ([I call it `n`](https://github.com/kohanyirobert/gnock/blob/master/gnock.groovy#L29)) which takes a value and creates a proper noun from it. If the value is an atom it's just returned, if it's a list it creates a new list with proper grouping/pairs ([here's the test case](https://github.com/kohanyirobert/gnock/blob/master/gnock.groovy#L268)).

As the Scheme code is just a rewrite of the Groovy code I've used the same idea. I wanted to be able to write S-expressions where I can omit non-essential brackets/parens. So I'm turning a proper list into an improper one with correct bracket/paren balance ([test case](https://github.com/kohanyirobert/snock/blob/master/snock.ss#L122)).

I'm [using the `n` function to transform the ugly `(list (list ... ` thing](https://github.com/kohanyirobert/snock/blob/master/snock.ss#L100) (as in `(tar (n (list ...`). I remember that I wanted to just use `(tar (cons ...` instead, but it didn't work for some reason ... or maybe I just lost track writing too many `(cons ...` and gave up. Right now it's ugly a bit, but I can still follow it ... kind of (next week from now I'll deny I wrote this :D).

philipcmonk:
Ah, I see it now.  `nock-eval` in mine is equivalent to your `tar` composed with `n`.  It's definitely easy to lose track of `cons`'s - it's always awkward to work against the grain of the language.

I'm not real sure why you ever need `(list` of a single item.  In nock `[a] == a`.  What breaks if you reduce that?

kohanyirobert:
>  In nock [a] == a. What breaks if you reduce that?

Just tried, nothing breaks. Yay!

> I'm not real sure why you ever need (list of a single item.

Artistic freedom I guess. Didn't really think about it. I can just raise an error when the input is `[a]`, but whatever, the code never gets there (unless in the test).

philipcmonk:
Cool, makes sense.

PR #14, opened 2016-06-03
adds js nock implementation

joemfb:
It had to be done ...

I've tested this pretty thoroughly, and I'm reasonably sure it's correct ;)

Let me know if you'd prefer it stripped down; it's obviously pretty long, and about half is the comments, module wrapper, exports, etc.

### DIFF
--- a/docs/nock/implementations/hoon.md
+++ b/docs/nock/implementations/hoon.md
@@ -1,7 +1,7 @@
 ---
 navhome: /docs
 title: Hoon
-sort: 9
+sort: 10
 ---
 
 # Hoon 
--- a/docs/nock/implementations/javascript.md
+++ b/docs/nock/implementations/javascript.md
@@ -1,0 +1,365 @@
+---
+navhome: /docs
+title: Javascript
+sort: 9
+next: true
+---
+
+# Javascript
+
+From [Joe Bryan](https://github.com/joemfb/nock.js/blob/master/nock.js):
+
+```js
+(function (self, factory) {
+  'use strict'
+
+  if (typeof define === 'function' && define.amd) {
+    define([], factory)
+  } else if (typeof module === 'object' && typeof module.exports === 'object') {
+    module.exports = factory()
+  } else {
+    self.nock = factory()
+  }
+}(this, function () {
+  'use strict'
+
+  /**
+   * Nock is a combinator interpreter on nouns. A noun is an atom or a cell.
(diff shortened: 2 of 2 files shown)


kohanyirobert:
I noticed your implementation, looks cool. I went through your readme file and I've tried and fed the reverse engineered Nock code for decrement to my interpreter

    [8 [8 [1 0] [1 6 [5 [1 0] 0 6] [0 0] 8 [1 0] 8 [1 6 [5 [0 30] 4 0 6] [0 6] 9 2 [0 2] [4 0 6] 0 7] 9 2 0 1] 0 1] 9 2 [0 4] [7 [0 3] 1 10] 0 11]

but it failed miserably, however this worked

    [10 [8 [1 0] 8 [1 6 [5 [0 7] 4 0 6] [0 6] 9 2 [0 2] [4 0 6] 0 7] 9 2 0 1]]

which is [another decrement formula](https://github.com/cgyarvin/urbit/blob/master/doc/book/1-nock.markdown).

Just to make sure I'm not completely inept I've tried the Clojure implementation as well and that failed too and then I noticed that this works

    [123456789 [8 [8 [1 0] [1 6 [5 [1 0] 0 6] [0 0] 8 [1 0] 8 [1 6 [5 [0 30] 4 0 6] [0 6] 9 2 [0 2] [4 0 6] 0 7] 9 2 0 1] 0 1] 9 2 [0 4] [7 [0 3] 1 10] 0 11]]

and returns 9, so basically the head of the cell is ignored.

I've wondered, how does the Nock code works for you "byitself"? (The first snippet.) Should it work?

I thought I messed up something in my implementation that's why I've experimented. Your code looks fine otherwise, so it should be fine, it's just weird.

joemfb:
Hey, thanks for digging in! I tried to make the `nock()` function flexible and easy, but it ended up more complicated ...

The default use case is obviously 2-ary, nock(subject, formula), but it supports two others:

- 1-ary, nock(formula), the `null` subject `[1, 0]` (Hoon `~`) is used
- n-ary (where n > 2), all the arguments are associated-right into a cell, the head is the subject, the tail is the formula

I wrote this up in the README, but it's a wall of text and it took me a minute to find it ... Probably easier to grok if it's structured as normal API documentation

galenwp:
This is great

kohanyirobert:
Yeah, I didn't read the whole readme, just the end :)

PR #25, opened 2016-06-07
minor Hoon twig corrections

joemfb:
- moves irregular `:cast` form example from `:like` -> `:cast`, corrects formatting
- adds irregular form examples for `:cons` (quoting rules seem confusing without examples), corrects formatting
- fixes broken link from `:call` -> `:door`

### DIFF
--- a/docs/hoon/twig/ket-cast/lus-like.md
+++ b/docs/hoon/twig/ket-cast/lus-like.md
@@ -16,11 +16,13 @@ within the span of `q`.  Otherwise, `nest-fail`.
 
 Regular: *2-fixed*.
 
-Irregular: `\`_foo\`bar` is `:like(foo bar)`.
-
 ## Examples
 
 ```
-~zod:dojo> :like('text' 97)
+~zod:dojo> :like('text' %a)
 'a'
+~zod:dojo> :like('text' 97)
+nest-fail
 ```
+
+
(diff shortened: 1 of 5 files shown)


docs/hoon/twig/ket-cast/lus-like.md
```diff
@@ -16,11 +16,13 @@ within the span of `q`.  Otherwise, `nest-fail`.
...
-
 ## Examples
 
 ```
-~zod:dojo> :like('text' 97)
+~zod:dojo> :like('text' %a)
 'a'
+~zod:dojo> :like('text' 97)
+nest-fail
 ```
```
joemfb: That example fails as written; not sure if that should be documented, or just removed...

galenwp: If it's broken we should probably take it out — but I'm not even sure what that was initially meant to do.

galenwp:
Unless someone can tell me what `:like` was for I'm gonna merge this...

joemfb:
`:like('text' %a)` produces `'a'`... I'll push an update removing the broken example

galenwp:
Thanks!

docs/hoon/twig/ket-cast/lus-like.md
cgyarvin: This error is actually caused by a minor flaw in the language.  `@t`97
works but ^-(@t 97) doesn't, because `` also has the magic power of casting
across auras.

Obviously this difference between regular and irregular syntax is not at
all the Hoon way -- this should be done better.

PR #33, opened 2016-06-08
Monday Madness

asgardiator:
1. New file:  `about/terminology.md`
  * a quick-reference guide to galaxy/star/planet/moon and suchlike
1. Throughout:
  * Move from `<blockquote>` (which does not parse Markdown) to `> ` (which does)
  * Tighten and disambiguate
  * (De)Capitalize
  * %s/ship/urbit/g


galenwp:
Is the idea behind `terminology.md` that it should become a glossary?

asgardiator:
Its purpose is vaguely "if you look at nothing else, look at this, and
please don't ask the same 5 questions in :talk"

asgardiator:
I'm 100% okay with axing terminology.md, and/or separating it into a new
pr.

docs/arvo/network.md
```diff
@@ -88,7 +86,7 @@ line).
 The second thing `++poke-noun` produces is our state.  `+>.$`
 refers to a particular address in our subject where our formal
 app state is stored.  It'll become clear why this is later on,
-but for now pretend that `+>.$` is a magic invocation that means
+but for now pretend that `+>.$` is a magic incantation that means
 "app state".
```
galenwp: I think 'invocation' is probably right.  More of a programming term?

galenwp:
Hm, hm.  My feeling is that we need a glossary, and your cosmic terminology could go in there.  @joemfb and I were discussing this in another thread.  There are plenty of words that need defining, and it should happen in one place (and be easy to link to).  It'd be even better to implement some kind of pop over.

The rest of your changes seem good to go (aside from 'incantation', which felt wrong to me).

asgardiator:
Hah! I was being silly. I'll commit it back to 'invocation' and we can
merge.

docs/arvo/network.md
philipcmonk: "incantation" actually feels a little better to me.  "invocation" implies "calling" in my mind, and `+>.$` isn't really doing that.  Not much of a difference, though.

galenwp: Oh now that I see the context maybe you're right.  You're the original author — this is really up to you!

joemfb:
I've got a glossary in progress - mostly Hoon so far. I can do an early PR (soon) and then keep working on it there, if you'd like. It's taking me some time to tease out the distinctions between concepts, and I'll have lots of questions ...

asgardiator:
Should I rename to `glossary.md`?

cgyarvin:
I'd approve of that.  I'll probably edit this PR but have a strong desire
to see it...

galenwp:
I saw some discussion on `:talk` discussion separating a glossary for concepts and a glossary for syntax.  That made sense to me.  Thought I'd add it here.

asgardiator:
Ah, great. I'll let the better glossary take its rightful place. That
should also tighten the focus of this PR as a whole.

On Wed, Jun 8, 2016 at 12:13 PM, Galen Wolfe-Pauly <notifications@github.com

joemfb:
Actually, I haven't covered any of the ship hierarchy. I think these glossaries should be combined.

asgardiator:
I'm up for that. We can discuss the glossary in your PR.  `cede glossary` was intended to separate the glossary question from this PR, which is more general language edits.

galenwp:
Seems ready to me.  Thanks!

PR #40, opened 2016-06-14
adds ;leave to :talk page

joemfb:
This seems to come up a fair amount in `:talk`...


galenwp:
It's true.  Good one.

Issue #21, opened 2016-06-04
Pandoc everything


galenwp:
We really should normalize our markdown formats.  In the past we have used [pandoc](http://pandoc.org/), which is great.

I guess this issue has two parts:
- Pandoc everything in the repo
- Create a `contributing.md` that explains the standard.

PR #41, opened 2016-06-14
Create contributing.md (& minor fixes)

asgardiator:
Part 1 of #21 
  * Create `contributing.md`
  * Link to twigs in `arvo/basic.md`
  * Hunt and eliminate stray comma


galenwp:
Sorry, I missed this one the other day.  I think that your contributing doc should at least link to https://github.com/urbit/urbit/blob/master/CONTRIBUTING.md.  

Or perhaps that doc should be duplicated in the docs?  It's annoying to maintain then together, and a repo should have a `CONTRIBUTING.md`.  What do you think?

asgardiator:
How's that?

galenwp:
Looks good.  Did you try rendering the switch away from `<blockquote class='blockquote'>`?  I hated taht style, but it was a bootstrap thing.  I think I added the needed CSS to drop the class.  Is that right?

asgardiator:
Yeah, you can see a rendered example already in
http://urbit.org/docs/arvo/basic/ so it seems to work just fine. Anyway,
<blockquote> seems to increase font size, which doesn't strike me as useful
or intended.

galenwp:
Yep, I did update that CSS.  Merged!  Thanks!

PR #43, opened 2016-06-15
fix #42 "aspirational" argument structure, examples

### LINKED ISSUE #42: Usage instructions out-of-date
Various usage instructions and examples under `using/filesystem` are now apparently out of date. For instance: `|merge %desk [~planet-name %desk]` now gives `nest-fail`, and only functions with `|merge %desk /~planet-name/desk=`; `|sync %desk ~planet-name` fails, requiring instead `|sync %desk ~planet-name %desk` (docs state the remote desk is optional if the same as the local desk); `|cp /path1 /path2 %init` fails, requiring no merge strategy argument.

ohAitch:


asgardiator:
Based on my conversations with @cgyarvin and @galenwp, "urbit" (lowercase u) is the new generic for "ship". Witness the latest few merged PRs where these changes were made.

cgyarvin:
"ship" is going to live.  It's just too natural.

docs/using/filesystem.md
```diff
@@ -108,28 +108,27 @@ Unmounts the path `%/web` from whatever name it was mounted as.
...
 
 ### `|sync desk plot [plot-desk]`
 
 Subscribe to continuous updates from remote `plot` on local `desk`.
-`plot-desk` can specify the remote `desk` name.  When omitted it's
-assumed to be the same as `desk`.  Moons have `|sync %home ~parent
+`plot-desk` can specify the remote `desk` name.  When omitted it is
+defaulted to being the same as `desk`. Ships have `|sync %home ~parent
 %kids` automatically set up (where `~parent` is the planet that issued
 the moon).
```
galenwp: 'where ~parent is the urbit that issued ~urbit' or similar.

galenwp:
The main thing is that the docs should be consistent.  @asgardiator is right that the goal was to use 'urbit' instead of 'ship' or 'planet' (unless you're referring to an urbit of a specific 'size').

In all the cases I see in this PR it seems like 'urbit' should be used instead of 'ship', for consistency with the rest of the docs.

ohAitch:
Done

PR #46, opened 2016-06-16
Fix failing network examples

alethiophile:
Hopefully this is correct. It matches what I've run against current urbit with success. #45


galenwp:
151-ization gone wrong.  Thanks!

PR #48, opened 2016-06-20
arvo exercises (formatting and broken examples)

joemfb:
Cleans up the formatting of the Arvo exercises, and fixes a couple of broken examples. Also fixes broken links in Arvo and elsewhere.

There are inconsistencies between relative and absolute links within the docs (there are both styles in this PR, as well). Do you have a preference between the two?

### DIFF
--- a/docs/arvo/basic.md
+++ b/docs/arvo/basic.md
@@ -29,8 +29,18 @@ at the [basic operation](/docs/using/admin) of your urbit.
 If you haven't copied in the examples repo, running the following
 commands from your urbit directory should do the trick:
 
-`$ cp -r {wherever-you've-pulled-to}/examples/gall/*/ arvo/`
-`$ cp -r {wherever-you've-pulled-to}/examples/dojo/*/ arvo/`
+Make sure you've mounted your `%home` desk:
+
+```
+~fintud-macrep:dojo> |mount %
+```
+
+Then, copy the example into it
+
+```
+$ cp -r {urbit-examples}/gall/*/ {your-pier}/home/
+$ cp -r {urbit-examples}/dojo/*/ {your-pier}/home/
+```
 
 Run an example to ensure it worked:
 
@@ -213,8 +223,8 @@ of the code. However, `b` is not an argument; `=|` sets `b` to the
 default value of whatever type it is declared as. Since the default
 value of an atom is `0`, b is set to `0`.
 
-> To produce the default value of any given type, use `$*` ('buctar')
-> followed by the type. Alternatively, use the irregular form `*`.
+> To produce the default value of any given type, use `*` ('tar')
+> followed by the type. This operaction is called *bunt*.
 >
 >         > *@ :: produce the default value of atom @
 >         0
--- a/docs/arvo/basic.md
+++ b/docs/arvo/basic.md
@@ -256,10 +256,10 @@ function call, then you have to specify every argument.
 
 **Exercises**:
 
-Tweak your code to complete the following excercises.
+Tweak your code to complete the following exercises.
 
 There are a few runes and some syntax that we have yet to cover that you
-will need to complete the excercises below. For these, please refer to
+will need to complete the exercises below. For these, please refer to
 our cheatsheat at the bottom.
 
 1.  Read and understand `++five` line by line.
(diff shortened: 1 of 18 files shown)


docs/arvo/basic.md
```diff
@@ -207,14 +217,14 @@ takes two children:
...
 (in other words, it declares a variable `b`) and executes the remainder
 of the code. However, `b` is not an argument; `=|` sets `b` to the
 default value of whatever type it is declared as. Since the default
 value of an atom is `0`, b is set to `0`.
 
-> To produce the default value of any given type, use `$*` ('buctar')
-> followed by the type. Alternatively, use the irregular form `*`.
+> To produce the default value of any given type, use `*` ('tar')
+> followed by the type. This operaction is called *bunt*.
 >
```
joemfb: It looks like this used to be the case, pre 151 (https://github.com/urbit/old-urbit.org/blob/master/bestiary/hoon/rune/irregular.markdown). It doesn't work now; is that by design?

docs/arvo/network.md
```diff
@@ -220,7 +221,7 @@ Let's walk through each of these elements step by step.
 If you look up `++bone` in `hoon.hoon`, you'll see that it's a number
 (`@ud`), and that it's an opaque reference to a duct. `++duct` in
 hoon.hoon is a list of `wire`s, where `++wire` is an alias for `++path`.
-`++path` is a list of `++span`s, which are ASCII text. Thus, a duct is a
+`++path` is a list of `++knot`s, which are ASCII text. Thus, a duct is a
```
joemfb: It looks like this has changed, per my reading of hoon.hoon

docs/arvo/subject.md
```diff
@@ -58,8 +58,8 @@ Hoon, we call the code executed the "formula" and its context the
       (add b $(b (add b 5)))    ::
                                 ::
     ++  sum                     ::
-      |=  a/@u                  ::
-      (add (five a) (three a))  ::
+      |=  {a/@u b/@u}           ::
+      (add (five a) (three b))  ::
```
joemfb: This is the solution to the previous exercise; not sure if it should be included here or not, but `sum` is called with two atoms...

Fang-:
The code on lines 23 and 24 of `docs/arvo/subject.md` should probably be removed, since they provide the answer to the chapter's first exercise.
(I'd make a PR myself, but figured it'd be neater to keep all these changes in one place, right?)

docs/arvo/basic.md
philipcmonk: As I understand it, `$*` is now (or always was?) `%$`, but that doesn't work either.  The irregular form is certainly `*`, so I agree with your change.

docs/arvo/network.md
philipcmonk: Correct.

philipcmonk:
@Fang- which exercise are you referring to?

docs/arvo/subject.md
philipcmonk: `++sum` should probably be just the single-argument version both here and where it's called.  I don't think there's any pedagogical reason to use the two-argument version?

philipcmonk:
I tend to prefer the absolute paths, but we should definitely make a decision, because we've been very inconsistent.  Maybe use relative paths if and only if it's a sibling or child? @galenwp @ohAitch ?

Fang-:
@philipcmonk "Modify :examples-sum to reset the counter when you poke it with 0."

ohAitch:
I prefer relative paths so that `|serve` doesn't affect things, or on the
flip side fully absolute ones with, like, urbit.org. (Optimally with /===,
but idk whether that will cause frontend problems)

cgyarvin:
Great PR!

I tend to prefer relative paths, simply because they provide fewer ways to
break things.  Does anyone have a reason to prefer absolute paths?

cgyarvin:
I'm not sure who should merge docs PRs, but I approve this PR.  Except that
"Galactc" in yu.md should be "Galactic."

(I would click the big green button, but my wrist has been slapped
repeatedly for doing so!)

philipcmonk:
@Fang- Do you mean docs/arvo/state.md?  If so, I believe you're correct.  Since Curtis merged this one, if you could go ahead and create another PR, that'd be great!

PR #49, opened 2016-06-21
Remove solution to exercise in Arvo's State chapter

Fang-:
As proposed in the discussion for PR https://github.com/urbit/docs/pull/48, removes the solution to the Arvo/State chapter's first exercise ("Modify :examples-sum to reset the counter when you poke it with 0.") from the example code given. The code in the docs now matches that in the examples repo.


philipcmonk:
Thanks!

Issue #30, opened 2016-06-07
several twigs have truncated explanations


joemfb:
- http://urbit.org/docs/hoon/twig/bar-core/cen-core/

> In this trivial example, `g` id ther ...
- http://urbit.org/docs/hoon/twig/cen-call/sig-open/

> In `:call` the core is a gate and we ...
- http://urbit.org/docs/hoon/twig/sig-hint/cen-fast/

> ... must already be registered.

galenwp:
This looks like it could be a mistake in porting some doc forward.  (Which I think we probably did programmatically).  

But, digging in, maybe not.  Here's an ancient version of `|%`: https://github.com/urbit/docs/blob/32c88d6ac989ed96dfeabf79a968686e33a7ed2f/docs/system/hoon/runes/br/brcn.md

galenwp:
Yeah, I thought I could find this by digging around in git history — but now I just have to ask @cgyarvin.  I have a vague memory of a conversation where he decided to rework some twig examples...

Fang-:
Descriptions are no longer truncated.

PR #55, opened 2016-06-23
adds rune-form twig expansions

joemfb:
As I read more of the kernel source, I've found the existing, keyword-based twig expansions to be confusing. I haven't memorized the keywords, and they're not heavily used anywhere, so I went ahead and translated the expansions into runes. This was a useful exercise for me, and I think it's a nice addition to the docs, but it does make them a bit busier...

Along the way, I found some issues with the expansions; I'll comment on those inline.


docs/hoon/twig/bar-core/hep-loop.md
```diff
@@ -12,10 +12,12 @@ sort: 4
...
-:core
-++  $  p
---
-``` 
+:trap  p
+```
+
+```
+=<($ |.(p))
+```
```
joemfb: I replaced the explicit core with trap, because of the rune description: "form a trap and kick it".

docs/hoon/twig/tis-flow/fas-var.md
```diff
@@ -11,7 +11,19 @@ noun with the subject.
...
+r
+```
+
+```
+?@  p
+  =+  p=q
+  r
+=+  ^-($=(p) q)
+r
 ```
```
joemfb: Note: the parenthetical nesting of the initial expansion is wrong. Also, this is a good example of the way these expansions can be confusing. `:coat(p)` isn't valid, but neither is `:coat(p q)`. I understand the actual expansions are macros on the AST ...

docs/hoon/twig/tis-flow/sem-rev.md
```diff
@@ -11,7 +11,19 @@ subject, inverted.
...
+q
+```
+
+```
+?@  p
+  =+  p=r
+  q
+=+  ^-($=(p) r)
+q
 ```
```
joemfb: same nesting problem here

docs/hoon/twig/wut-test/gal-deny.md
```diff
@@ -11,10 +11,14 @@ sort: 10
 
 ```
 :if  p
-  :zpzp
+  !!
 q
```
joemfb: Neither `:zpzp` nor `:fail` work as keywords for `!!`, so I replaced them all.

docs/hoon/twig/wut-test/bar-or.md
```diff
@@ -10,13 +10,25 @@ sort: 9
 ## Expands to
 
 ```
+:loop
 :ifno  p
```
joemfb: I added missing `:loop` twigs to several of the `?` expansions

galenwp:
This looks awesome.

docs/hoon/twig/wut-test/bar-or.md
ohAitch: I think the |- and ?~ are part of the macro code traversing the list of option twigs, and not the expanded output.

docs/hoon/twig/wut-test/hep-case.md
```diff
@@ -11,13 +11,23 @@ union, with no default.
...
+```
+|-
+?.  q
+  :lost
+?:  ?=(p.i.q p)
+  q.i.q
+$(q t.q)
+```
+
 ## Syntax
```
joemfb: `:lost` is another non-existent `twig` expansion; it looks to me like `[%lost]` is expanded in the AST, which signals the compiler to actually throw `mint.lost`, but there's no twig to signify that, and, therefore, no valid expansion

docs/hoon/twig/wut-test/bar-or.md
joemfb: At the very least, it's adjacent to the expansion: https://github.com/urbit/arvo/blob/master/arvo/hoon.hoon#L7319.

That raises the overall question: what purpose should these the expansions in the docs serve? I think, to the degree it's possible, they should be valid twigs with the actual behavior of the twig in question.

ohAitch: Sure: the specific issue I see is the conflation of e.g. runic `:ifno` and emitted `[%if ...]`, and I'm not sure what a non-confusing way to present that would be. The old style(see e.g. the `$fry` expansion comment nearby) uses `{}` to represent quasi-unquote, invoking interpolation syntax, but due to recent damage to the grammar that now means `$:`.

ohAitch: To be more concrete, `|(a b c)` doesn't expand to `|-(?~(p | ?:(a & $(xx something)))`, but rather into `?:(a & ?:(b & ?:(c & |)))`. The `:loop` is purely compile-time.

joemfb: Sorry, which `$fry` comment?

joemfb: Yeah, I understand. That same distinct is true of any twig that shows a list unwrapping with `i.p` and `t.p`. I guess I only took my objection so far ...

ohAitch: https://github.com/urbit/arvo/blob/master@{2016-6-22}/arvo/hoon.hoon#L7268-L7277

joemfb: I see, thanks

ohAitch: Or `$dip` right above it for a more complicated one involving a `$(q.gen t.q.gen)`.

joemfb: These changes adding `|-` to the `?` runes are inconsistent, but I don't think they're worse than what was there before. I puzzled over those expansions for a while, trying to figure out if `$` could actually recurse into a conditional ...

But maybe that's just me ;)

ohAitch: Okay yeah they're not, having the `$` without a referent is definitely
worse. Thank you for stumbling through our docs trying to bring light to
the gaps in them :3

docs/hoon/twig/cen-call/sig-open.md
```diff
@@ -13,6 +13,10 @@ sort: 3
 :rap(p :make(q +6 r))
 ```
 
+```
+=<(p %=(q +6 r))
+```
+
```
joemfb: This is another example of the ambiguity in the expansions. I puzzled over this one for quite a while.

I started with this

```
%~($ |=(a/@ +(a)) 2)
```

And then tried this

```
 =<($ %=(|=(a/@ +(a)) +6 2))
```

which doesn't compile, and can't even be typed in dojo. Eventually, I reread the twig description, so that `p` is a `wing`, and got this to work:

```
 =<($ =+(q=|=(a/@ +(a)) %=(q +6 2)))
```

I thought about making the case for that expansions, till I saw the list unwrapping in the `?` twigs ...

/cc @ohAitch

ohAitch: I think

=+  q=|=(a/@ +(a))
=<($ %=(q +6 2))

should also work?

joemfb: yeah, definitely.

My point is just that it wasn't clear to me, reading that expansion, that `q` needs to be a "reference" to an arm defined elsewhere, and it took a lot of trial and error to figure that out.

ohAitch: Or even

=+  r=2
=+  ^=  q
    |_  a/@
    ++  p  +(a)
    --

So that ~(p q r) is =<(p q(+< r)). Though idk if conflating pqr the subtwig
positions and pqr the arm, wing, and twig referents respectively is a good
idea.

cgyarvin: Yeah.  Basically, Joe, you're taking this exercise quite a bit more
seriously than I did.  That's good, of course, but...

The original idea was that the macro expansions in ++open would be
readable.  Unfortunately they're... somewhat readable.  They do work
perfectly.  Of course.

But these expansions in the doc were always pseudocode.  As pseudocode, I
was a bit sloppy and loose with them.  There used to be comments in the
hoon.hoon source that tried to establish a rigorous pseudocode for defining
Hoon macros, but there's no such thing as a rigorous pseudocode, eh?

Also, the undocumented runes you've discovered were undocumented on purpose
-- they're (generally) deprecated.  That's fine.  Not documenting something
is a pretty cheesy way of deprecating it.

You're obviously contributing on a pretty high level here, so I encourage
you to try and think hard about how exactly this should be done!  Just
filling you in on a bit of the history...

ohAitch: Okay no: it's misdocumented entirely, since ~($ |=(a/@ +(a)) 2) does in
fact work. What %~ actually expands to is %*(p q +< r), which does have the
=+ you mentioned in it.

joemfb: Thanks, that's helpful context. And, ironically enough, `%*` is, in fact, undocumented ;)

ohAitch: Ah yes, that was on the list. Well, it's =+ %= (or =+ %_ maybe?), with
limb-wrapping to make the expansion hygienic.

cgyarvin: And also deprecated :-)

But whatever.  Clearly we can't do these things in the usual "Curtis sits
in his basement and decides what to deprecate" way anymore...

joemfb:
The inline comment threads above make it clear that documenting the twig expansions is a *hard problem*. I'd like to make it an easy problem, by redefining it.

It seems that twigs could be classified in 3 ways: "core", "syntax sugar", and "macros". (I clearly don't have @cgyarvin's gift for naming things...)

- core twigs:  `=>`, `%=`, `|%`, `:-`, etc. The irreducibles
- syntax sugar twigs:  `=<`, `%-`, `|-`, `|=`, `:_`, etc.
- macros: `?&`, `?|`, `:~`, `;:`, etc.

"core" twigs are definitionally non-expandable, and not a problem. "syntax sugar" twigs are trivially (for some value of) expandable, and also not a problem, I think. It's those pesky "macro" twigs that aren't really expressable in valid syntax.

I propose that, assuming these distinctions are valid and well-formed, they be made at least somewhat clear in the docs. For instance, the "macro" twigs could have "Pseudo Expansions" or somesuch, in actual pseudocode or approximated in Hoon, but clearly differentiated from the "syntax sugar" twig expansions.

joemfb:
I've rebased this PR into two commits, attempting to split along those lines.

ohAitch:
For natural runes(i.e. core twigs), it might be informative to show
nock expansions: =+ is 8, ?: is 6, etc., though this seems a lot lower
priority than other possible doc.

Something like "non-finite-state reductions" definitely make sense like a
separate category, though I'd probably throw :~ into the "straightforward
sugar" pile, and %- as of 151 into "macro magic".

joemfb:
Of course, after I posted this, I read https://github.com/urbit/old-urbit.org/tree/master/doc/hoon/tut and saw that there were already names for `natural` and `synthetic` runes. I should've known no possible vector of Hoon concepts remained unnamed...

I think Nock expansions would be very cool.

`:~` expands to a pair-wise list reduction of the AST; it's straightforward, but not a literal expansion that can be copied and run. Doesn't `%-` still reduce to `%~($ p q)`?

And of course, my proposal leaves unresolved the question of how to actually document the macro magic...

ohAitch:
:~ is in fact non-finite state, but "~[1 2 3] expands to [1 2 3 ~]" is
a sufficient explanation in a way few others(though perhaps ;:) have.

%^(a b c d) reduces to %~($ a b c d) reduces to %*($ a ,.&1.+< b, ,.&2.+<
c, ,.|2.+> d): if you can link into macro-rune docs, it's fine, but the
path down to =+ %= has definitely stopped being the nonmagic =+(a $.-(+< [b
c d]:+)).

philipcmonk:
Is this ready to merge, then?

joemfb:
I don't think so. The distinction between natural and synthetic runes is already present in the docs (although it could be clearer). Does it make sense to draw a distinction between those synthetic runes that can be directly translated into valid Hoon expressions, and those which cannot (ie, the "macro" runes in my previous comment)? And is that something you'd like to move forward on?

My first commit covers the simpler synthetic runes; it just needs a fix for the `:open` (`%~`) expansion. And I don't know of a hygienic expansion that doesn't use deprecated runes...

The second commit adds rune-form expansions for the more advance synthetic runes ("macro"), but it doesn't distinguish them from the simpler forms in any way. That could be rectified with a custom heading or a short note after the expansion on each page - the greater obstacle remains in deciding how to expand them. Right now, they're basically AST transformations; clear enough with context, but not necessary helpful as explanations, particularly for new users.

I could split this into two PRs if you'd like to merge the simpler expansions first, or could continue working on it here.

joemfb:
I've made a pass at the advanced synthetic twigs, adding explicitly-marked pseudocode expansions, and separately titled Compiler Macro -- the best heading I could come up with... I've also added notes about macro-hygiene where the expansions appear to be non-hygienic.

And, on a personal note, it's pretty cool how much more sense these all make after 2 weeks ;)

cgyarvin:
The expansions are always hygienic!  Good lord!  "Young man, I don't know
what you're talking about.  My equations are correct."  I also note that he
or she who understands ;~, understands everything.

It's pretty cool how deep into Arvo you've gotten in 2 weeks.  Has it only
been 2 weeks?

Not to push my luck too far, but one thing that would be really neat is if
you can produce a small retrospective essay on  the general experience of
learning to program in Hoon/Urbit.  We'd love to put any such thing up on
the blog urbit.org (by anyone who wants to write one, not just Joe...)

cgyarvin:
Or at least, either hygienic or biblical :P

joemfb:
Well, I meant two weeks since I opened this PR ... ;)

I'm about a month into learning Hoon... I'll think about writing something up

docs/hoon/twig/wut-test/hep-case.md
joemfb: Github's not showing it here, but I replaced `:lost` with `~|(%mint-lost ...)` ...

ohAitch: It is a :lost though: it's only a problem if you try to compile it with a
non-void type, and ++open doesn't discriminate on what the input type is.

joemfb:
And the hygiene comments are in regards to the Hoon expansions in the docs -- I mean no ill will towards `++open`!

docs/hoon/twig/wut-test/hep-case.md
joemfb: What's a small lie among friends in search of a greater truth?

joemfb: ... ;)

I guess it depends on the purpose of the expansions; `:lost` is a syntax error, so it seems like a cruel kind of nerd-sniping to include it in the official docs. Does the distinction between it and a tracing printf matter to anyone who won't be reading the compiler source?

joemfb:
This PR is now handling several different issues:

- adding rune-syntax expansions for simple synthetic runes
- rounding out the truncated discussion section for `%~` (see #30)
- adding comments to explain that non-hygienic Hoon expansions are illustrative-only
- adding descriptive pseudocode "compiler macro" expansions in keyword and rune syntax for advanced synthetic twigs

Would it be easier to review as separate PRs? I think the first two are non-controversial, but the last two are more disruptive ...

philipcmonk:
I like all four points, honestly.  Do you have any more concerns before I merge this?

joemfb:
I'm comfortable with these changes, but every commit after the first includes some amount of editorializing on my part.

- Should the hygiene comments make it explicit that the expansions are *always* hygienic (or be re-worded in some other way)?
- Are the pseudocode expansions clear without being (too) wrong?
- Are you comfortable with my use of `~(%mint-lost ...)` in place of `:lost`?

cgyarvin:
I want to go over it, but I'm happy to do that *after* it's merged...

philipcmonk:
I don't think we need to emphasize that expansions are always hygienic -- one of our systemic problems has been pulling in random facts in the name of "completeness", when it's actually the wrong place for it.

cgyarvin:
I have merged the mega-PR with rune expansions!

I do feel this area needs a little revisiting in future -- there needs to be one format that's not pseudocody at all.  Even the compiler macros are still really pseudocode.  They're understandable, though.

I must insist on one point, however: sentence case in all headings :-)  It's modern!  It's fresh!

PR #57, opened 2016-06-24
adds explanation of :moar

### LINKED ISSUE #24: :moar should be more clearly documented
You can kind of figure out out based on usage, but it deserves its own entry, probably in the twig reference.

joemfb:
to the relevant twigs (`|=`, `|*`, `|-`, and `|.`)

closes #24


joemfb:
Not sure if the explanation is sufficiently clear; maybe it should say `:make($ ...)`? If the reader follows through to the actual `:make` documentation, it becomes clear.

PR #65, opened 2016-07-02
Provide actual ++hard source code.

Fang-:
`++cury`'s source code was being shown under `++hard`'s documentation. Replaced it with `++hard`'s actual source code.


galenwp:
Thanks!

PR #66, opened 2016-07-02
Fix: Several Documentation Issues

etscrivner:
**Overview**
In Hoon overview Syntax documentation, idiomatic rune FizzBuzz example:
- Actually use `end` face to determine program termination.
- Replace `text` with `pave` so output is identical to non-idiomatic rune example.

In Web usage documentation:
- Add note regarding file deletion bug in order to prevent confusion for new users attempting to follow documentation.
- Provide link to tree rendered version of `test.md` so users know where to go to see fully rendered result of JSX tag.

### DIFF
--- a/docs/hoon/syntax.md
+++ b/docs/hoon/syntax.md
@@ -159,7 +159,7 @@ with the [original](../demo).
   |=  end/atom
   =/  count  1
   |-  ^-  (list tape)
-  ?:  =(100 count)  ~
+  ?:  =(end count)  ~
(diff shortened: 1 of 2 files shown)


galenwp:
Change to web looks good to me.  @cgyarvin does that change in syntax work for you?

cgyarvin:
Yes.  I had a branch where ++pave got renamed ++text, but it didn't get synced.  Thanks Eric!

PR #68, opened 2016-07-07
adds stdlib docs for new gates (++find/++fand, ++spin/++spun)

joemfb:
Also corrects a couple of examples

- `++spin` and `++spun` added in urbit/arvo#176
- `++find` and `++fand` added in urbit/arvo#186

### DIFF
--- a/docs/hoon/library/2b.md
+++ b/docs/hoon/library/2b.md
@@ -882,6 +882,132 @@ Examples
 
 
 ***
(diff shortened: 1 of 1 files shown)


chc4:
Thank you very much for doing that! I completely forgot about documenting the arms, to be honest.

joemfb:
Sure, just let me know if I made an misrepresentations ;)

PR #69, opened 2016-07-10
subdirectory off by one error

juped:
bbtcat

### DIFF
--- a/docs/arvo/basic.md
+++ b/docs/arvo/basic.md
@@ -38,8 +38,8 @@ Make sure you've mounted your `%home` desk:
 Then, copy the example into it
 
 ```
-$ cp -r {urbit-examples}/gall/*/ {your-pier}/home/
-$ cp -r {urbit-examples}/dojo/*/ {your-pier}/home/
+$ cp -r {urbit-examples}/gall/*/*/ {your-pier}/home/
+$ cp -r {urbit-examples}/dojo/*/*/ {your-pier}/home/
 ```
 
 Run an example to ensure it worked:


joemfb:
I think this is wrong. I went through those exercises, using those exact commands

Ex:

- `dojo/euler1` contains `gen/`
- `gall/click/` contains `app/`, `mar/`, `web/`, etc.

philipcmonk:
@juped Care to defend your honor?

juped:
You want to copy the *contents* of the bf, euler1, etc. directories into the home desk, not those directories. I guess it’s not right that there’s a trailing slash; it should say examples/dojo/*/* in case we add any files that should be in the root of the home desk. But we’re not going to do that so it’s a minor correction.

ohAitch:
*is not at all surprised by trailing slashes in cp commands executed
verbatim doing different things on different systems*

joemfb:
Agreed that `/*/*` would also work, but it's an unnecessary change, at least on OS X:

```sh
$ mkdir ~/tmp
$ cp -r gall/*/ ~/tmp/
$ ls ~/tmp/
app       mar       readme.md sur       web
$ rm -r ~/tmp/
$ mkdir ~/tmp
$ cp -r gall/*/* ~/tmp/
$ ls ~/tmp/
app       mar       readme.md sur       web
```

joemfb:
And, yes, the trailing-slash-or-not behavior of `/bin/cp` is totally confusing

juped:
It’s absolutely a necessary change, because someone came in with the broken behavior.

cp -r is not standardized, is the main problem here. But everyone uses it. (cp -R is standardized, I think.)

joemfb:
As I understand it, the `-r`/`-R` distinction, where it exists, has only to do with copying or following symlinks. And on most implementations, they're synonyms.

In any case, the trailing slash is definitely wrong:

```sh
$ mkdir ~/tmp
$ cp -r gall/*/*/ ~/tmp/
$ ls ~/tmp/
examples pages
```

juped:
Terrible! Nice catch. There's a reason I use rsync, and being unable to remember these details is it.

Since it's the trailing slash behavior that's different between platforms, it has to be /*/* to be portable. BSDs and rsync use it to mean "the contents of this directory", but GNU has no way to specify that you want to copy the contents of a directory at all.

/*/* is broken and doesn't work, of course, because * doesn't glob to everything. Fortunately, the things it doesn't glob to are also things ignored by clay, so that should be fine to put in the instructions.

It should really be capital R, though. -r isn't even guaranteed to exist by any standard, and we should keep the awful pretense that there are standards in Unix up as long as possible...

I am completely without honor, of course, since I erred in favor of a GNU nonstandard behavior rather than a standard behavior. Only getting 7 GNU developers to renounce Stallman will redeem me...

joemfb:
"Do you renounce Stallman?"

> "I do renounce him."

"And all his works?"

> "I do renounce them."

joemfb:
To be honest, I almost always copy into tmp directories -- I'm never quite sure what's going to happen ...

PR #75, opened 2016-07-20
Remove extra @

artlogic:


galenwp:
Yes, thanks.  How did we let this sit for so long?  Sorry!

PR #81, opened 2016-07-25
Fix a couple typos.

whereswaldon:
Nothing too important, just trying to fix a few things that I mentally tripped over while reading the docs.


galenwp:
Thanks!

PR #82, opened 2016-07-25
Remove unimplemented ;who <station> command

whereswaldon:
I've tried to learn my way around :talk twice now, and both times this unimplemented command has tripped me up. This seems like our good friend "premature documentation." Can we remove it to spare future urbit pioneers the confusion?

### DIFF
--- a/docs/using/messaging.md
+++ b/docs/using/messaging.md
@@ -188,8 +188,6 @@ you're subscribed to.
 
 `;who` - list everyone in all your stations
 
-`;who station` - list everyone in that station
-
 ### Station Glyphs
 
 Glyphs are assigned by station hash out of the following list


galenwp:
`;who` with no `<station>` does work, right?

whereswaldon:
Not currently. I was about to check if that's an issue, actually.

galenwp:
Okay, it's definitely broken.  I don't understand what happened here.

ault011:
@whereswaldon did you try from a comet or a planet? I could swear it was working on my planet a few days ago, but it has since broken. I just tried it on my comet, and while I was able to type and enter the command, I did not see any output. Booting into another planet now to test.

Issue #50, opened 2016-06-22
Code in /arvo/network/ is different from urbit/examples code


knubie:
Specifically in the section "Sending a message to another urbit" the docs describe a "target" as an "urbit and app" with the code
```
[ost %poke /sending [to-urbit-address %pong] %atom 'howdy']
```

However in the [examples/pong.hoon code](https://github.com/urbit/examples/blob/master/gall/pong/app/examples/pong.hoon#L15) the line is

```
[ost %poke /sending [to dap] %atom 'howdy']
```

where `%pong` is replaced with `dap`.

PR #84, opened 2016-07-31
Provide twig examples in both keyword and rune form. 

hidwes-matped:
#50 #83


galenwp:
This is impressive.

galenwp:
Looks good to me.  Thanks for the work here.  As per the conversation on https://github.com/urbit/docs/issues/83 I think this is a worthwhile effort.  I'm also not convinced that the keywords are essential — but we really haven't adequately tested them.  The documentation still has a long way to improve.

PR #86, opened 2016-08-04
Key association clarification.

eaterjolly:
It should be clear from the get go what a person is actually given.
A person unfamiliar with crypto's limitations, might make bad assumptions.
___
There should probably be more said and more clarified elsewhere, but, for the mean time, this edit raises enough questions for the reader to make a prudent individual weary of making any assumptions. 
___
I realize the concept of a "will" may still be young, and a bit unclear with how it should be defined. 
But, it should still be brought up early, as from what I can tell it's quite important.

### DIFF
--- a/docs/about/glossary.md
+++ b/docs/about/glossary.md
@@ -15,7 +15,7 @@ to create a new semantic level in which one can be absolutely precise."
 
 ## Ships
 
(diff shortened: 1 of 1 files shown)


galenwp:
Sorry, we missed this somehow.

eaterjolly:
No, I appreciate the ability to contribute~!

PR #88, opened 2016-08-09
Document =~

hidwes-matped:
Don't know how to make this show up in the parent document lists. Set Sort: to 11 in the yaml. 
See #54.

### DIFF
--- a/docs/hoon/twig/tis-flow/sig-tow.md
+++ b/docs/hoon/twig/tis-flow/sig-tow.md
@@ -1,0 +1,50 @@
+---
+navhome: /docs
+sort: 12
+---
+
+# `:tow  =>  "tissig"` 
+
+`{$tow p/(list twig)}`: compose many twigs.
+
+## Produces
+
+The product of the chain composition. 
+
+## Syntax
+
+Regular: *running*.
+
+## Examples
+
+```
+~zod:dojo> =foo :new  n/@
+                :rap  :tow  increment
+                            increment
+                            increment
+                            n
+                      ==
+                :core
+                ++  increment
+                  :loop  
+                  ..increment(n +(n))
+                --
+~zod:dojo> foo
+3
+```
+
+```
+~zod:dojo> =foo =|  n/@ 
+                =<  =~  increment
+                        increment
+                        increment
+                        n
+                    ==
+                |%
+                ++  increment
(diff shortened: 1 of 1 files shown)


galenwp:
Annoyingly, `sort` can't overlap.  So you probably want to increment the `sort` field in https://github.com/urbit/docs/blob/master/docs/hoon/twig/tis-flow/tar-aka.md (and anything `> 11`) .

cgyarvin:
I wouldn't mind someone rethinking the sort field so that this kind of
manual number juggling wasn't needed...

On Tue, Aug 9, 2016 at 10:58 AM, Galen Wolfe-Pauly <notifications@github.com

galenwp:
I'm open to suggestions.  It's a little weird.  You could have two that overlap and just increment the rest or something.  But you end up with confusing data.  Even better would be for Urbit to notice the difference and actually just update the files.

cgyarvin:
I'd make the sort field a string, and have it sort alphabetically.  A
little weird but you can always make a gap.

Urbit updating the files would be a little weird and "Windowsy."

On Tue, Aug 9, 2016 at 11:02 AM, Galen Wolfe-Pauly <notifications@github.com

galenwp:
Alphabetical sort is interesting, but a bit counterintuitive.

cgyarvin:
So is Urbit!

You could also allow decimals... sort=11.5...

On Tue, Aug 9, 2016 at 11:08 AM, Galen Wolfe-Pauly <notifications@github.com

galenwp:
Could just use longer integers then. 100, 200, and so on. Then fill in the gaps. Still ugly though.

ohAitch:
Hm what is the alternative, a direct "next: " field?

galenwp:
I don’t mind that.

hidwes-matped:
So, wait: should I close this pr and resubmit with a `sort: 12` (which is clearly what I had meant to do in the first place....)?

ohAitch:
Just pushing another commit to the branch should be fine
The `next: ` proposal can go in a urbit/tree issue for future implementation.

hidwes-matped:
Amended.

galenwp:
https://github.com/urbit/tree/issues/28 for the YAML discussion

PR #89, opened 2016-08-10
edited small typo and fixed broken link

harporoeder:


philipcmonk:
Looks good to me (fixes: #90).  Our choice of using `/docs/using/shell/`-type urls (versus `/docs/using/shell` means relative urls will forever be annoying.

galenwp:
I think we can use relative paths.  The URL could be `../filesystem`, I think.

philipcmonk:
We can, yes.  It's just that HTML intended `filesystem` to mean sibling, not child.

galenwp:
I remember having between 1 and 20 arguments about this, and even reading the spec.  @ohAitch, why do we do it this way?

ohAitch:
Because it's a sibling; otherwise you'd have to put the current file name
in any links to children, from docs.md docs/foo docs/bar etc.

philipcmonk:
Sounds reasonable enough.  It's contrary to accepted practice, but that's never stopped us in the past.

ohAitch:
Accepted practice is that files and directories are different things, which
is the main mismatch here; we could do "everything is a file" instead, but
that is still somewhat confusing, and makes the links uglier.

galenwp:
I'm merging this, but the discussion can continue.

PR #100, opened 2016-08-26
fix error in %= and add runish

hoclun-rigsep:


joemfb:
This is a good catch!

I'd like to request a couple changes -- IMO, this is easier to read:

`:pin(foo :rap($ :make(+2 +6 bar)))`

and 

```
=+  foo
=<  $
%=(+2 +6  bar)
```

(I particularly dislike tall-form `%=` with only one change ... My $0.02.)

juped:
I dislike all uses of wide-form `%=`, for what that's worth. Tall or irregular are fine, but wide is weird because it's unbalanced.

hoclun-rigsep:
Everyone loves `=<` but I think in this expository role `=>` is better, showing a succession of subjects in order.

cgyarvin:
This is a funny error.  Good catch indeed!

I would not want to use %= with only one change in actual combat, as it were.  However, it is a nice way to show the tall form of the rune.  Again, joemfb's code is the right idiomatic way to write it, but the ~hoclun-rigsep way is the simplest explanation for a beginner.

hoclun-rigsep:
(As a beginner I feel I have some insight here...)

PR #103, opened 2016-08-28
Consistently use ctrl-key instead of ^key

Fang-:
As proposed in #36, all instances of key combo notations like `^X` have been replaced with `ctrl-x`.

Not entirely sure what to do about the output examples in Urbytes, ie:

> And we're done!  Press `ctrl-d` to exit:
```
~zod:dojo> ^D
$
```

I'd completely remove the `^D` there because it doesn't show up in the CLI either, but then what would those two lines be there for. Are they even necessary to begin with?

### DIFF
--- a/docs/about/runtime.md
+++ b/docs/about/runtime.md
@@ -513,7 +513,7 @@ no means the case.
 
 In deep memory, think of yourself as if in a signal handler.
 Your execution context is extremely fragile and may be terminated
-without warning or cleanup at any time (for instance, by ^C).
+without warning or cleanup at any time (for instance, by `ctrl-c`).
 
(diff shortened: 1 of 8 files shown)


galenwp:
I would just take out the `^D` there.  Actually, yeah, you can just take out that codeblock.

joemfb:
There's also a new `|exit` generator -- you might want to promote that instead...

Fang-:
I think we want to keep in the codeblocks then, since we're typing a command that produces output?

ohAitch:
^D is still how you exit non-immortal apps though; also, |exit is a bit of
a hack, though can probably be relied upon to continue existing

PR #105, opened 2016-09-26
Fix unfinished sentence

nickmooney:
Added a tautological reason to run Urbit -- could be replaced by a non-tautological reason


galenwp:
Could also just say 'is to have fun' or something.  But this is funny.  I'll take it!

PR #111, opened 2016-12-06
Mention the quiet flag in talk documentation.

Fang-:
What it says on the box. Quiet flag from urbit/arvo#300.

I'll update this to include the flag in urbit/arvo#308 if it gets accepted.

### DIFF
--- a/docs/using/messaging.md
+++ b/docs/using/messaging.md
@@ -192,6 +192,10 @@ a publicly readable publicly writeable station.
 You'll see presence notifications when people enter or leave stations
 you're subscribed to.
 
+`;set quiet` - Turn off presence notifications
+
+`;unset quiet` - Turn on presence notifications
+
 `;who` - list everyone in all your stations
 
 ### Station Glyphs


galenwp:
Ha, works for me.

ohAitch:
To me this is kind of starting to sound like we should
default-suppress presence notifications, and only use them to update
the web talk room occupancy display.

galenwp:
That's fine with me

cgyarvin:
100% agreed, let's try the patch

galenwp:
By 'try the patch', you mean accept this?  Because muting them would require a separate patch...

cgyarvin:
That's what I mean!  Sorry.

Fang-:
If nobody's on this yet I'll probably pick this up in the coming days. Do we prefer `quiet` to be on by default, or do we want to replace it with a `loud` option?

r0br0dz:
Quiet by default. Loud as an option similar to Verbose. Seems standard to
me and good. :)
Great project and vision btw

PR #113, opened 2017-01-05
Add python to apt-get dependencies

ohAitch:
This is redundant in most environments, but apparently commonmark will fail to build otherwise.


ohAitch:
(fedora aws etc should probably also have it, OS X comes with python iirc)

galenwp:
@ohAitch you're certainly welcome to merge this if you believe it needs to be merged?

ohAitch:
I emailed @juped and he pointed out that the Ubuntu python package is categorized as `important`, i.e. "provided by default, don't uninstall unless you know what you're doing". I'm guessing dply just provides a super stripped environment, which I'm not too bothered about supporting.

PR #114, opened 2017-01-27
think this should all be binzod and nod dozbud

heow:


galenwp:
Yep, think that's right.  This didn't get touched when we changed `@p`.  Thanks!

PR #115, opened 2017-01-27
line numbers are off, whitespace

heow:
whitespace usage really threw me, it's not mentioned until later in the hoon documentation


galenwp:
Great catches. Thank you!

PR #116, opened 2017-01-29
changed basic operation to administration

heow:
When I first clicked on "basic operation" I was surprised to find Unixy admin stuff.  Broke it up into administration: everything "outside" of Urbit and operation: everything "inside" Urbit.  

At least it makes sense to me, I won't feel bad this gets kicked back.

### DIFF
--- a/docs/using/admin.md
+++ b/docs/using/admin.md
@@ -2,10 +2,10 @@
 navhome: /docs
 next: true
 sort: 3
-title: Basic operation
+title: Basic Administration and Operation
 ---
 
-# Basic operation
+# Basic Administration
 
 <div class="row">
 <div class="col-md-8">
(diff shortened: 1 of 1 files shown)


galenwp:
This seems reasonable.  Give me a bit to think it over, and to try it out in a browser to see how it feels.

heow:
Title of "Basic Administration and Operation" is too long for the navigation, going to play with it a bit.

galenwp:
What about 'administration and operation'?  I agree that administration sounds like something you do from Unix.

heow:
"Administration and Operation" work well, it's not too wide on the docs/ screen but just a hair too wide inside admin/.  (see screenshots)

heow:
screen1
screen2

heow:
"subject oriented programming" has the same too-wide issue, both are readable

galenwp:
Doesn't look *too* terrible.  We could add some kind of fade on the side of the nav.  (It should expand when you mouse over it, by the way).

What browser is that?  The underlines seem to be rendering at the bottom of the line height.  Firefox?

heow:
Yes it's Firefox where it looks like a separator, Chrome renders as a proper underline.

galenwp:
Okay, I think the display bug is fixed.  Want to push a fix for that title change?

galenwp:
Thanks!

PR #120, opened 2017-02-05
Boot comet not fake ~zod in Urbytes.

keatondunsford:
Fake ~zods are kind of orthogonal to learning Hoon as a newbie, and having to checkout from Arvo master to the maintenance branch to get a working ship to boot is both cumbersome and confusing.

Switching the Urbytes instructions to boot comets makes setup as easy as `brew install urbit && urbit -c comet`. Plus, the devs coming to learn Hoon are people we want on comets hanging out in Talk in case they get stuck. So if they haven't joined already through the Quickstart, then it's just one `join ~star/urbit-meta` command away from getting in, not having to wait for another boot cycle. There's also a greater sense of identity in the comet model, as our current Talk underscore `~nocsur_soczod` notation is as memorable as planets outside of `our` and my-really-long-comet.urbit.org.
 
Really, the only downside is you lose that kind of badass feeling that you are coding as `~zod`, the ruler of the Urbit universe.

So either master needs to become stables and always remain that way (as I assume the Urbytes were written presuming a stable master), or this should probably be merged.

I updated Urbyte 0 to satisfy all of the above to give a flavor of what it might look like. If y'all think this is a good idea, I'll go ahead and update Urbytes 1 and 2 next and we can merge this.

I also tried to improve the onboarding to get people into Talk who could not be already and to subscribe to our mailing list (*hint*).

Also one small doc change I must have made a while ago but didn't PR.

### DIFF
--- a/docs/byte/0.md
+++ b/docs/byte/0.md
@@ -93,7 +93,7 @@ messages (you are booting an operating system, after all!).
 At the end, you'll see something like:
 
 ```
-ames: on localhost, UDP 58730.
+ames: on localhost, UDP 55360.
 http: live (insecure, public) on 8080
 http: live ("secure", public) on 8443
 http: live (insecure, loopback) on 12321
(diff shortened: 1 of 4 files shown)


galenwp:
@cgyarvin can you read this and sign off as copy editor?

keatondunsford:
Approve?

cgyarvin:
No, I need to see a clean commit which just contains the changes from the
original document.  Can you please ask Ray to show you how to do this?

PR #136, opened 2017-02-09
Change navhome path to /doc/. Fixes #127.

### LINKED ISSUE #127: Multiple links broken on root docs page when browsing back from other pages
To reproduce:

1. Load urbit.org/docs/. notice the link is in the top paragraphs are all correct. 

2. Browse to a sub-page (for example visit http://urbit.org/docs/using/install/)

3. Click the urbit-tilde in the top left corner to return to the main page (don't use the back button)

4. Notice all the links in the top paragraphs are now missing /docs and thus route to pages 
that don't exist.

keatondunsford:
Fixes a bug in our YAML where navigating home via our tilde icon in the top left corner of the docs wouldn’t put a `/` after urbit.org/docs. So clicking on a link like ‘Hoon syntax’ would direct to urbit.org/hoon/syntax not urbit.org/docs/hoon/syntax, causing an `Error: empty path`.

### DIFF
--- a/docs/about.md
+++ b/docs/about.md
@@ -1,5 +1,5 @@
 ---
-navhome: /docs
+navhome: /docs/
 sort: 1
 ---
 
--- a/docs/about/overview.md
+++ b/docs/about/overview.md
@@ -1,5 +1,5 @@
 ---
-navhome: /docs
+navhome: /docs/
 next: true
 sort: 1
 title: Technical overview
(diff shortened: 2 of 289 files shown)


keatondunsford:
@galenwp I tested this and made sure it was all good, and also searched via Atom to make sure there were no accidental `navhome: /docs//` or other typos. Should be all good!

galenwp:
It does appear to be referenced in the issue.  This looks good to me.  Will deploy today.  Thanks!

PR #142, opened 2017-02-10
Correct docs/arvo order. Fixes #129

### LINKED ISSUE #129: Markdown YAML order sorting incorrectly.
The reason the docs/arvo pages seem out of order (Security Drivers and API's coming before basic Hoon material) is because the YAML sorting is not functioning properly. Tree is sorting `11` and `12` before `2` and `3`. I tried fixing this by changing pages 0-9 to 00-09. This didn't work, but changing the page range from 1-12 to 11-22 fixed this.

Going to PR this so it works, but not sure if there's a change that can be made to make 1-12 work (just to make to make it normal, as I imagine this is going to be pretty annoying for people in the future wanting to, say, host their own blogs or static pages through Tree).

keatondunsford:
Let's at least just resolve this temporarily until we fix this YAML stuff in
Tree. I filed the same issue in Tree and we'll work it out as soon as we
can.

### DIFF
--- a/docs/arvo/api.md
+++ b/docs/arvo/api.md
@@ -1,7 +1,7 @@
 ---
 navhome: /docs/
 next: true
-sort: 12
+sort: 22
 title: API Connectors
 ---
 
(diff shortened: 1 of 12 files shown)


galenwp:
I really don't want to merge this, but we'll just fix it again later.

PR #147, opened 2017-02-14
Push to www

keatondunsford:
A few different commits grouped into a single push-to-www PR. They’re pretty clearly separated and commented on.

There are a few more things I can do, but this is a good place to start / push to www right now while I work on the other relatively minor fixes.

Thanks!

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -12,33 +12,6 @@ sort: 4
 
 <div class="col-md-12 head">
 
-<div class="col-md-4">
-
-<b>Community & contact</b>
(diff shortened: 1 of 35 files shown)


galenwp:
Okay, so this is ready — yes?

keatondunsford:
Good now unless you want me to add a `Community code` section for Fang's string processing library and Elliot's MUD.

galenwp:
I think we have more than that, so we'll collect and put that into a separate batch.

PR #148, opened 2017-02-15
Better Hoon doc ordering.

keatondunsford:
Think this is how the order was supposed to be originally but somehow
got messed up (likely sorting bug).


galenwp:
It looks to me like this was part of the last PR.  Is that correct?

keatondunsford:
Yes, oops. You can close this.

PR #154, opened 2017-03-22
Create test directory and files in sandbox instead of home since we'r…

jfranklin9000:
…e serving sandbox.

Please double-check.

### DIFF
--- a/docs/using/web.md
+++ b/docs/using/web.md
@@ -34,18 +34,18 @@ To switch the desk you're serving from:
 
     |serve %sandbox
 
-To host a file on the web try putting the following in `/home/web/test.md` (from Unix):
+To host a file on the web try putting the following in `/sandbox/web/test.md` (from Unix):
 
     # Hello
 
     This is a simple markdown file.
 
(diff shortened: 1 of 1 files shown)


joemfb:
FWIW, I kind of favor the opposite. It's worth documenting that you *can* switch the desk you're serving, but I don't think this should recommend that everyone do so. My $0.02 ...

jfranklin9000:
The previous page `http://urbit.org/~~/docs/using/filesystem/` Quickstart section creates a `%sandbox` desk and mounts it, then the Quickstart section of this page `http://urbit.org/~~/docs/using/web/` serves `%sandbox` but creates files in `home`.  It's inconsistent and confusing.  It needs clarification one way or another.

joemfb:
I absolutely agree with that. And I forgot the filesystem doc already has you creating and mounting `%sandbox`, so your changes seem like the right approach.

galenwp:
The idea was to avoid possibilities of creating merge conflicts. That's why we use %sandbox.

galenwp:
Anyway, thanks for catching this.  Looks correct to me.

PR #156, opened 2017-04-07
Fix typos

keatondunsford:
Collection of small typos I've caught recently.

### DIFF
--- a/docs/using/filesystem.md
+++ b/docs/using/filesystem.md
@@ -175,7 +175,7 @@ version, allowing merges to be rerun.
 the reverse of `%this`. This is different from `%init` because the new
 commit has both sides in its ancestry.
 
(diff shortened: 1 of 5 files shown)


docs/using/filesystem.md
```diff
@@ -175,7 +175,7 @@ version, allowing merges to be rerun.
 the reverse of `%this`. This is different from `%init` because the new
 commit has both sides in its ancestry.
 
-`%fine` - "fast-forward" merge. This succeeds iff one head is in the
+`%fine` - "fast-forward" merge. This succeeds if one head is in the
```
galenwp: I think this is intentional.  'Iff' means 'if and only iff': https://en.wikipedia.org/wiki/If_and_only_if#Origin_of_iff

PR #162, opened 2017-05-21
Update Ubuntu reqs

mcassano:
Debian doesn't provide libuv1, but Ubuntu does.  So I separated Debian and Ubuntu and added 'python libuv1-dev' which does result in a good build on Ubuntu.  The Debian is incomplete though, I don't know how to get libuv1 easily onto Debian.

### DIFF
--- a/docs/using/install.md
+++ b/docs/using/install.md
@@ -74,9 +74,12 @@ Which can usually be installed with the following one-liners:
     # Mac OS X [Macports]
     $ sudo port install gmp libsigsegv openssl autoconf automake cmake
 
-    # Ubuntu or Debian
+    # Debian (Incomplete, missing libuv1)
     $ sudo apt-get install libgmp3-dev libsigsegv-dev openssl libssl-dev libncurses5-dev make exuberant-ctags automake autoconf libtool g++ ragel cmake re2c libcurl4-gnutls-dev
 
+    # Ubuntu
+    $ sudo apt-get install libgmp3-dev libsigsegv-dev openssl libssl-dev libncurses5-dev make exuberant-ctags automake autoconf libtool g++ ragel cmake re2c libcurl4-gnutls-dev python libuv1-dev
+    
     # Fedora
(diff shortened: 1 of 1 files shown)


cgyarvin:
Hang on a sec, trying to get our Debian expert online...

juped:
If you look at line 50, these instructions are for building urbit 0.4.3 (and more generally the 0.4 series), which bundles libuv (https://github.com/urbit/urbit/tree/v0.4.3/outside/libuv-v1.7.5). So the dependencies are fine.

For the eventual 0.5 issue, Debian stretch (which has libuv1) has been in full freeze for a while and is likely to be promoted to stable before the 0.5 series is released; if it isn't we'll probably just have separate legacy debs for jessie. I think the libuv 0.10 on jessie doesn't work and you'd have to use the libuv1 out of backports. See https://github.com/urbit/debian for the package metadata.

Python is part of the default install on both these distributions.

galenwp:
Are the installation instructions for master documented anywhere?

Should we start a branch of the docs?

mcassano:
To provide more context, my steps to reproduce this issue are:
- Fresh Ubuntu 17.04 from Linode
- apt-get update
- apt-get dist-upgrade
- reboot
- download urbit-0.4.3.tar.gz, untar
- `make` and I got errors that Python and uv.h were missing

This was after I already determined that libuv 0.10 in Debian was insufficient.  So I do think these two additions of `python` and `libuv1-dev` are required for the 0.4 series but of course I could be mistaken.  Thanks for considering.

juped:
It might just be Python missing (breaking the libuv build). I think some Ubuntu server images remove it. Perhaps it's worth redundantly adding python because of this.

cgyarvin:
Thanks, Ray, can you file the appropriate bug?

Sorry for the trouble, Michael!

mcassano:
I installed python before libuv1-dev and still got the "missing uv.h" build error.  I think you have to explicitly install it.  I will reproduce with a 17.04 from ubuntu tomorrow.

More than happy to help, no problem.

juped:
Thanks. I'll spin up a VM on my end as well.

PR #163, opened 2017-05-29
Fixed a minor typo

joshuareagan:
Reversed a mistaken ordered pair from `[a b]` to `[b a]`; this was in the explanation of the subject passed to the loop arm.


galenwp:
Looks right.  Thanks!

PR #164, opened 2017-05-31
Arvo network message doc: Boot a comet, not a moon

keatondunsford:
A number of developers new to Urbit have been reading these docs and
running these examples, only to realize that booting a moon ships their
planets for unknown reasons. This is a simple doc fix that can push
people to comets in the meantime. And this is probably easier to do
than starting a moon anyways.

But maybe all of these docs should point people towards doing these on
fake ships? In an ideal Urbit world, dumb examples never sink your
planet, but unfortunately this can be the case now, and perhaps we
should make that explicit to early developer adopters.

### DIFF
--- a/docs/arvo/network.md
+++ b/docs/arvo/network.md
@@ -174,10 +174,10 @@ Run it with these commands:
     >=
 
 Replace `~sampel-sipnym` with another urbit. The easiest thing to do is
-to start a moon, a sub-identity of your urbit. If you don't know how to
-start a moon, see [the user admin section](/docs/using/admin/). Don't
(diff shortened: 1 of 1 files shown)


cgyarvin:
Good catch.  I'd definitely go with fake ships for *right* now...

Fang-:
Delving into fake ship setup in the arvo docs seems inappropriate, but might be useful to have in using/admin.md.  
For simple examples like this, I'd say there's nothing wrong with instructing the reader to boot a comet, while just referring to fake ship docs for the "proper & quick" way.

galenwp:
Comets seem fine.  I'll take this.

PR #165, opened 2017-06-04
++gte, ++gth, ++lte and ++lth produce %.y or %.n, which are loobeans

jfranklin9000:


Fang-:
Grepping through the docs, many other places (2b, 2f, 2h, arvo docs, and more.) seem to be using "boolean" where "loobean" is intended. This also raises the question: do we want to use "loobean" exclusively when referring to the Hoon type? Does hoon have boolean tests or loobean tests?

Not the biggest issue, but I feel consistency is an important part of good docs. Good catch @jfranklin9000 !

galenwp:
I honestly don't know what the right thing is.  It's still a 'truthy' value.

@belisarius222 or @cgyarvin want to weigh in here?

joshuareagan:
It seems like a good revision to me.  Loobean is the type, and anyway it's not a hard concept to learn.

belisarius222:
Hoon and nock only have loobeans. Vere deals with both loobeans and booleans, so in the C code it's especially important to be clear about this.

Other than the definition of a loobean (which probably bears repeating in multiple places in the docs), I'd say we should just replace all instances of boolean with loobean.

cgyarvin:
Agreed.

I hate the word “loobean,” though. Surely not too late to reconsider.

joshuareagan:
I too dislike the name.

Remember that floating "yes/no" character from the original Tron move, called "Bit"?  We could do "Bit".  [bit](https://www.youtube.com/watch?v=_fGujzulsas)  But I see we already have a ++bit in several cores of the standard Hoon lib...

"toggle"

"Georgian" (first name of Boole=George)

"Priestian"/"Grahamian" --- Graham Priest is a contrarian logician who believes that the law of non-contradiction is false, i.e., he believes that for some P, P&~P is true.  Inverting Booleans is contrarian...

I don't suppose we could just call them "Booleans" after all.  Would that be too dishonest?  Some people add and multiply these things...

cgyarvin:
But then there is the virtue of “loobean”: it explains itself...

joshuareagan:
Sort of.  I didn't recognize the backward "Boole" the first time I saw it.  Maybe it should be "Eloob" to be a true reversal.  "Eloobian" sounds about as gross as "loobian" though.

galenwp:
I’ve never really liked it. Too cute.

joshuareagan:
There is the generic and boring "truth value".  Gets the job done...

If we want a logician's name:  De Morgan, Venn, Frege, Russell, Lewis Carroll, CS Peirce (pronounced "purse"), Godel, Tarski, Martin Lob (variation on reverse Boole, pronounced "lobe"), Curry, Church, Carnap, Quine, Kripke, George Boolos (close to Boole but different)...

I think "Lobean" (after Martin Lob, pronounced "lobe-e-an") sounds a lot better, and [Lob's theorem](https://en.wikipedia.org/wiki/L%C3%B6b%27s_theorem) is actually pretty neat.

galenwp:
I love names like ‘truth value’.  It might be too long here.  Lobe is kind of nice.

joshuareagan:
"Loobean" looks like documentation-only terminology (no `++loob`), so the length of "truth value" doesn't seem like much of a strike to me.

With "Lobean" we don't have to change `$bean` in `++base`, and it's still clear what we mean.

Fang-:
In "loobean" vs "lobean", I'd choose the former. Wouldn't call it "too cute". "loob" associates more closely with "bool" (oo) than "lobe".

cgyarvin:
I don't see any consensus on this issue so I vote to leave it the way it is
now :-(

jfranklin9000:
`urboolean`

May be too close to `boolean` though.

cgyarvin:
Urbean?

joshuareagan:
Urbool or urbean, either one sounds better to me

cgyarvin:
I slightly favor urbean over urbool.

(Please note that urbean is not pronounced "er bean" or, dear god, "your
bean."  It rhymes with "lesbian.")

Fang-:
But then what are we going to call Urbit's denizens once Mars actually gets colonized? We need a contingency plan for that, you know. (^:

Future-proofing aside, urbean feels like a stronger departure from boolean than urbool does. I think that's a feeling we're shooting for, so +1. (But then, it's also less immediately apparent that we're just talking about simple booleans here...)

belisarius222:
We built our bike shed backward ... I would stick to either loobean or boolean. Of the two, I prefer boolean. Just because we have the convention of 0 as true and 1 as false doesn't prevent it from being considered boolean, and it has very few practical implications, unless you're deep in vere.

From [wikipedia](https://en.wikipedia.org/wiki/Boolean_data_type):

> Languages with no explicit Boolean data type, like C90 and Lisp, may still represent truth values by some other data type. Common Lisp uses an empty list for false, and any other value for true. C uses an integertype, where relational expressions like i > j and logical expressions connected by && and || are defined to have value 1 if true and 0 if false, whereas the test parts of if, while, for, etc., treat any non-zero value as true.[1][2] Indeed, a Boolean variable may be regarded (and implemented) as a numerical variable with one binary digit (bit), which can store only two values.

joshuareagan:
We shouldn't be shooting for a major departure from Booleans!  Hasn't the
standard response to complaints been "'1 is true, 0 is false' is just a
convention, our convention works just as well, it's nbd"?

My slight preference is the more transparent "urbool"; the type of data is
pretty clear to beginners.

cgyarvin:
The case of being "deep in vere" does come up a lot, though.  In general,
you'll see this issue becoming important in any kind of external
interaction.

cgyarvin:
I could support saying "urbool" and having a ++bool.

joshuareagan:
Boolean is fine too... It *is* a Boolean type, albeit with it's value flop
quirk.

joshuareagan:
Good point

cgyarvin:
I had a new idea on this subject: "bollean" and ++boll.  "Bollean" is close
enough to "boolean" that it's familiar, it sounds different enough to
distinguish the two, it doesn't sound horrible the way "loobean" does, and
"boll" is a word.  It also pays homage to the great German director, Uwe
Boll:

https://www.theguardian.com/film/2016/oct/28/uwe-boll-director-bloodrayne-alone-in-the-dark

But, this may be stupid.  Discuss.

ohAitch:
Soft namespace collision with ;~ go/nads terminology

keatondunsford:
I personally think these all sound stupid, but I don't have any better ideas off the top of my head. _Thinking_...

keatondunsford:
Though I'm leaning towards @belisarius222's stance -- why make this weird if it's still a boolean? "Booleans done the right way", or something.

belisarius222:
"bollean" looks like a typo to me. Now I do want to watch Alone in the
Dark, though.

joshuareagan:
Confusion at the vere level is the main reason to distinguish from
"boolean". Maybe the potential for confusion is a worthwhile cost to pay
for a convenient name, but that's a judgment call.

I'm okay with stupid-sounding names, but saying "loobean" makes me want to
shower afterward.

galenwp:
Agree that bollean looks like a typo.  I'm not sure what's wrong with truth value.

joemfb:
I agree with @belisarius222, they are *booleans*. I don't think the name has to fully capture both the semantics and the representation.

PR #169, opened 2017-06-19
Typo

juped:
erroneous hyphen->dash unification


Fang-:
Good catch!

juped:
I didn't catch it, someone reported it

PR #170, opened 2017-06-20
|sync examples had two on one line.

jfranklin9000:
This patch got added to another pr I sent earlier.  Here it is by itself.


Fang-:
Thanks!

PR #172, opened 2017-06-20
Clean up Glossary formatting and add table of Hoon/JavaScript analogies

cxk280:
The formatting changes are very minor (consistency of italics, for example). The table of Hoon/JavaScript analogies is in response to a suggestion of a "rosetta stone section" by @galenwp in issue #137.

I chose JavaScript because it is so widely used and is the language with which I am most familiar. The table is preliminary, and I will update it on an ongoing basis as I dive deeper into Hoon. There is a possibility my analogies may be incorrect at times--anyone may feel free to make corrections.


Fang-:
I'm not entirely sure how to feel about the JavaScript comparison table.  
The core and arm definitions feel kind of incorrect, just "object" and "variable/function of the object" might be a bit closer.  
I'd be inclined to call `|-` more of a GOTO, but JavaScript doesn't have those. 

Perhaps a better approach might be to take general programming terminology, rather than saying "this is like x in y language". Language-agnostic terminology is broader, and thus less likely to miss the mark.

But then, I'm someone who has already learnt these things and (some of) their nuances, so I might not be the best judge for this. I can definitely see it being a good starting point though!

cxk280:
Regarding the core and arm analogies, I based this off a specific characteristic of Map noted by [MDN](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Map):

> The keys of an Object are Strings and Symbols, whereas they can be any value for a Map, including functions, objects, and any primitive.

That is correct for Hoon, right? (I'm still learning.) If not, definitely happy to change it to simply Object.

As far as `|-` ('barhep') goes, I see your point but as you acknowledge GOTO doesn't exist in JavaScript. For what it's worth, note that the existing Hoon keyword for 'barhep' actually is `:loop`.

I see your point about language-agnostic terminology. One could do a separate list with using pseudocode, of course. No one has to use this table. But as someone learning Hoon who knows JavaScript fairly well, I know that the process would be much easier if a comprehensive list like this already existed.

docs/arvo/network.md
```diff
@@ -15,9 +15,9 @@ All we've written up until now are just shell commands that produce a
 value and then disappear. To listen for and receive messages from other
 planets, we'll need an app. Let's look at a very simple one:
 
-    ::  There is no love that is not an echo
+    ::  Accepts any noun from dojo and prints it out
     ::
-    ::::  /hoon/echo/examples/app
```
Fang-: This notation (reverse path, including extension) is actually a standard used by all official code. We'd like to keep that consistent, even across examples.

Fang-:
Another thing about using JavaScript-based examples, is that it restricts your audience to people who know JS intimately enough to understand the nuances of your comparisons.  
A comparison table is something that will no doubt be useful in the official docs, but it'll need to be discussed a bit more. Don't want to give people any wrong ideas to build off!

That said, all your other changes (aside from the one I just commented on) seem fine, and I'd like to pull them in! Can you separate the comparison table into its own PR, so we may continue discussion there? Thanks!

keatondunsford:
Thanks for the PR! At first glance, I think I agree with @Fang- here about the JS comparison; for some reason it doesn't feel right. I'll reply again soon with more detailed feedback. This is definitely something we need though in general, so keep toying around with stuff!

docs/arvo/network.md
cxk280: Got it. Will revert. Please note that it may not be apparent to the naive user, so do we want to mention this convention somewhere?

cxk280:
@Fang-, I have made the changes you requested, so it should all good to merge. I'll create a separate pull request to continue discussing the Hoon/JavaScript table.

Fang-:
Looks good. The `hoon_javascript_analogies.md` file is still there, once that's out I'll pull these in. Thanks! (:

cxk280:
Oops! There you go.

PR #173, opened 2017-06-21
Add information about ;what to messaging docs

ngzax:
Question came up in urbit-meta today. Seemed like a good addition.

This is ~hosneb-rosler / ~darlur, ftr. ;)


Fang-:
Heh, `;what's output could definitely use some prettier formatting.

Good to have it in the docs! Not to be nitpicky, but if you can add two spaces after the command itself (so the line after it actually line-breaks), then I'll merge this!

Issue #168, opened 2017-06-15
Latest release


Fang-:
The [installation instructions](https://github.com/urbit/docs/blob/master/docs/using/install.md) link to files with version numbers in their name. We could update these manually for every release, but either of the following solutions seem more robust to me.

* Link to `media.urbit.org/dist/xyz/latest.file` instead, and set up a small script to copy in the latest release from the [GitHub releases page](https://github.com/urbit/docs/blob/master/docs/using/install.md) whenever it gets added.
* Just plainly link to the GitHub releases page from the docs and/or refer to package managers. (I have heard rumors of `brew` containing a non-latest version of Urbit, we may want to automate releasing to package managers as well if at all possible.)

xykj61:
This has effectively been solved by electing to have people build from source for now. We're also working on install and upgrade scripts for Vere that will automate pulling the latest urbit/urbit master code and making the new binary. WIP.

ohAitch:
Oh, so we're no longer going to be uploading a `.deb`?

PR #175, opened 2017-06-22
link to 4.5 in install instructions

BorisVerk:
but someone should update the files on https://media.urbit.org/dist instead.

Issue https://github.com/urbit/docs/issues/168 is related.

### DIFF
--- a/docs/using/install.md
+++ b/docs/using/install.md
@@ -36,18 +36,18 @@ below.
 
 **FreeBSD**
 
-- Download `urbit-0.4.3.txz` [here](https://media.urbit.org/dist/freebsd/urbit-0.4.3.txz).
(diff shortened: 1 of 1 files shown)


Fang-:
As the issue you linked described, we don't like having GitHub file links in our docs for this. If you can change these to points to the expected media.urbit.org URLs instead, we'll make sure to put those files there and then merge this.

BorisVerk:
commit message has the wrong url, but the url in the patch is right.

ohAitch:
I don't see a freebsd .txz on https://github.com/urbit/urbit/releases

BorisVerk:
https://github.com/urbit/docs/commit/1f0096cbe4e24ca65a3778b832dd46d84aa745f2

PR #176, opened 2017-06-22
Add Table of Hoon/JavaScript Analogies to Glossary

cxk280:
Continued from #172.

@Fang-, I see a few different options right now:

1. Merge as-is
2. Put on separate page instead of main Glossary page, reached by a link along the lines of _If you know JavaScript, **this** may help_
3. I put it in a Github Gist
4. I put it on my personal site

Since I am of the position it is useful, I will at least do either 3 or 4. It is also worth keeping in mind that other languages are possible in the future.

Would love to hear opinion from @galenwp or even @cgyarvin.


cxk280:
One more point to consider: if the ultimate consensus is not to bring JavaScript into the official docs, what should be done about the multiple references to Lisp? As someone who does not know Lisp, these confuse me.

cgyarvin:
I'd say it's not so much a problem with referencing JS, as a problem with expressing exact parallels between Hoon and JS features.

If Hoon was shaped like JS, it probably should have been JS.  In some places it is shaped like JS.  In other places it isn't.  But there are so many basic parts of Hoon (like `:-`) where the comparison to a JS feature is just confusing, because there's no real match.

I think it is very nice to have a Hoon-JS comparison.  But I would either keep it vague, sporadic and inexact (as the various translations into "Lispese" are), or do it slightly differently.

I would do the comparison at a higher level, by presenting simple, purely functional computations in JS and their equivalents in Hoon.  The primary key of the table would be JS, not Hoon.  And the computations would be multiline tesuji, not language primitives.

The idea would be: here are a bunch of simple things you can do in Hoon, by translating from simple patterns and algorithms in a language you know well.  Everything in the table would have to be simple in Hoon, of course, and optimal not necessarily simple or pretty in JS.  You would probably write it by starting with some Hoon tesuji and translating them into JS.  Because every day is marketing day  ;-)

cxk280:
Great, that gives more clarity on how I can be helpful. Thanks for the thoughtful response, Curtis. :-)

PR #178, opened 2017-06-23
Make bulleted list for exercises in first two Urbytes

cxk280:


docs/byte/1.md
```diff
@@ -344,26 +344,24 @@ $
 
 ## Questions and/or exercises
 
-As always, these questions are optional and only for fun.
```
Fang-: I'd probably leave this bit in. People shouldn't feel like they're missing out if they can't/don't want to take the time for these.

Fang-:
Looks good! Two nitpicky things though. First, the comment I left above. Second, there's a paragraph above the bullet list of questions:

> Here's a question to think about.

Now that it's not a """single""" question anymore, that should be rephrased.

Aside from those points, I'm happy to merge these changes in. (:

docs/byte/1.md
cxk280: I'll put it back. Please note that I still disagree, though, as I don't really see the questions scaring someone away who read the entire page, and any opportunity to trim down the number of words will probably be useful.

cxk280:
Both points are resolved.

Fang-:
Cool, thanks again!

PR #179, opened 2017-07-07
Fixes #26: Adds proper anchor tags to sections for linking

### LINKED ISSUE #26: Hoon concepts page has broken links
`mint`, `twig`, `span`, and `limb` have broken links on the Hoon concepts page (http://urbit.org/docs/hoon/concepts/). It's not clear where those links should go.

ngzax:
Links now work for mint, twig, span, and limb.

Ran this through pandoc per the contributing docs, but not sure if that was the best idea since it makes the diff look much bigger than it really is.

### DIFF
--- a/docs/hoon/concepts.md
+++ b/docs/hoon/concepts.md
@@ -1,229 +1,236 @@
 ---
-navhome: '/docs/'
-next: True
+navhome: /docs/
 sort: 13
+next: true
 title: Concepts
 ---
 
 # Concepts and terminology
 
-Usually, a new language is an improvement on one you already know. If it
-isn't, it probably at least shares concepts with a large family of
-relatives. These concepts come with words, like *type* or *function*,
-that map cleanly onto its semantics.
+Usually, a new language is an improvement on one you already
+know.  If it isn't, it probably at least shares concepts with a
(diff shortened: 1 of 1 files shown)


Fang-:
Wow yeah, pandoc sure made a mess out of the diff here. I'd be perfectly fine with merging this without pandoc's changes sprinkled all over though! If you can amend your commit to not include all that noise then I'll pull this in for you. Thanks! (:

ngzax:
@Fang- : Thanks, updated. With the help of %clay (actually pretty interesting, I am going to write this up) I was easily able to get back the version before pandoc. 

I wonder if we should update the /about/contributing/md to mention that you should use pandoc to *intially* format the doc but that small edits shouldn't require it. 

Also, the note about "Pandoc will convert the terminating --- in YAML metadata to ..." appears to be out of date. I have pandoc 1.19.2.1 (latest installed via brew) and it didn't do it.

Fang-:
Cool, thank you!

A PR to the contributing.md would be welcome, even if just as a platform to discuss the pandoc requirements.

ngzax:
Sorry for the delay, but the PR for contributing.md is here: https://github.com/urbit/docs/pull/186

PR #183, opened 2017-07-13
Adjusting Description of dot-tar

joshuareagan:
There is a sense in which `.*` is like Nock 2, but I think that's misleading.  `p` and `q` are evaluated against the Hoon subject using Hoon rules, and then the latter product is evaluated against the former using Nock rules.  This isn't really Nock 2, which is, of course, entirely in Nock.  The current description confused me when I was first learning Nock.

This is a nitpick, but I think we can, and should, be a little more clear here.

### DIFF
--- a/docs/hoon/twig/dot-nock/tar-nock.md
+++ b/docs/hoon/twig/dot-nock/tar-nock.md
@@ -5,7 +5,7 @@ sort: 1
 
 # `:nock  .*  "dottar"`
 
-`{$nock p/seed q/seed}`: evaluate with Nock `2`.
+`{$nock p/seed q/seed}`: Nock evaluation.
 
 ## Produces
 
@@ -16,6 +16,10 @@ Nock of formula `q` and subject `p`, with span `%noun`.
 Regular: *2-fixed*.
 
 ## Discussion
+
+The rune `.*` evaluates twigs `p` and `q` against the 
+Hoon subject, and then evaluates in Nock the latter 
+product against the former.
 
 Note that `:nock` can be used to bypass the type system,
 though its product contains no type information.  It's


joshuareagan:
Am I off base here?  Flaming is fine---the burns help me learn.

Fang-:
Hey, sorry for taking so long to get back to you!

I asked around, and this ("evaluates twigs, and then evaluates in nock") may be a bit too pedantic. The twigs themselves are also evaluated in nock, they're just compiled first.  
I personally think the specification of "evaluate latter against former" does add a little bit of clarity to the docs, even though that is the default mode of operation.

joshuareagan:
Fair.  There are probably better ways of describing it.

I think the main thing that bothers me is calling this "Nock 2".  This is straightforwardly false, for the simple reason that Nock 2 doesn't take twigs, and `.*` does.  Would it be better if I changed the title only, and left everything else alone?

Fang-:
Do you feel the same about all the other dot-runes? Nock doesn't know about seeds either, yet `.=` accepts them while claiming to be Nock `5`.

joshuareagan:
It's a good question.  The description "test for equality with Nock 5" seems accurate for `.=`, though.  Yes, the twigs are evaluated before the equality test is applied, but surely Nock 5 is how that test is implemented.  Nock 5 isn't the *only* thing that's happening with `.=`, but it's pretty clear how Nock 5 is being applied.

The same isn't true of `.*`.  It's not that the twigs are evaluated and *then* Nock 2 is applied.  We're saying the twig evaluation is part of the implementation of Nock 2.  I wouldn't make a fuss except this caused me confusion as I was first learning it, and I spent a non-trivial amount of time figuring out what was happening.

joshuareagan:
wait I take it back

Fang-:
Hah, are you cool with me closing this PR then?

joshuareagan:
I'm actually leaning toward arguing we should revise both, but I'm thinking it over.

Fang-:
(To outsiders looking in, closing was agreed on by @joshuareagan in talk.)

PR #185, opened 2017-07-20
expand/explain :dip

joemfb:
hopefully this is clarifying ...

### DIFF
--- a/docs/hoon/twig/sem-make/sig-dip.md
+++ b/docs/hoon/twig/sem-make/sig-dip.md
@@ -56,6 +56,11 @@ so you don't need to either.
 
 `:dip` is often used in parsers, but is not only for parsers.
 
+This can be thought of as user-defined function composition; instead of simply
+nesting the gates in `q`, each is passed individually to `p` with the product
+of the previous gate, allowing arbitrary filtering, transformation, or
+conditional application.
+
 ## Syntax
 
 Regular: *1-fixed*, then *running*.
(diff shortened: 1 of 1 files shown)


galenwp:
Okay, this has been around for a while.  @cgyarvin can you comment on this?

Fang-:
I am not he who has been summoned, but I can weigh in since this discussion has come up a few times before.

>This can be thought of as user-defined function composition

That's a nice and concise wording of the conclusion we usually arrived at. The first function acting as "glue" between the rest.  
It may or may not be nice to mention a parser example specifically, since `;~` is used with those so often (and the current example code isn't super informative), but I'd call this addition a-okay as is.

cgyarvin:
Yeah this is good.

Pipes being glued together is the right metaphor.  There are two core
models: a pipe and a joint (ie glue).  The joint gate gets a pipe product
and another pipe, with its sample preset to the first pipe's input.  Since
the joint combining two pipes has direct access to the first pipe's input,
no information is lost.  Hence you can have a parser combinator like ++pose.

We are almost certainly under-using this language feature all over the
place.  It is a little powerful for my taste.  Too Haskellish.  But when
you need it, as in parser combinators, you really need it.

joshuareagan:
FWIW the changes Joe proposes make the rune a lot clearer to me.

ohAitch:
+1 on the changes

While we're on ;~, for examples I'd use ++biff instead of a custom
cord->tape thing, since in many ways ++unit is the minimum viable
monad; but that's probably a separate task/PR

keatondunsford:
Yeah, this looks fine for now. Thanks Joe!

PR #187, opened 2017-08-02
Adds an Index page for the Standard Library. Fixes #98.

### LINKED ISSUE #98: No anchor links or index for standard library
Finding the documentation for a given arm requires one to grep the source or else guess which section it's in. Also can't link to anything from other parts of the docs.

ngzax:
This is a **Proposal** and is not ready for merge. I wanted to get this out there, however, before I put in more work to make sure I am going down the right path.

This is a big "itch" of mine because when you read examples, for example in this excellent post: https://urbit.org/fora/posts/~2017.7.20..09.13.38..a812~/, you see gates like cork, crip, and trip and it is difficult to find them in the current 1a, 1b, &c. structure of the standard library docs.

This is unfortunately hampered by the fact that *[tree](https://github.com/urbit/tree)* doesn't properly follow anchor links when sending to another page. (They do work intra-page). This is an existing bug that is already affecting the docs and I am going to try to take a stab at fixing that as well. But, for now, for this patch you'll just have to imagine it works. :wink: At least it takes you to the right section.

Comments on the formatting and presentation are welcomed.

### DIFF
--- a/docs/hoon/library/1a.md
+++ b/docs/hoon/library/1a.md
@@ -4,7 +4,7 @@ navhome: /docs/
 
 
 # 1a: basic arithmetic
-### `++add`
+### <a name="add">`++add`</a>
 
 Add
 
@@ -43,7 +43,7 @@ Examples
     1.337
 
 ***
-### `++dec`
+### <a name="dec">`++dec`</a>
 
 Decrement
 
@@ -84,7 +84,7 @@ Examples
 
 
 ***
-### `++div`
+### <a name="div">`++div`</a>
 
 Divide
 
@@ -129,7 +129,7 @@ Examples
 
 
 ***
-### `++gte`
+### <a name="gte">`++gte`</a>
 
 Greater-than/equal
 
@@ -170,7 +170,7 @@ Examples
 
 
 ***
-### `++gth`
+### <a name="gth">`++gth`</a>
 
 Greater-than
 
@@ -210,7 +210,7 @@ Examples
 
 
 ***
-### `++lte`
+### <a name="lte">`++lte`</a>
 
 Less-than/equal
 
@@ -253,7 +253,7 @@ Examples
 
 
 ***
-### `++lth`
+### <a name="lth">`++lth`</a>
 
 Less-than
 
@@ -299,7 +299,7 @@ Examples
 
 
 ***
-### `++max`
+### <a name="max">`++max`</a>
 
 Maximum
 
@@ -340,7 +340,7 @@ Examples
 
 
 ***
-### `++min`
+### <a name="min">`++min`</a>
 
 Minimum
 
@@ -367,7 +367,7 @@ Source
       ^-  @
       ?:  (lth a b)  a
       b
-        
+
 Examples
 --------
 
@@ -381,7 +381,7 @@ Examples
 
 
 ***
-### `++mod`
+### <a name="mod">`++mod`</a>
 
 Modulus
 
@@ -408,7 +408,7 @@ Source
       ^-  @
       ?<  =(0 b)
       (sub a (mul b (div a b)))
-        
+
 
 Examples
 --------
@@ -421,7 +421,7 @@ Examples
 
 
 ***
-### `++mul`
+### <a name="mul">`++mul`</a>
 
 Multiply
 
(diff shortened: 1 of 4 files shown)


Fang-:
This is so good!

I'm not entirely sure how I feel about sorting these alphabetically rather than by the sections in [hoon.hoon](https://github.com/urbit/arvo/blob/master/sys/hoon.hoon). If I'm looking for things to do with lists, I ideally want to be able to ctrl-f for "list" and see all arms under "2b: list logic".

That may make it slightly too similar to the existing docs pages to be of very much use, but I think having this index function as a quick-ref would be neat. Maybe a format along the lines of the following might be nice?

Arm | Brief | Use
--- | --- | ---
[find](#) | First index in list | `{nedl/(list) hstk/(list)}`
[fand](#) | All indices in list | `{nedl/(list) hstk/(list)}`

That's just my two cents though, and very much up in the air still. Not to mention "he who does the work...". It'd be cool to get feedback from more people on this, see what everyone's looking for in an index page like this.

Also, +1 for name tags on function arms for those #links, but the next iteration of our markdown-to-html rendered will automagically put those in for you, like Github already does. So spare yourself the effort there!

ngzax:
Thanks for your feedback, @Fang-.

I think I headed down the alphabetical route because that's how an index is traditonally organized. 😄 .
Moreover, if you know that you want all arms for lists, you already can go to the root page of the standard library (http://localhost:8080/docs/hoon/library/) and search there. I also just kind of like the fact that now you can just sort of browse and see all the different arms and maybe stumble upon one that you hadn't known about before.

As a compromise, what if we retained alpha ordering, but added the section to it like so:

Arm | Brief | Use | Section
-- | -- | -- | --
find | First index in list | {nedl/(list) hstk/(list)} | 2b: list logic |
fand | All indices in list | {nedl/(list) hstk/(list)} | 2b: list logic |

I will post something about this PR on /fora to see if we can get a wider audience for comments.

Is there any way I can see the in progress version of tree? (I assume that's what you're calling " our markdown-to-html") I looked in the public repo and I don't see any in progress branches.

joemfb:
I'm inclined to think that alphabetical is best. (I guess that's ironic, since I argued against it for the glossary ...) It comes down to which reading patterns should be prioritized: I think an index is most useful for looking up gates referenced somewhere else. And the pages themselves are already grouped by section.

I'm conflicted when it comes to including the sample mold in the table. Is it helpful context, or extra, confusing information? I'll have to leave that to others.

Finally, it's worthing noting that `hoon.hoon` and `zuse.hoon` have been completely refactored on `cc-release`. Section boundaries are different, and gates have been renamed and/or namespace'd (for instance `++sein` will be `++sein:title`). I'm not sure how long it will take to pull this together (or when `cc-release` is expected to be ready) but it's worth anticipating those changes. Maybe the fully namespace'd arm name should be in a separate column?

galenwp:
It wouldn't be all that difficult to make a table like the sample one that @ngzax has drawn up above (https://github.com/urbit/docs/pull/187#issuecomment-320234418) sortable.

It's a toss up to me whether alphabetical or section makes more sense.  If you're looking for a 'kind' of gate then you want to sort by section.

vvisigoth:
Nothing of substance to add, just want to give my support to this idea and lodge another vote in favor of a sortable list, since I can anticipate using one while reading code and while writing.

ngzax:
@galenwp : Thanks for your feedback. 

One of the reasons I made it the way it is now is because I couldn't see any way using any of the table markdown syntaxes I have ever used to make a table in urbit markdown. Is there a syntax? If not, would it need to be added via **tree?**

I am more than willing (and able) to work on this but as I indicated in my comment to @Fang- above I am concerned that there is a newer version of tree coming soon that isn't publicly visible yet.

galenwp:
Currently, this would be added using some kind of tree component. 

The main thing that's changing in the tree update we're working on is the ability to do server-side includes. Basically, you'll be able to write hoon and use %ford to generate your pages. So, this index page could be a function of all the library docs. 

@ohaitch, is this stable enough for someone else to experiment with?

ohAitch:
For stuff like this you already have to write hoon in ford if you want to generate them dynamically, see [`.tree-index`](https://github.com/urbit/arvo/blob/master/ren/tree/index.hoon) which is doing something similar. Once you have it generating the right html, it can be plugged into current tree by adding a tree-json clause, or the future infrastructure more directly from ford.

galenwp:
I mean, isn't the twig reference (https://urbit.org/docs/hoon/reference/) doing something similar client-side?

ohAitch:
That one's less auto-generated, but yes

ngzax:
Thanks guys, I'll check out those pages and see what I can do.

Fang-:
@ngzax Did you see what you can do for this yet? (:

keatondunsford:
This is something that @rmariani is working on as we speak, so I'm going to close this PR for now, and we can always refer back to this if we need.

PR #188, opened 2017-08-28
Add information about turning timestamps on/off in talk

ngzax:
In a discussion in Talk last Friday, I learned for the first time that talk has the ability to turn timestamps on and off using the ;set command. As this was previously undocumented, I decided to add a section to the help doc that discusses talk.


Fang-:
Heh, I neglected to add it into the docs when I added that feature in. Good to finally have it, thanks!

PR #189, opened 2017-09-04
upgrade centar docs

belisarius222:
Add discussion and a couple more examples.


cgyarvin:
This is the one rune I never use.

ohAitch:
!,

belisarius222:
It showed up [here in hoon.hoon](https://github.com/urbit/arvo/blob/maint-20170614/arvo/hoon.hoon#L9316), which is used by the ford rune `/:`, so I was trying to make sense of it.

keatondunsford:
At first glance, usage of this rune looks really ugly compared to the standard centis irregular syntax for this simple example. But it's in the system and should be documented, plus there might be some special case I'm not thinking about where it'd be useful.

Thanks for the update!

ohAitch:
An example with bunting a mold would be useful, yeah

Fang-:
[For example.](https://github.com/Fang-/arvo/blob/talk-split/app/talk-agent.hoon#L415)

PR #190, opened 2017-09-06
docs for ford runes

belisarius222:
Here's the rendered markdown of the new file:
https://github.com/urbit/docs/blob/ford-docs/docs/arvo/internals/ford/runes.md

This should show up as a link on the main ford docs page, under arvo internals.

### DIFF
--- a/docs/arvo/internals/ford/commentary.md
+++ b/docs/arvo/internals/ford/commentary.md
@@ -198,9 +198,9 @@ retrieve it from.
               ==                                            ::
 
 This is how we represent the static resources hook files can load. The
-discussion of their use from a user's perspective is documented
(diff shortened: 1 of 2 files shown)


cgyarvin:
Could use a quick glossary... I know what a horn is but not everyone does.

belisarius222:
@cgyarvin good point; I added a description of horns and some links to and from the ford commentary page. No other concepts jumped out at me as needing to be in a glossary, but it's quite possible I missed something.

galenwp:
This seems perfectly merge-able.  If you want to make further changes, we'll just add them then.  Maybe we can start a fora discussion tomorrow.

PR #192, opened 2017-09-15
Fix /+ and /- docs: ship, case, and multi-arg

belisarius222:
[note: requires ford update before merging]


belisarius222:
@keatondunsford I think this should be ready to merge now.

keatondunsford:
Cool.

PR #196, opened 2017-09-30
fixing an incorrect expansion of ;;

joshuareagan:
The old expansion of ;; was definitely wrong.  I adjusted it so it now matches the relevant code in ++open of hoon.hoon:

https://github.com/urbit/arvo/blob/eee8697782c04e6cbb33a1df682c5fcfa6ccb91d/sys/hoon.hoon#L6724

### DIFF
--- a/docs/hoon/twig/sem-make/sem-fry.md
+++ b/docs/hoon/twig/sem-make/sem-fry.md
@@ -11,13 +11,13 @@ sort: 1
 
 ```
 :pin  :name(a (p q))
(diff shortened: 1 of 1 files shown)


joshuareagan:
I guess the line of hoon.hoon that really matters is this one:

https://github.com/urbit/arvo/blob/eee8697782c04e6cbb33a1df682c5fcfa6ccb91d/sys/hoon.hoon#L6731

Fang-:
This looks much more correct than the original.

PR #197, opened 2017-09-30
Highlight what "cast syntax" is

joshuareagan:
We don't define casting before this point in Urbytes, so it would be helpful to emphasize the part of the example being referred to by "cast syntax".


Fang-:
More clarity is more better. Good stuff!

Issue #160, opened 2017-05-11
explanations of binary tree addressing should use binary, not integers


tcikoritys:
Earlier today in talk, I had asked what the difference was between +2.something and (snag 2 something). This became a discussion about binary trees, addressing, and how to get each item specifically from the tree with +number.something. This syntax was also described in Urbyte 2, but the numbers and terminology used are not well explained, and to me made Urbit seem more foreign and hard to grasp. I would suggest adding that those are binary numbers, and each item in the tree can be retrieved separately by adding a 1 to the beginning of the number. For example, the first item, item 0 is 10 (@ud 2). The second, item 1, is 110 (@ud 6). Item 2 is 1110 (@ud 14), and so on. I would also add that the tail can be retrieved (if wanted) by changing the 0 to a 1. So the individual third item in a binary tree (the head of that cell) is 1110 (@ud 14), and the tail of that cell (the fourth item onward) is 1111 (@ud 15).

With much appreciation,
~lodhul-ricbur

cgyarvin:
This is certainly true and also a great way to describe tree addressing!

galenwp:
Yeah, this makes sense.  Would happily accept a PR that made this change.  

Do we cover tree addressing outside of Urbyte 2?

joshuareagan:
It's explained briefly in the Nock documentation:

https://urbit.org/docs/nock/explanation/

xykj61:
Looks like Josh's PR above solves this. Thanks for the suggestion!

PR #199, opened 2017-09-30
Expanding explanation of tree addressing

joshuareagan:
Following up on discussion at issue #160:

https://github.com/urbit/docs/issues/160


Fang-:
This seems good to me. Thanks!

Issue #145, opened 2017-02-14
`? ~~Ürbit` test in Urbyte 1 broken.


keatondunsford:
@galenwp thinks this is a bug in our UTF-8, not the syntax itself. @cgyarvin, can you verify this?

Should return:
```
> ? ~~Ürbit
  @t
'Ürbit'
```

Edit: By "broken", I mean Dojo blocks the command from being entered (just beeps at you a lot).

PR #200, opened 2017-10-01
Removed broken `? ~~Ürbit` example

joshuareagan:
This example is broken (issue #145).

We still have people trying this out in :dojo and becoming confused when it doesn't work.  Someone was struggling with this today and brought it up in urbit-meta.  We should delete it for now.  Broken examples are frustrating.  When the problem in #145 is corrected, then we can (and should) add it back in.


Fang-:
I agree, we definitely don't want people running into "but they told me this works" situations.

PR #201, opened 2017-10-02
More hand-holding in FizzBuzz demo

joshuareagan:
I tried to make the explanation of the FizzBuzz demo more approachable for those who lack functional programming experience (or any programming experience, for that matter).  Feedback is welcome, but FWIW I remember finding a few lines really puzzling when I first worked through this as a beginner a few months ago.  I think these changes sand away some of the sharper edges.

### DIFF
--- a/docs/hoon/demo.md
+++ b/docs/hoon/demo.md
@@ -32,8 +32,15 @@ Can you understand this code by just looking at it?  Maybe.
 But let's explain it anyway.
 
 Hoon is a functional language.  FizzBuzz is usually defined as a
-side effect, but Hoon has no side effects.  This code will
-*make* a list of strings, not *print* a list of strings.
+side effect, but officially Hoon has no side effects.  This code 
+will *return* a list of strings, not *print* a list of strings. 
+(Of course, if you run this code from the Urbit `dojo` (shell) you 
+will be able to see the product.)
+
+The code produces a list in which each item is either a number 
+(as a string), "Fizz", "Buzz", "Fizzbuzz", or the value `~`.  `~` 
+is null, which indicates the end of the list and hence will always 
+be last.
 
 Line 1 defines a function of one argument, `end`, an unsigned
 integer.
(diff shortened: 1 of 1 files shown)


docs/hoon/demo.md
```diff
@@ -42,22 +48,28 @@ Line 2 declares a variable, `count`, with initial value `1`.
 
 Line 3 begins a loop.
 
-Line 4 casts the product of the loop to a list of `tape`
-(string as a character list).
+Line 4 defines the product of the loop as a list of `tape`s.
```
Fang-: Not sure if I feel comfortable with "defines" here. I'd probably go with the more verbose "define the type of the product of the loop as".

"casts" definitely captures this better, but I guess it's not safe to assume the reader knows what casting it. (Might still be worth mentioning that this is a "cast" though.)

joshuareagan: They know it's a `:cast` from the keyword in the code, they just don't know what that means.  I don't want to repeat the word as I'm explaining that line.

Your verbose version works too but I think it may be unnecessary--- "a list of tapes" underdefines the product, so the reader can infer it's a product type from the context.

Then again, maybe I am assuming a bit too much about what is readily inferred.

Fang-: Keywords will be removed soon, so then "cast" would never show up.

I guess you're right that it under-defines the product. The product that hasn't even been defined itself.

If we end up getting questions about this in urbit-meta, we can always look at it again.

joshuareagan: That's a good point about keyword removal.  I'd forgotten about that.

Will it really be called casting after the change, or will only us early adopters call it that?  Will future Martians say "ket-hepping" instead?

Fang-: It'll still be a same thing, which is practically just a cast as we know it from other languages, it just won't be labeled in the compiler as such. Who knows what future martians might call it. If enough old-timers call it what it is, then surely a "cast" will always be understood to be `^-` (or even `^+`).

belisarius222:
Yeah, the fact that ^+ and ^- both mean "cast" is a good example of why
getting rid of keywords will be nice. I believe they're gone from master
already, and Keaton is working on removing them from the docs.

One thing I didn't realize about ^- until much later was that it casts the
expression under it. It has nothing to do with the gate itself; you can put
a ^- anywhere. Gates themselves don't define their product type. We
generally cast the rest of the gate body in the first line so we can tell
what the product type will be, but the casting is separate from the gate
definition (as opposed to, say, c or Haskell).

Putting the cast at the top of the gate is sometimes just an annotation for
readability purposes and not strictly necessary. Sometimes, however, it is
necessary. The common example of this is you want to produce a list, e.g.
`(list tape)`, but your expression actually returns `["foo" "bar" ~]`,
which nests inside `(list tape)` but can't necessarily be used as one in
other pieces of code, because it doesn't have the right span (type).
Casting to list standardizes the span so the result can be used in arms
like `++turn`. I'm still learning some of this, but it looks like it needs
the `i` and `t` faces of a true list, and also needs its span to
recursively include the possibility of being null, whereas the span of a
concretely defined list without a cast only includes the possibility of
null at the end, where it's actually found. So when we try to use that in
an arm that expects a real list, which is a "fork" type (union) of either
`i` and `t` or `$~` (span of null), then when that arm tries to check if
it's null, it's not a valid operation, because the type system already
knows it's not null. So in this case casting to a list generalizes the type
somewhat. I think this is also called the TMI problem: an arm won't work on
a list without the fork type because it has "too much information" about
the span: it knows it's not null.

```
> (turn ["a" "b" ~] lent)
/~zod/home/~2017.9.29..00.15.58..bac9/arvo/ford:<[1.349 24].[1.349 52]>
-find.i.a
find-fork-d
ford: build failed ~[/g/~zod/use/dojo/~zod/inn/hand
/g/~zod/use/hood/~zod/out/dojo/drum/phat/~zod/dojo /d //term/1]

> (turn `(list tape)`["a" "b" ~] lent)
~[1 1]

> (turn [i="a" t=[i="b" t=~]] lent)
/~zod/home/~2017.9.29..00.15.58..bac9/arvo/ford:<[1.349 24].[1.349 52]>
-need.{i/"" t/$~}
-have.$~
nest-fail
ford: build failed ~[/g/~zod/use/dojo/~zod/inn/hand
/g/~zod/use/hood/~zod/out/dojo/drum/phat/~zod/dojo /d //term/1]

> (turn `(list tape)`[i="a" t=[i="b" t=~]] lent)
~[1 1]

> -:!>([i="a" t=[i="b" t=~]])
p=#t/{i/"" t/{i/"" t/$~}}

> -:!>(`(list tape)`[i="a" t=[i="b" t=~]])
p=#t/*""

> ?~  ["a" "b" ~]  %.y  %.n
/~zod/home/~2017.9.29..00.15.58..bac9/arvo/ford:<[1.349 24].[1.349 52]>
mint-vain
ford: build failed

> ?~  `(list tape)`["a" "b" ~]  %.y  %.n
%.n
```

belisarius222:
Not sure how much of that should be included in the intro docs, but maybe
some docs on how casting works in hoon would be useful, and the fizzbuzz
example could link to them for more details.

PR #205, opened 2017-10-03
Partial Rewrite of Hoon Syntax doc

joshuareagan:
....or at least linking them to the glossary.  Some of the explanations I give are pretty dumb, but in these cases a link to the real definition is right there.  I think this kind of bootstrapping can be useful, but maybe it's too inexact.  Slap me down if you must, but let's not leave this doc as unfriendly as it currently is for beginners.

### DIFF
--- a/docs/hoon/syntax.md
+++ b/docs/hoon/syntax.md
@@ -72,13 +72,22 @@ higher-order programming; write-only Hoon can certainly be
 # Design
 
 Hoon expressions are called 
-*[twigs](https://urbit.org/docs/about/glossary#twig)*. The name 
-of a twig is its 
-*[stem](https://urbit.org/docs/about/glossary#stem)*.  Because 
-each twig stem has its own structure, each has its own parser
-rule. We'll describe these grammars as we go through the stems.
+*[twigs](https://urbit.org/docs/about/glossary#twig)*, and at the 
+beginning of each twig is a 
+*[rune](https://urbit.org/docs/about/glossary#rune)*.  Each rune 
+type has a name, called a 
+*[stem](https://urbit.org/docs/about/glossary#stem)*.  Finally, 
+each stem has its own structure, e.g.:
 
-But Hoon has a general syntax design with common principles and
+> :-
+>   (add 12 13)
+> (sub 10 7)
+
+This twig uses the rune `:-` (stem: `cons`) to produce an ordered 
+pair: `[25 3]`.  There are two subexpressions following the `:-`, 
+corresponding to the two parts of the `cons` structure.  
+
+Hoon has a general syntax design with common principles and
 regularities.  We'll cover those here.
 
 ## Glyphs and characters
@@ -139,48 +148,19 @@ principles at all.  All irregular forms are flat.
 
 Some twigs have *only* irregular forms.
 
-## Keywords versus runes
+## Runes
 
-A regular form starts with a 
-*[sigil](https://urbit.org/docs/about/glossary#sigil)*, which is 
-either a *keyword* or a 
-*[rune](https://urbit.org/docs/about/glossary#rune)*&mdash;at 
-the programmer's choice.
-
-A *keyword* is `:` and then the stem label&mdash;e.g., `:cons`.
+As mentioned earlier, a regular form starts with a 
+*[rune](https://urbit.org/docs/about/glossary#rune).
 
 A *rune* is a pair of ASCII punctuation marks (a digraph)&mdash;e.g.,
 `:-`.  The first glyph in the rune indicates the category&mdash;e.g.,
 `:` runes make cells.  Runes can be pronounced by their glyphs or
-by their stem -- for `:-`, you can say either "colhep" or "cons".
-
-Keywords may be less challenging for early learners.  Rune style
-is less cluttered and distracting for experienced programmers;
-the system codebase is all in rune style.   Beginners can start
-with keywords and move to runes; start with runes and stick
-with runes; or start with keywords and stick with keywords.
-
-Here's the FizzBuzz demo code, in idiomatic rune syntax.  Compare
-with the [original](../demo).
-
-```
-  |=  end/atom
-  =/  count  1
-  |-  ^-  (list tape)
-  ?:  =(end count)  ~
-  :_  $(count (add 1 count))
-  ?:  =(0 (mod count 15))
-    "FizzBuzz"
-  ?:  =(0 (mod count 5))
-    "Fizz"
-  ?:  =(0 (mod count 3))
-    "Buzz"
-  (pave !>(count))
-```
+by their stem&mdash;for `:-`, you can say either "colhep" or "cons".
 
 ## Tall regular form
 
-Tall regular form starts with the sigil, followed by a `gap`
+Tall regular form starts with the rune, followed by a `gap`
 (any whitespace except `ace`). After this is a 
 *[bulb](https://urbit.org/docs/about/glossary#bulb)*&mdash;the 
 contents of the twig&mdash;whose own twigs are separated by 
@@ -188,7 +168,7 @@ contents of the twig&mdash;whose own twigs are separated by
 
 There are four body subtypes: *fixed*, *running*, *jogging*, and
 *battery*.  Stems with a *fixed* number of subexpressions
-self-terminate.  For instance, `:if` has three subexpressions and
+self-terminate.  For instance, `?:` has three subexpressions and
 is self-terminating.  Otherwise the twig is terminated by a
 `gap`, then either `==` (*running* or *jogging*, most twigs) or
 `--` (*battery*).
@@ -312,7 +292,7 @@ conventional example:
 
 ## Flat regular form
 
-Flat regular form starts with the sigil, followed by `pal`
+Flat regular form starts with the rune, followed by `pal`
 (`(`, left parenthesis), followed by a bulb whose subexpressions
 are separated by `ace` (one space), followed by `par` (`)`, right
 parenthesis).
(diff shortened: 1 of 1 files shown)


Fang-:
Additionally, while you're touching this,

> Runes can be pronounced by their glyphs or by their stem -- for `:-`, you can say either "colhep" or "cons"

Since the stems (keywords) are going to be changed to be more or less equivalent to the glyphs (ie `%clhp`, we should probably start getting rid of mentions like these in the docs.

joshuareagan:
Before I make the change, just so I'm clear, are we killing the old keywords entirely?  I.e. will `:cons` no longer work once the change is made?

Second question: will `:clhp` take the place of `:cons`, so that we effectively have new keywords?  Or will there be no keywords at all?

Fang-:
Yes, the old keywords will be entirely gone, and using `:kwds` in Hoon code will no longer be a thing you can do.

joshuareagan:
In that case, I need to go back and change some things in the FizzBuzz example doc too.  I'll do that first and then swing back around and hit this one again.

joshuareagan:
This is definitely a work in progress, but I've taken the keyword parts out.

Fang-:
> This is definitely a work in progress

Looks pretty good to me, ping whenever you feel this is ready!

joemfb:
I was going to ask if "stem" really carries its weight in introductory content like this. (I had it in my mind that "stem" and "bulb" were general terms for "tag" and "payload" in a "page" of a `$%` tagged-union.)

But then I looked around, and it appears to be documentation-only terminology for tagged-unions, but only in the case of twigs. So I'll broaden my question: do "stem" and "bulb" carry their weight *anywhere*?

joshuareagan:
This is one of the things I was hesitating over last night---I don't think we need to talk about stems and bulbs in this doc.  (As for whether those terms are altogether unnecessary, I don't know.)

cgyarvin:
I am honestly on the fence with this one.  We need some kind of terminology for a tagged union... but...

belisarius222:
The question of how we name tagged unions should be orthogonal to the question of how we describe hoon syntax. The fact that those things are currently complected indicates a layer violation. We shouldn't be teaching people the internal data structures of the compiler at all when they're learning hoon. Just because the compiler parses an expression into some particular data structure doesn't imply that a beginner should be confronted with that concept when they're trying to learn the language itself.

Teaching the syntax in terms of the compiler internals is teaching in terms of the implementation rather than the specification. It's good for people to learn _what_ the compiler does, but forcing them to learn _how_ it does what it does creates a bootstrapping problem: in order to learn the language, a student first has to learn how the compiler works, but the compiler is written in the language they don't understand yet!

We should have separate documentation on how the compiler works, and we can discuss `++twig` in full detail there, with all its stems and bulbs. However, a lesson on compiler internals must be presented _after_ the lessons on syntax.  We can and should remove extraneous information about internals from the basic docs. Enforcing twin disciplines of modularity and unidirectional dependency in our docs will prevent us from creating an ouroboros of circularly dependent lessons.

An extreme version of this position would also indicate that we should stop using the word "twig" in the intro docs. I don't think I'd go that far, though. It's ok to present new words here and there, especially ones that are easily absorbed and that will ease future learning. We just need to make sure we stay within a vocabulary budget, since there's cognitive cost to each new term encountered. I don't think the concept of a twig puts too much cognitive load on the student, and it's used a lot both in these docs and later docs, so I think it's worth introducing it now. Stem and bulb are rarely used (I forgot those terms even existed until re-reading these docs), and introduction of those concepts doesn't aid much in understanding the syntax compared to an explanation without using them, so I'm in full agreement with @joemfb and @joshuareagan that we should remove them from these docs.

joshuareagan:
Okay, I've gone back and edited/rewritten a few chunks of the original document.  I removed references to "stems" and "bulbs", and I tried to make sure I defined all Urbit-specific words in a fairly gentle way.  Some of these changes may be a little controversial, but I'm honestly not sure.  I tried to make revisions that I would have liked to be there when I was first learning.

I'm very confident that the result is more friendly to newcomers.  Objections are welcome.

docs/hoon/syntax.md
```diff
@@ -112,76 +94,113 @@ fas /           pat @               wut ?
...
+The *running* body has a list of *children* (i.e., subexpressions). 
+The *jogging* body has a list of child pairs, where the members of 
+each pair is separated by a `gap`.  The *battery* body is
+a list of symbol-child pairs, separated by a gap, prefixed by `++`
 and then a gap.
 
-This definition is enough to parse.  But the proper shape of the
-whitespace gaps demands an informal convention.
+This definition is enough to write Hoon that will parse.  But the 
+proper shape of the whitespace gaps rquiress an informal convention.
```
belisarius222: rquiress -> requires

joshuareagan: hah, fixed

belisarius222:
Looks good to me!

joemfb:
Should it be tall vs flat, or tall vs wide? It seems like a pretty even split across all the docs, with wide pre-dating flat.

docs/hoon/syntax.md
```diff
@@ -7,22 +7,19 @@ title: Syntax
 
 # Hoon syntax
 
-Hoon's semantics are simple, but unusual.  Hoon's syntax is
-complex, and unusual.  Sorry.
+Hoon's syntax is unusual, and to beginners it can look a lot more 
+complex than it really is. However, we believe that anyone who 
+learns Hoon syntax will find that it has a number of advantages.
```
joemfb: I think that this section might perhaps be a little bit too deferential. I'm inclined to keep the assertion that Hoon's semantics are simple. And its syntax really is complex.

joshuareagan:
Huh, I didn't realize that.  I have no stance.  Either "flat" or "wide" seems fine.  We should probably pick one for the sake of consistency.

docs/hoon/syntax.md
```diff
@@ -97,100 +94,114 @@ fas /           pat @               wut ?
...
-You don't need to memorize these glyph names, but it helps.  Code
-is often vocalized or subvocalized.   "pal" is easier to say,
-aloud or silently, then "left paren."
+You don't need to memorize these glyph names, but it might help. 
+Code is often vocalized or subvocalized. "pal" is easier to say,
+aloud or silently, than "left paren".
 
 Note that the list includes two separate whitespace forms: `ace`
-for a single space, `gap` for 2+, comment or newline.  In Hoon,
+for a single space, `gap` for 2+, comment, and newline.  In Hoon,
```
joemfb: This should be "or", `gap` is vaguely like "any of the following". But i think this might be too much parsing detail. A single space vs newlines/2-or-more is the important distinction.

docs/hoon/syntax.md
```diff
@@ -97,100 +94,114 @@ fas /           pat @               wut ?
...
+pairs). We pronounce runes using their glyph names&mdash;for `:-`, 
+we say "colhep".
+
+Each rune is followed by one or more subexpressions. The number and 
+kind of subexpressions depend on the rune.
+
+```
+:-
+  25
+3
```
joemfb: I'm 100% opposed to bare runes on their own line. In this case, I have a strong preference for

```
:-  25
3
```

docs/hoon/syntax.md
```diff
@@ -97,100 +94,114 @@ fas /           pat @               wut ?
...
-
-A *rune* is a pair of ASCII punctuation marks (a digraph) - e.g.,
-`:-`.  The first glyph in the rune indicates the category - e.g.,
-`:` runes make cells.  Runes can be pronounced by their glyphs or
-by their stem -- for `:-`, you can say either "colhep" or "cons".
+```
+.=
+  22
+23
+```
```
joemfb: Ibid

belisarius222:
I prefer "flat," since the spaces are actually less wide than in tall-form. And I agree we should pick one.

belisarius222:
Also, "flat" conveys that it's the one-line form.

joshuareagan:
I'm sold.

docs/hoon/syntax.md
joshuareagan: (To others: we're battling this out in urbit-meta now.)  I remain unconvinced.  The syntax is unusual, but it isn't as complex as it looks.  We don't want to demoralize newcomers who have heard rumors of Urbit's supposed inscrutability.

docs/hoon/syntax.md
joshuareagan: I'm open to whatever seems good here.  I have no strong views about this.

docs/hoon/syntax.md
joshuareagan: (1) New lines seem important to me, and we use them all over this doc.

(2) The sentence reads more naturally to me with the "and" but I don't have strong feelings about it.  If most disagree I'll switch it back with no fuss.

joemfb: Put another way, ace vs gap is both important and very simple. I think bringing in comments is confusing and unnecessary (I'm sure most people don't consider comments to be whitespace, and I doubt many are thinking in terms of the parsing strategy this is implicitly explaining).

But this is stylistic editorializing, maybe I should just open a PR after you land this ...

joshuareagan: I can live with dropping comment here.

joshuareagan: changed

docs/hoon/syntax.md
Fang-: Having read bits and pieces of the discussion in urbit-meta, I have to side with Joshua here.

Telling people "this is hard, good luck lol!" is setting them up for failure. Someone with little faith in their own ability to learn new things will jump ship _really_ hard at that. A little reassurance like "this is gonna be weird, but it's totally doable" goes a long way.

(Also, really, it's not so much the syntax that's complex, it's just that there's lots of runes to learn. I feel it might also be useful to point out that you only need to know a small subset of the runes to write working (albeit not very pretty) Hoon.)

keatondunsford:
This looks mergeable to me. @joshuareagan, want to keep working on keyword removal on another PR?

joshuareagan:
Yes, I'll do that.  Thanks

PR #206, opened 2017-10-05
Changing demo.md from keywords to runes

joshuareagan:


Fang-:
Good stuff! My only little gripe with this is that you introduce the term `rune` and a link to the relevant documentation on line four, instead of right off the bat. Not sure if you want to integrate this into the introductory paragraph or just say "`|= [rune](#)`".

joshuareagan:
yeah that's definitely confusing...

joshuareagan:
Fixed

docs/hoon/demo.md
```diff
@@ -41,33 +38,38 @@ The code produces a list in which each item is either a number
...
+*[rune](https://urbit.org/docs/about/glossary#rune)* to define a 
+function of one argument, `end`, which is of the type `atom` (i.e. 
+an unsigned integer).
 
 Line 2 declares a variable, `count`, with initial value `1`.
 
 Line 3 begins a loop.
 
 Line 4 defines the product of the loop as a list of `tape`s. 
 (A `tape` is a list of characters, one of Hoon's two string
```
joemfb: This is not technically correct; `(list @c)` is a list of "characters". A type is a list of character-bytes, which can be a very different thing. Probably not relevant here, but I just can't help myself ...

joshuareagan: There's a lot that isn't technically correct in this doc.  Hoon doesn't have loops!

Fang-: I think it's fair to trade in a bit of technical correctness for some slightly slimmer semantics.

PR #207, opened 2017-10-05
removed Keyword definition in glossary

joshuareagan:


Fang-:
On line 216, you want to do ``[`?:`](url)`` instead of `` `[?:]`(url)`` to do the link properly.

`[?:]`(#) --> [`?:`](#)

joshuareagan:
Fixed, good catch

Issue #202, opened 2017-10-03
Glossary Organization


joshuareagan:
I'm going to start adding some definitions to the glossary---it's pretty short right now.  But the organization seems a bit haphazard.  In a way it seems like there are sections: a ship section, a Nock section, and a Hoon section---but these seem not to have much internal order.  I'm sure most people use something like ctrl+f to find what they want anyway, but it still seems like we should clean this thing up.

Alphabetical order seems like the natural choice.  Some things we can group together---it would be a shame to break up the Nock section, say---but otherwise alphabetical?  Thoughts?

galenwp:
I think alphabetical order per-section makes sense.  It'd also be nice if each definition had an id so you could `#link` to it.

joshuareagan:
Good idea.

joemfb:
This time around, I am choosing to have no dog in this fight. Historical context is here: https://groups.google.com/forum/#!topic/urbit-dev/bjQ2B767yS8.

joshuareagan:
The glossary is a reference document, not a tutorial, so I think alphabetizing it makes the most sense.  I'm not a hug fan of the sections, though I understand why others like them.  But for a glossary, the three main priorities should be (1) formal correctness, (2) completeness of coverage, and (3) it's easy to look up a term on the page.  Naturally most people will use ctrl+f anyway, but it's standard practice to alphabetize as well.

xykj61:
Yeah, you're right. K, I'll merge your stuff. Thanks!

PR #209, opened 2017-10-12
alphabetized definitions within glossary sections

joshuareagan:


Fang-:
Not entirely sure how to feel about alphabetizing that tiny list of Nock terms at the start. The order there feels like the canonical one to me, makes a lot of sense, and the list is short enough to search without scrolling all over the place.  
Maybe that's just me though. I suppose there's something to be said for consistency with the rest of the page.

Also, some of the headers are `###### headers`, others are `<h6 id="word">headers</h6>`, and some are both. Not sure how Urbit's markdown rendering deals with this, but GitHub makes two `<h6>` elements for those that are both. It'd be nicest to have them all be  `<h6 id="word">`, so that they all can be #linked to.

joshuareagan:
issue #202

keatondunsford:
@joshuareagan @belisarius222 I'm thinking that the order of the Glossary as a whole should match the new proposed Hoon tutorial outline: i.e. it should go from Nock up. Even if those docs aren't written, I think the outline is good enough to start rearranging this section and potentially start adding more concepts.

What do you think about this?

joshuareagan:
I left a comment on the issue thread about the glossary: #202

PR #212, opened 2017-10-26
Arvo retrofit

mattlevan:
Updated code blocks and explanations to comply with Hoon %143 using urbit/examples:retrofit branch. Progress up through `security-driver.md`. I think `app/gh.hoon` needs to be fixed before I can continue work on the `api.md` doc as the GitHub issues exploration doesn't work correctly for me. I'll be looking into that further soon...

### DIFF
--- a/developer/docs/arvo/generators.md
+++ b/developer/docs/arvo/generators.md
@@ -90,27 +90,27 @@ The app can easily be modified to use this (`/examples/app/ping.hoon`):
 ::                                                      ::  6
 !:                                                      ::  7
 |%                                                      ::  8
-++  move  {bone card}                                   ::  9
-++  card  $%  {$poke wire dock poke-contents}           ::  10
+++  move  [bone card]                                   ::  9
+++  card  $%  [$poke wire dock poke-contents]           ::  10
           ==                                            ::  11
-++  poke-contents $%  {$atom @}                         ::  12
+++  poke-contents $%  [$atom @]                         ::  12
                   ==                                    ::  13
 --                                                      ::  14
-|_  {bow/bowl $}                                        ::  15
+|_  [bow=bowl $]                                        ::  15
 ::                                                      ::  16
 ++  poke-ping-message                                   ::  17
   |=  ping-message                                      ::  18
   ~&  ping+'Message sent!'                              ::  19
-  ^-  {(list move) _+>.$}                               ::  20
+  ^-  [(list move) _+>.$]                               ::  20
   :_  +>.$                                              ::  21
   :_  ~                                                 ::  22
   [ost.bow %poke /sending [to dap.bow] %atom message]   ::  23
 ::                                                      ::  24
 ++  poke-atom                                           ::  25
-  |=  arg/@                                             ::  26
+  |=  arg=@                                             ::  26
   ~&  ping+'Message received!'                          ::  27
-  ^-  {(list move) _+>.$}                               ::  28
-  ~7  ping+message+(@t arg)                             ::  29
+  ^-  [(list move) _+>.$]                               ::  28
+  ~&  ping+message+(@t arg)                             ::  29
   [~ +>.$]                                              ::  30
 ::                                                      ::  31
 ++  coup  |=(* `+>)                                     ::  32
@@ -136,11 +136,11 @@ everything nicer. Since it's specific to `:ping`, let's put it in
 
 ```
 :-  %say
-|=  {^ {to/@p message/?($~ {text/@t $~})} $~}
+|=  [^ {to=@p message=?($~ [text=@t $~])} $~}
 [%ping-message to ?~(message 'howdy' text.message)]
 ```
 
-A couple of new things here. Firstly, `message/?($~ {text/@t $~})` should be 
+A couple of new things here. Firstly, `message=?($~ [text=@t $~])` should be 
 read as "the message is either null or a pair of text and null". Generator 
 argument lists are always null-terminated, which makes it convenient to accept 
 lists in tail position (which are particularly annoying without generators). 
--- a/developer/docs/arvo/generators.md
+++ b/developer/docs/arvo/generators.md
@@ -1,5 +1,5 @@
 ---
-navhome: /developer/docs
+navhome: /developer/docs/
 next: True
 sort: 18
 title: Using generators with apps
--- a/developer/docs/arvo/generators.md
+++ b/developer/docs/arvo/generators.md
@@ -90,10 +90,10 @@ The app can easily be modified to use this (`/examples/app/ping.hoon`):
 ::                                                      ::  6
 !:                                                      ::  7
 |%                                                      ::  8
-++  move  [bone card]                                   ::  9
-++  card  $%  [$poke wire dock poke-contents]           ::  10
++=  move  [bone card]                                   ::  9
++=  card  $%  [$poke wire dock poke-contents]           ::  10
           ==                                            ::  11
-++  poke-contents $%  [$atom @]                         ::  12
++=  poke-contents $%  [$atom @]                         ::  12
                   ==                                    ::  13
 --                                                      ::  14
 |_  [bow=bowl $]                                        ::  15
--- a/developer/docs/arvo/generators.md
+++ b/developer/docs/arvo/generators.md
@@ -1,5 +1,5 @@
 ---
-navhome: /developer/docs/
+navhome: '/developer/docs'
 next: True
 sort: 18
 title: Using generators with apps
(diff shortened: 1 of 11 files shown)


developer/docs/arvo/generators.md
```diff
@@ -7,141 +7,149 @@ title: Using generators with apps
...
+`[%receiving 'heyya']`. You can test this yourself with a couple of comets.
+
+This is an annoying way to invoke the app, though. The `[]` are mandatory, 
+and optional arguments are hard. Let's make a generator, `+send` to make 
+everything nicer. Since it's specific to `:ping`, let's put it in 
+`/gen/ping/send.hoon`:
+
+```
+:-  %say
+|=  [^ {to=@p message=?($~ [text=@t $~])} $~}
```
keatondunsford: Want to fix these curly braces here?

developer/docs/arvo/generators.md
```diff
@@ -7,141 +7,149 @@ title: Using generators with apps
...
+[%ping-message to ?~(message 'howdy' text.message)]
+```
+
+A couple of new things here. Firstly, `message=?($~ [text=@t $~])` should be 
+read as "the message is either null or a pair of text and null". Generator 
+argument lists are always null-terminated, which makes it convenient to accept 
+lists in tail position (which are particularly annoying without generators). 
+`?(a b)` is the irregular form of `$?(a b)`, which is a union type. Thus, 
+`?(a b)` means the type of anything that's in either `a` or `b`. Thus, in our 
+case, `?($~ {text/@t $~})` is either null or a pair of text and null.
```
keatondunsford: Also here.

Additionally, all symbols like `$foo` and `$~` can be switched to `%foo` and just `~`. If you need to manually switch from "value/noun mode" to "pattern/mold mode", you can use `,`, like `,%foo` and `,~`.

developer/docs/arvo/generators.md
```diff
@@ -159,20 +167,22 @@ else `c`". It is roughly equivalent to `?:(=(~ a) b c)`.
...
-    [State](/developer/docs/arvo/state) so that you can run `:examples-sum|add 5`
+-   Create a generator for `:sum` from
+    [State](/docs/arvo/state) so that you can run `:sum|add 5`
     to add numbers to it.
 
--   Create a generator for `:examples-click` from [Web
-    Apps](/developer/docs/arvo/web-apps) so that you can run
-    `:examples-click|poke` to poke it.
+-   Create a generator for `:click` from [Web
+    Apps](/docs/arvo/web-apps) so that you can run
```
keatondunsford: These can stay `/developer/docs/...` for now. Galen and I might change this back, but let's keep things consistent for now.

developer/docs/arvo/http.md
```diff
@@ -7,81 +7,99 @@ title: HTTP requests and timers
...
+::  Up-ness monitor. Accepts atom url, 'on', or 'off'   ::  1
+::                                                      ::  2
+::::  /===/app/up/hoon                                  ::  3
+  ::                                                    ::  4
+!:                                                      ::  5
+::                                                      ::  6
+|%                                                      ::  7
++=  move  [bone card]                                   ::  8
++=  card                                                ::  9
+  $%  [$hiss wire $~ $httr [$purl p=purl:eyre]]         ::  10
```
keatondunsford: I forgot to go in and change all these symbols as per the above comment (i.e. `$purl` to `%purl`), so I'll do that soon and then you can just re-copy-paste in the updated examples.

developer/docs/arvo/http.md
```diff
@@ -7,81 +7,99 @@ title: HTTP requests and timers
...
 
-There's two kinds of cards we're sending to arvo. A `%hiss` move tells
-`%eyre` to make an HTTP request. It expects a `(unit iden)`, which will
-be null here because we aren't doing any authentication; a mark, in this
-case the http request mark `&httr`; and some data in the form of a
-`cage`, which is a `{mark vase}`. This is confusing, so the mold in the
-code quoted above includes a bunch of otherwise unnecessary faces for
-the purpose of illustration. Compare: `{$hiss p/(unit iden) q/mark
-r/cage}` from `zuse`.
+There's two kinds of cards we're sending to arvo. A `%hiss` move tells `%eyre`
```
keatondunsford: Just a PSA, I'm going to try and normalize everywhere stopping the usage of `%vane` instead of Vane. It gets annoying when comparing to other symbol values in the same context, like right here.

I kind of want to do the same for `:app` instead of App (I've been doing that mostly everywhere with Talk and Dojo, because why not, and we'll have new Talk soon that allows capital letters), but that might be a slippery slope since we also have `+generators`, `/+libraries` and `/-surfaces`.

developer/docs/arvo/http.md
```diff
@@ -95,47 +113,46 @@ r/cage}` from `zuse`.
...
 
-When you send this request, you can expect a `%sigh` with the
-response, which we handle later on in `++sigh-httr`.
+When you send this request, you can expect a `%sigh` with the response, which 
+we handle later on in `++sigh-httr`.
 
-For `%wait`, you just pass a [`@da`]() (absolute date), and arvo will
-produce a `%wake` when the time comes.
+For `%wait`, you just pass a [`@da`](/../../hoon/library/3c/) (absolute date), 
+and arvo will produce a `%wake` when the time comes.
```
keatondunsford: We should also capitalize Nock, Hoon and Arvo everywhere we see it. Not hugely important, but we should be consistent.

developer/docs/arvo/http.md
```diff
@@ -145,63 +162,76 @@ progress, then we send an HTTP request.
...
 
-When the HTTP response comes back, we handle it with
-`++sigh-httr`, which, along with the wire the request was sent
-on, takes the status code, the response headers, and the response
-body.
+> With this, notice how the move is followed by a `~`. This is a convenient 
+> shortcut for creating a list of a single element. It's part of a small 
+> family of creating a list of a single element. It's part of a small family 
+> of such shortcuts. `~[a b c]` is `[a b c ~]`, `[a b c]~` is `[[a b c] ~]` 
+> and `\`[a bc]` is `[~ a b c]`. These may be mixed and matched to create
```
keatondunsford: Looks like `[a bc]` is a small typo.

developer/docs/arvo/http.md
```diff
@@ -145,63 +162,76 @@ progress, then we send an HTTP request.
...
+HTTP request as before. Otherwise, we set `in-progress` to false.
 
 Let's try it out:
 
 ```
-~fintud-macrep:dojo> |start %up
+~fintud-macrep:dojo/examples> |start %up
 >=
-~fintud-macrep:dojo> :examples-up &atom 'http://www.google.com'
+~fintud-macrep:dojo/examples> :up &atom 'on'
```
keatondunsford: It's probably better style to pass in `%on` instead of `'on'`, though technically it doesn't matter.

developer/docs/arvo/marks.md
```diff
@@ -7,125 +7,130 @@ title: Marks
...
-We've used predefined marks already, but we haven't yet created
-our own marks.  Let's write a sample mark of our own, then chain
-it together with some preexisting ones to have fun with type
-conversions.
+We've used predefined marks already, but we haven't yet created our own marks. 
+Let's write a sample mark of our own, then chain it together with some 
+preexisting ones to have fun with type conversions.
 
-Let's make a small "examples-cord" mark.  "Cord" is a name we use for `@t`
+Let's make a small "cord" mark. "Cord" is a name we use for `@t`
```
keatondunsford: Maybe link to `++cord` in hoon.hoon, or point to the library section on strings.

developer/docs/arvo/marks.md
```diff
@@ -7,125 +7,130 @@ title: Marks
...
+  |%                                                    ::  8
+  ++  atom  `@`cod                                      ::  9
+  --                                                    ::  10
+--                                                      ::  11
 ```
 
-`/?  314` is the required
-version number, just like in apps.  After that everything's in a
-`|_`, which is a `|%` core, but with input arguments.  In our
+`/?  314` is the required version number, just like in apps. After that
```
keatondunsford: I'm actually not sure what the new Hoon+Zuse version number is, but we should check and change this as it's now something below `314`.

developer/docs/arvo/network.md
```diff
@@ -15,38 +15,43 @@ All we've written up until now are just shell commands that produce a value and
...
-      [~ +>.$]                  :: produce a cell of empty list and our state
-    --                          :: end of poke-noun arm definition
+```
+::  Accepts any noun from dojo and prints it out        ::  1 
+::                                                      ::  2 
+::::  /===/app/echo/hoon                                ::  3 
+  ::                                                    ::  4 
+!:                                                      ::  5 
+|%                                                      ::  6 
++=  move  {bone card}                                   ::  7
```
keatondunsford: Yeah, I'll make the examples styling more consistent since now I think I actually know all the things to check/change. Then we can make all this code consistent with that.

developer/docs/arvo/network.md
```diff
@@ -15,38 +15,43 @@ All we've written up until now are just shell commands that produce a value and
...
-    --                          :: end of poke-noun arm definition
+```
+::  Accepts any noun from dojo and prints it out        ::  1 
+::                                                      ::  2 
+::::  /===/app/echo/hoon                                ::  3 
+  ::                                                    ::  4 
+!:                                                      ::  5 
+|%                                                      ::  6 
++=  move  {bone card}                                   ::  7 
++=  card  $%  $~                                        ::  8
```
keatondunsford: ^ same comment applies to symbols.

developer/docs/arvo/network.md
```diff
@@ -88,28 +93,41 @@ Let's look at another example (edit your code in
...
+  ~&  square+error+'Poke failed. Error:'                ::  26
+  ~&  square+error+err                                  ::  27
+  [~ +>.$]                                              ::  28
+::                                                      ::  29
+```
 
 A few things have changed. Firstly, we no longer accept arbitrary nouns because
 we can only square atoms (integers, in this case an unsigned one). Thus, our
-argument is now `tom/@`. Secondly, it's `++poke-atom` rather than `++poke-noun`.
+argument is now `tom=@`. Secondly, it's `++poke-atom` rather than `++poke-noun`.
```
keatondunsford: I should probably change these stupid names.

developer/docs/arvo/network.md
```diff
@@ -155,54 +175,60 @@ Sending a message to another urbit
...
+~fintud-macrep:dojo/examples> |start %pong
+>=
+~fintud-macrep:dojo/examples> :pong &urbit ~sampel-sipnym
+>=
+```
 
 Replace `~sampel-sipnym` with another urbit. The easiest thing to do is to start
 a comet, a free and disposable Urbit identity. If you don't know how to start a
-comet, see [the user setup section](/developer/using/setup/). Don't forget to start
+comet, see [the user setup section](/developer/docs/using/setup/). Don't forget to start
```
keatondunsford: Nope, the above path is current, but Galen and I don't like that this breaks things, so we'll probably change all of these links back in a different commit.

developer/docs/arvo/subscriptions.md
```diff
@@ -16,144 +16,147 @@ that it's built into arvo.
...
+::                                                      ::  61
+--                                                      ::  62
+```
 
 Cheat sheet:
 
 -   `&` (pam) can either be the boolean true (as can `%.y`, `0`), or
     the irregular wide form of the `?&`
-    ([wutpam](../../hoon/twig/wut/pam)) rune, which computes
+    ([wutpam](../../hoon/twig/wut-test/pam-and)) rune, which computes
```
keatondunsford: These should go back to `/foo/bar` from `/foo-word/bar-word`.

developer/docs/arvo/subscriptions.md
```diff
@@ -312,8 +319,10 @@ Otherwise, if our input is `%off` and we're already subscribed (i.e. `val`
 is false), then we unsubscribe from `:source` and set `val` back to true (`&`), 
 again using our handy inverted cell constructor mold `:_`:
 
-    :_	+>.$(val &)
-    ~[[ost.bow %pull /subscribe [our.bow %source] ~]]
+```
+    :_  +>.$(val &)                                     ::  28
+    ~[[ost.bow %pull /subscribe [our.bow %source] ~]]   ::  29
```
keatondunsford: Maybe we should change this to [foo bar baz]~ like the other apps for consistency.

developer/docs/arvo/web-apps.md
```diff
@@ -147,8 +193,8 @@ Please pardon the awkwardness of the name 'examples-click-click', but we used it
 
 # Frontend
 
-That's all that's needed for the back end.  The front end is just
-some "sail" html (Hoon markup for XML) and javascript.  Here's `/web/pages/examples/click.hoon`:
+That's all that's needed for the backend. The frontend is just some "sail" html
```
keatondunsford: Let's capitalize and italicize _Sail_ here.

keatondunsford:
Alright, finally merging this. Thanks @mattlevan!

keatondunsford:
Actually, going to make a few of those commented changes, then merge it.

mattlevan:
Cool! Thanks for fixing those last few things I never got around to, @keatondunsford!

PR #214, opened 2017-11-08
Update install.md

todlur:
specified that openssl 1.0.0 is the dependency instead of openssl, because some suites (specifically in my case with debian's 'stretch') don't have backports for openssl because specifically the issue I had with installing was one of its dependencies, libssl1.0.0 which seems to be an explicit required version. libssl1.1 and libssl1.0.2 both did not complete the urbit install. I was able to get libssl1.0.0 from jessie and install it onto stretch which solved the issue.

### DIFF
--- a/docs/using/install.md
+++ b/docs/using/install.md
@@ -64,7 +64,7 @@ Urbit depends on:
     curses
     gmp
     libsigsegv
-    openssl
+    openssl 1.0.0
 
 Which can usually be installed with the following one-liners:
 


galenwp:
Should we update the related one-liners?

cgyarvin:
We should update the code, I think, so we don’t have this weird historical dependency?

keatondunsford:
@belisarius222 would probably know the right way to approach this, having just dealt with building the new `urbit-0.5.0` package releases.

belisarius222:
Lines 5 and 11 of this file determine Urbit's build-time and run-time dependencies on debian:
https://github.com/urbit/urbit/blob/retrofit/debian/control

Doesn't look like either of them specify a version. I only tested the debian package on ubuntu, and I have no idea what debian "stretch" implies in terms of package versions, so I'm not sure what the moral of this story is.

Maybe the moral is we should pin the versions of all our dependencies, just to be on the safe side.

todlur:
Well the thing about changing the one-liners depends on what OS someone is on and _their suite's repositories_, like Debian Stretch is actually **forced** to install the insufficient dependencies for the Urbit install using apt-get, as mentioned there are no backports for libssl1.0.0

The readme file for installation would probably need the clarification too, and even the deb package could do with the change too, so that it can support Stretch. [The Jessie package I used](https://packages.debian.org/jessie/libssl1.0.0) has not conflicted with anything else using higher versions of libssl (like OpenVPN uses libssl1.0.2). 

Big picture, the problem I see is the outdated dependency for libssl1.0.0 in general, newer suites will probably stop backporting old versions and I think it's obvious to point out that it will make it hard for noobs to complete the Urbit install in the future.

Although I'm too much of a noob to know whether the underlying dependency for the version is fragile to be changed or not but yeah.

belisarius222:
@todlur Ah I see, "stretch" is the newest version of debian. I thought it was some bleeding-edge version. We should almost certainly update our dependencies so they're included in the standard distros. 

I would say that pinning the version in the meantime is a good idea, but on my ubuntu machine I seem to  have libssl=1.0.2 and openssl=1.0.2 installed, and `apt-cache policy libssl-dev` and `apt-cache madison libssl-dev` both report only version 1.0.2 as available, so I have a feeling merging this would break ubuntu builds.

So, it seems to me that for whatever reason, libssl-dev=1.0.2 and openssl=1.0.2 work with Urbit on ubuntu 16.04 but not on debian 9 ... that doesn't fit exactly with my model of how these things should work, but I do think we should hold off on merging this change until we confirm that the install works on ubuntu this way. @cgyarvin this also indicates we might be better off just fixing the code to use the new version, as you suggested.

todlur:
@belisarius222 Weird, the whole reason I considered bringing this up was because someone mentioned that their libssl=1.0.0 on their Ubuntu machine running Urbit. I agree that adjusting the code is the simplest fix-all solution.

ngzax:
ooh, ooh, `urbit-0.5.0`? Is that available to be used or do we wait until the breach?

keatondunsford:
@ngzax Haha, wait for the breach!

keatondunsford:
@joemfb, does the Meson setup fix this issue?

joemfb:
@keatondunsford, I'm not sure. I think we'll need to test.

keatondunsford:
@eglaysher, do you know if this is an issue anymore?

keatondunsford:
I don't think we're hitting any issues with the latest iteration of our new Meson installs on different platforms with our pinned dependencies. So, I'm going to close this for now, and we can reopen it if we hit any dependency issues with OpenSSL.

PR #218, opened 2017-12-10
Update to Demo, Syntax, and Concepts docs

joshuareagan:
Minor revisions, clarifications, removing the words "twig" and "span" from the Syntax and Concepts docs.

### DIFF
--- a/docs/hoon/concepts.md
+++ b/docs/hoon/concepts.md
@@ -10,25 +10,19 @@ title: Concepts
 Usually, a new language is an improvement on one you already
 know.  If it isn't, it probably at least shares concepts with a
 large family of relatives.  These concepts come with words, like
-*type* or *function*, that map cleanly onto its semantics.
+*variable* or *function*, that map cleanly onto its semantics.
 
-In Hoon we throw away almost all these words and invent new
+In Hoon we throw away many of these words and invent new
 ones, which mean almost the same things.  But why?
 
 ## Motivation
 
 Hoon has no close relatives.  The normal jargon of programming
-maps poorly onto its semantics.  For example, although Hoon is a
-typed language, the term "type" is not formally defined in Hoon.
-"Type" is a great word, but Hoon has three separate concepts
-which could comfortably claim it.  We say "type" all the time; we
-mean [`span`](#span), or `mold`, or `mark`, or possibly all three.
-
-Other common programming concepts used only informally in Hoon
-are "function," "object," "event," "expression," "variable,"
-"label," "closure", "environment," "scope," and probably a few
-more.  We do use these words informally, but we're always making
-an analogy whose precision depends on context.
+maps poorly onto its semantics. Some of the common concepts used 
+only informally in Hoon include "function," "object," "event," 
+"variable," "label," "closure", "environment," "scope," and probably 
+a few more.  We do use these words informally, but we're always 
+making an analogy whose precision depends on context.
 
 Hoon has concepts *like* all these abstractions, but they remain
 *false cognates*.  The closer an inexact cognate, the more
@@ -38,7 +32,7 @@ a few new words is a small price for avoiding this pain point.
 ## Concepts
 
 A few major Hoon concepts: `noun` (data), `nock` (interpreter),
-[`mint`](#mint) (compiler), [`span`](#span) (type), [`twig`](#twig) (expression),
+[`mint`](#mint) (compiler), [`hoon`](#hoon) (AST node),
 `gate` (function), `mold` (constructor), `core` (object), `mark`
 (protocol).
 
@@ -53,20 +47,21 @@ The simplest noun syntax uses brackets, like `[a b]` for the cell
 of `a` and `b`.  Brackets nest right; `[a b c d]` is `[a [b [c
 d]]]`.  You'll also see braces: `{a b c d}`.
 
-A noun list, by convention, points right and is zero-terminated.
-Hoon uses tuples (improper nouns) more freely than Lisp; only
-genuinely iterative structures are null-terminated lists.
+A noun list, by convention, goes from left to right and is 
+null-terminated. That said, Hoon uses tuples (improper nouns) more 
+freely than Lisp; only genuinely iterative structures are 
+null-terminated lists.
 
 When an atom is used as a string (a `cord`), the bytes are always
-UTF-8 in LSB first order.  When you see `foo`, `$foo`, `%foo`,
+UTF-8 in LSB-first order.  When you see `foo`, `$foo`, `%foo`,
 etc, this refers to the integer `0x6f.6f66`, aka `7.303.014`.
-(Hoon syntax breaks integer atoms in the German style.)
+(Hoon syntax uses `.` to break integer atoms in the German style.)
 
 One common operation on nouns is `slot`, a tree addressing
 scheme which maps an atom to a subtree.  The whole noun is slot
 `1`; the left child of `n` is `2n`, the right child `2n+1`.
 
-Nothing in Nock, Hoon or Urbit can create cycles in a noun or
+Nothing in Nock, Hoon, or Urbit can create cycles in a noun or
 detect pointer equivalence.  Nouns are generally implemented
 with reference counting, and have a lazy short Merkle hash to
 help with equality testing and associative containers.
@@ -82,16 +77,54 @@ it's still fun and useful to [learn more](../../nock).
 
 ## Compilation concepts
 
+### <a name="hoon">`hoon`</a> (AST node)
+
+A [`hoon`](../reference) is the result of parsing a Hoon source 
+expression into an AST node. Because every Hoon programs is, 
+in its entirety, a single expression of Hoon, the result of parsing 
+a whole Hoon program into an AST is a single `hoon`.
+
+As the noun that the parser produces, a hoon is a tagged union of
+the form `[tag data]`, where the tag is a constant such as `%brts`
+which matches up with the appropriate type of data (often more 
+`hoon`s).  For example, the expression `:-(p q)`, once parsed into 
+an AST, becomes the tagged union:
+
+```
+[%clhp p=hoon q=hoon]
+```
+
+The %clhp is for "colhep".
+
+Keep in mind that `p` and `q` would be parsed too.  So if `p` is `2` 
+and `q` is 17, the parsed result is:
+
+```
+[%clhp p=[%sand p=%ud q=2] q=[%sand p=%ud q=17]]
+```
+
+To parse Hoon source into a hoon AST, use `ream` on a `cord`, e.g.:
+
+```
+(ream ':-(2 17)')
+```
+
+Try it in the `:dojo` to get:
+
+```
+[%clhp p=[%sand p=%ud q=2] q=[%sand p=%ud q=17]]
+```
+
 ### <a name="mint">`mint`</a> (compiler)
 
-[`mint`](#mint) is the Hoon compiler.  It maps a cell `{span twig}` to a
-cell `{span nock}`, where a [`span`](#span) is a type, a [`twig`](#twig)
-is a compiled expression (AST), and a `nock` is a Nock formula.  `mint` accepts
-a subject type and a source expression; it produces a product type
+[`mint`](#mint) is the Hoon compiler.  It maps a cell `[type hoon]` to a
+cell `[type nock]`, where a [`type`](#type) is a type, a [`hoon`](#hoon)
+is a parsed expression (AST), and a `nock` is a Nock formula.  `mint` accepts
+a subject type and a parsed source expression; it produces a product type
 and an executable formula.
 
 Calculating the output type from the input type and the source
-code is called "type inference." If you've used another typed
+code is called "type inference". If you've used another typed
 functional language, like Haskell, Hoon's type inference does the
 same job but with less intelligence.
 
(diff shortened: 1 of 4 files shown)


docs/hoon/concepts.md
```diff
@@ -82,16 +77,53 @@ it's still fun and useful to [learn more](../../nock).
 
 ## Compilation concepts
 
+### <a name="hoon">`hoon`</a> (AST node)
+
+A [`hoon`](../reference) is the result of parsing a Hoon source 
+expression into an AST node. Because every Hoon programs is,
```
belisarius222: programs -> program

docs/hoon/concepts.md
```diff
@@ -82,16 +77,53 @@ it's still fun and useful to [learn more](../../nock).
...
+a whole Hoon program into an AST is a single `hoon`.
+
+As the noun that the parser produces, a hoon is a tagged union of
+the form `[tag data]`, where the tag is a constant such as `%brts`,
+which is matched up with the appropriate type of data (often more 
+`hoon`s).  For example, the expression `:-(p q)`, once parsed into 
+an AST, becomes the tagged union:
+
+```
+[%clhp p=hoon q=hoon]
```
belisarius222: This is a bit confusing. The *type* of the parsed expression is a tagged union, which includes all the possible tags, not just the `%clhp`. You'd need a `?-` or equivalent to get the type to be expressed this way. Maybe mention that `++type` is a tagged union, and the `%clhp` subtype looks like this. The *value* of the parsed expression, as you explain below, has that form.

docs/hoon/concepts.md
```diff
@@ -82,16 +77,53 @@ it's still fun and useful to [learn more](../../nock).
...
+```
+
+The `%clhp` is for "colhep". Keep in mind that `p` and `q` would be 
+parsed too.  So if `p` is 2 and `q` is 17, the parsed result is:
+
+```
+[%clhp p=[%sand p=%ud q=2] q=[%sand p=%ud q=17]]
```
belisarius222: I think it's worth mentioning that `[%sand p=%ud q=2]` means a the type of a constant atom with value 2 and aura `@ud` (maybe we shouldn't mention auras yet). Could say something like:
```
Notice `p` and `q`, both of type `hoon`, have been parsed as `%sand`-tagged `hoon`s, which represent constant atoms.
```

docs/hoon/concepts.md
```diff
@@ -105,52 +137,21 @@ which makes your program more readable anyway.  Also, the dumber
...
-b
 ```
-or the tall rune syntax (here on one line, with double spaces):
-```
-%-  a  b
+(mint:ut %noun [%clhp p=[%sand p=%ud q=2] q=[%sand p=%ud q=17]])
 ```
 
-### <a name="span">`span`</a> (type, as range)
+The `%noun` is type information, and the tagged union after that
```
belisarius222: I don't actually know what the type is supposed to represent. It seems worth mentioning. Is it the inferred output type of the nock formula?

docs/hoon/concepts.md
```diff
@@ -105,52 +137,21 @@ which makes your program more readable anyway.  Also, the dumber
...
+is the hoon.
+
+### <a name="type">`type`</a> (type, as range)
 
-A `span` is a Hoon type.  It defines a set (finite or
-infinite) of nouns and ascribes some semantics to it.  There is
-no syntax for spans; they are always defined by inference (ie, by
+A `type` defines a set (finite or infinite) of nouns and ascribes 
+some semantics to it.  There is no direct syntax for defining 
+types; they are always defined by inference (ie, by
```
belisarius222: ie -> i.e.

docs/hoon/concepts.md
```diff
@@ -179,31 +180,32 @@ A `mold` is a constructor function (`gate`).  Its sample is any
...
+to validate untrusted network input.
 
 ### `face` (named variable)
 
 In a conventional language, we have a scope, environment or
 symbol table.  Declaring a variable, like `var foo: atom`, adds
 the name `foo` to the table with type `atom`.
 
-The Hoon equivalent is `:var  foo  atom`.   But Hoon has a
+The Hoon analogue is `=|(foo atom)`.   But Hoon has a
```
joemfb: This would be something like `=/  foo=atom` or `=/(foo=atom 1 ...)`. `=|(foo atom)` is "pin the bunt of *foo* and produce *atom*". But I'm not sure if the example is clarifying here ...

docs/hoon/concepts.md
belisarius222: I would also mention something about how the types are assembled out of base types defined in ++type. The type inference assembles complex types out of the simpler built-in types.

docs/hoon/concepts.md
```diff
@@ -179,31 +180,32 @@ A `mold` is a constructor function (`gate`).  Its sample is any
...
+a variable named `foo`, whose value is of type `atom`; it's a subtree 
+of type `foo:atom`.
 
 ### <a name="limb">`limb`</a> (attribute or variable reference)
 
 A [`limb`](../twig/limb/limb), like `foo`, is Hoon's equivalent of a variable
-reference.  A limb is a [`twig`](#twig); given a subject [`span`](#span),
-[`mint`](#mint) resolves it to a Nock formula and a product [`span`](#span).
+reference.  A limb is a [`hoon`](#hoon); given a subject [`type`](#type),
+[`mint`](#mint) resolves it to a Nock formula and a product [`type`](#type).
```
belisarius222: I think we should also mention that the hoon compiler always resolves a limb to a tree axis lookup, even though there are multiple kinds of limb. Since a face always refers to a specific tree axis within a member of a type, using a face as a limb still resolves to a numeric axis at compile-time.

docs/hoon/concepts.md
belisarius222: Good point. Another possibility would be `=|(foo=atom foo)`, which will print `0`.

docs/hoon/concepts.md
joshuareagan: I think I'm not understanding you here.  I'm saying that `[%clhp p=hoon q=hoon]` is a tagged union in the same sense that `17` is an `@` or an `atom`.  Yes, `@` is a mold which defines the type `atom`, but it's pretty clear that I don't mean `17` is a mold or a type.  In fact, it seems clear to me that `++hoon` isn't a tagged union, because it's a mold!  In the same way, `@` itself isn't an `atom`...

What am I missing?

docs/hoon/concepts.md
joshuareagan: I thought about that.  Honestly though I'm wondering if I've already gone too far by mentioning `%clhp`.  It's a little tutorial-ish for a definitional doc.  (I assume we want a tutorial/spec distinction in our docs.)  I think it's worth bending the standards a little to make our definitions clearer.  Examples are often helpful.  But I don't want to get too far into the weeds, as long as the general idea is clearish.

That said, saying a little to explain `%sand` seems like a fair suggestion, and I'm not particularly opposed to it if everyone likes the idea.

docs/hoon/concepts.md
joshuareagan: It's not that, because entering:

`(mint:ut %noun [%clhp p=[%sand p=%ud q=2] q=[%sand p=%ud q=17]])`

...results in:

`[#t/{@ud @ud} q=[%1 p=[2 17]]]`

That's more type info than we gave mint. I don't know what that first type represents.

docs/hoon/concepts.md
joshuareagan: I was just trying to match the pseudo-code `var foo: atom`, which defines foo without giving it a value.  We don't exactly have that, so I tried to give the closest approximation.  I'm open to ideas about how to make that section clearer and more helpful.

docs/hoon/concepts.md
joshuareagan: I made the change.

docs/hoon/concepts.md
joshuareagan: change made

docs/hoon/concepts.md
joemfb: `%noun` in your example is the type of the subject.

joshuareagan: Thanks Joe!

docs/hoon/concepts.md
belisarius222: The term "tagged union" usually refers to a type, not a value. So `++hoon` is a mold that normalizes to a tagged union type. When you say that `[%clhp p=hoon q=hoon]` is a tagged union, that is confusing to me because it's really just one subtype of the full tagged union type. To me, the full tagged union would be the `$%  [%clhp ...]  [%cnts ...] ... ==`.

Also, saying `[%clhp p=hoon q=hoon]` is a tagged union is different from saying that `17` is an `atom` because it's not an instance of the type the way `17` is an instance of `atom`. An instance of the type would look more like `[%clhp p=*hoon q=*hoon]`, or your later example with the `%sand`s.

Does that make sense?

docs/hoon/concepts.md
belisarius222: `=|` is indeed hoon's way to declare a variable with minimal initialization, so it is the closest approximation. I think a more generally useful concept to translate from other languages is `var foo: atom = 3`, though, which would correspond to `=/` or `=+`, and of those two `=/` is probably the better choice since it lets you express the type of the variable in tall form. Either way is fine, but it should be correct syntax, which requires having the third argument to `=|`. Maybe something like: `=|(foo=atom <sub-hoon>)`

keatondunsford:
Alright, looks like some edits were made in response to the comments, but that @joshuareagan you might still have a few more to go. Want to ping me when you think this is ready to merge?

docs/hoon/concepts.md
joshuareagan: There is something funky going on with language here.  I hope it's okay if I probe a bit to get a better sense of what's going wrong.  Pedantry ahead...

>  The term "tagged union" usually refers to a type, not a value.

Agreed!  In the same way, "integer" refers to a type, not a value; and we still say "15 is an integer".  This usage clearly means "15 is an instance of the type, integers".  Even though 15 is of a sub-type (say, the odd integers, or integers divisible by 5, etc.) of the full type of integers.

So it still seems okay to say (with a little imprecision):

"`[%clhp p=hoon q=hoon]` is a tagged union."

Oh wait, no it doesn't.  (I am/was wrong!)  Which tagged union type?  There are lots!  `++type`, `++hoon`, `++card`... these all (seem to) define distinct tagged union types.

The problem is that "tagged union" doesn't seem to be (simply) a type.  It's a *type* of type.  (Does this seem right?)

Accordingly, I *don't* think we should say that *this* is a tagged union either: `$% [%clhp ...] [%cnts ...] ... ==`.  That's because this is a mold, and a mold is a gate, not a type.  This mold *defines* the type for parsed Hoon ASTs, without itself *being* that type.  (`@` isn't a type, it's a mold! Speaking loosely sinks ships...? or it lets you avoid being an annoying pedant, or something....)

Here's what's confusing me though.  All tags look like this: `%foo`.  First a % and then some text (and/or integers).  "$foo" isn't a tag!  So the mold for a tagged union doesn't have tags.  The type defined by the `++hoon` mold is a set that contains the union of the tag-types `[%brts (something)]`, `[%clhp (something)]`, etc.

This is one case where I can't blame Tlon for the confusing vocabulary.  A "tagged union" is a union that isn't itself tagged, but which contains pairs in which the LHS is tagged.

Does this all seem right?

In this case, what's the handy-dandy way to refer to an instance of the `++hoon` tagged union type?  In gloriously ambiguous (but probably totally safe) fashion, do we call it a `hoon`?

Thanks in advance for providing me a CS education without compensation.

joshuareagan: Despite my disclaimer about pedantry, the above post is rather sloppy... if it's not obvious what is meant, don't bother slogging through it, just yell and I'll try again...

joemfb: A type can be a tagged-union (ie, tagged-union is a sub-category of the category type). A value can be an instance of a type.

We say that 15 *is* an integer, but we don't say that 15 *is* a numeric type -- the relationship is further attenuated. Similarly, a given noun may be an instance of a type that is a tagged-union, but saying the noun *is* a tagged-union is overly dis-intermediated.

ohAitch:
- $foo is going away as syntax, precisely because the distinction between
it and %foo is pedantic more frequently than helpful
- $%([%& a] [%| b]) is definitely a tagged union between a and b: the %&
and %| are the const tags, and are consed on so you can in practice
distinguish whether you have an a or a b

docs/hoon/concepts.md
joshuareagan: That sounds right Joe.

joshuareagan:
Yes, I see what you mean about $foo and %foo.  I think that's what got me.

belisarius222:
Josh, this mostly seems right to me. My only quibble would be that
typically a tagged union is a pair of a tag and a union of subtypes. Hoon
is no different, except that you group each tag-and-subtype pair when you
define the tagged union (as opposed to, say, C, where you'd have a struct
with the tag type as the first element and the union type as the second
element). `$foo` is a mold that normalizes to the tag `%foo`, so I suppose
it's slightly sloppy to say `$foo` *is* a tag, but the two are so closely
related that we often conflate them when discussing Hoon informally, and
the hoon 143 syntax unification makes the two things look exactly the same
in most cases. The transformation between value and mold that normalizes to
the value is performed opaquely by the compiler, but that's done cleverly
so that you can define something that looks like an example of the type,
and the compiler will generate the normalizing mold for you. We also
typically use another shorthand of saying a mold "is" a type. Outside of
these docs, that doesn't usually cause issues, because there's no way of
defining a type except for molds, and molds aren't used for anything except
for defining types. I'm not sure if that completely addresses your post, so
please let me know if it didn't.

Yes, there are many types that could be classified as tagged unions. I
agree with @joemfb's explanation.

joshuareagan:
Ted, that seems clear and right to me.  Thanks!

docs/hoon/concepts.md
joshuareagan: Turns out `%noun` isn't the subject type, but rather it's type info against which the product type is type-checked.

docs/hoon/concepts.md
joshuareagan: Change made

joshuareagan:
Okay, I'm ready for another round.

ohAitch:
hmm, I think you just want to use (~(mint ut %noun) %noun foo), since the
whole point of the section was transitioning from "text can be parsed into
an ast" into "the ast is then compiled to nock"

joshuareagan:
I'm conflicted.  On the one hand, you're right, why talk about parsing if we don't show how ++ride uses that to compile?

On the other hand, if we use (~(mint ut %noun) %noun foo), then we have to explain the (~(...) ... ...) syntax, we have to talk about both %nouns, and it's just a less friendly exposition.  Lots of complications.  Part of what I hope to achieve is helping make the docs approachable enough that charges of Hoon's inscrutability won't stick.

People need to know about parsed `hoon`s in any case, because they're in `++type` and in many other places, so it's not necessarily bad if we use `++ride` to talk about the compiler.  We can mention in passing how `++ride` uses `++ream` and `++mint` to work.

ohAitch:
Honestly your best bet might be ++slap

joshuareagan:
I decided to go with the `(~(mint ut %noun) %noun <hoon>)` version. 
It's really not that complicated, and people need to learn how to call 
inside cores sooner or later.

Anything else?

docs/hoon/concepts.md
```diff
@@ -53,20 +47,21 @@ The simplest noun syntax uses brackets, like `[a b]` for the cell
...
 etc, this refers to the integer `0x6f.6f66`, aka `7.303.014`.
-(Hoon syntax breaks integer atoms in the German style.)
+(Hoon syntax uses `.` to break integer atoms in the German style.)
 
 One common operation on nouns is `slot`, a tree addressing
 scheme which maps an atom to a subtree.  The whole noun is slot
 `1`; the left child of `n` is `2n`, the right child `2n+1`.
 
-Nothing in Nock, Hoon or Urbit can create cycles in a noun or
+Nothing in Nock, Hoon, or Urbit can create cycles in a noun or
```
keatondunsford: @joshuareagan @belisarius222 We should probably file an issue for this -- whether or not official Urbit documentation should use the Oxford comma.

belisarius222: The Oxford comma is recommended by good people, people who care about precision and programmers.

PR #220, opened 2017-12-11
Updating the Basic Types Doc

joshuareagan:
The revisions to the first half were somewhat more heavy-handed and may be controversial.  Happy to discuss.

### DIFF
--- a/docs/hoon/basic.md
+++ b/docs/hoon/basic.md
@@ -7,85 +7,125 @@ title: Basic types
 
 # Basic types
 
-A `span` is a set of nouns and an interpretation of these nouns.
-Officially, we don't use the word "type" in Hoon.  But if we had
-to say "type," it would mean `span`.
+A Hoon `type` is a set of nouns. There is no syntax for directly 
+defining the data type of a Hoon expression. Instead, the compiler 
+infers the type of a Hoon expression by calculating the range of 
+possible values of that expression.
 
-## Type, span and mold
+Working with a type in Hoon requires the use of a `mold`. Each 
+type is defined by a mold. To define a custom type, you must create 
+a custom mold.
 
-There is no Hoon syntax for a span.  The programmer never defines
-a span explicitly.  It is always produced as the inferred range
-of an expression (`twig`).
+## Molds
 
-But we still need simple, well-formed expressions that produce
-regular and well-shaped ranges, for three reasons.
+A mathematical function is a mapping from one set (the domain) to 
+another (the range). Pass an object from the domain to some 
+function *f*, and *f* returns an object from the range. In Hoon we 
+can think of a `gate` as a function, where the gate's `sample` is 
+its argument.
 
-First: in most cases, this trivial generator should be much
-simpler than the computation itself.  Casting the actual result
-to its ideal shape makes sure we know what we're building.
+A `mold` is an idempotent function (i.e., gate) that takes any noun 
+as its input (i.e., its domain is the set of all nouns). An 
+idempotent function *f* is one such that, if *f(x)* terminates, 
+*f(x) = f(f(x))*.  The product range of a mold is the type defined 
+by that mold. That is, each mold's type is the set of possible 
+outputs of that mold.
 
-Second: we can define a standard form for such generators, and
-the standard form is useful.  The standard form is a constructor
-function, or *mold*.
+Essentially, molds are constructor functions. You can pass a mold 
+any data you like, and the product is guaranteed be of the type 
+defined by that mold. Molds are the perfect tool for validating 
+untrusted foreign data: just call the relevant mold with that data. 
+But this use case is unusual for beginners; otherwise you shouldn't 
+call molds directly.
 
-A mold is an idempotent function (`gate`), accepting any noun.
-(An idempotent function is one such that *f(f(x))* equals *f(x)*
-if *f(x)* terminates.)  The product range of the function is the
-span, or *icon*, of the mold.
+More frequently, molds are used when we want to `cast`. We use a 
+cast when we want the Hoon compiler's type-checker to test whether 
+a Hoon expression is guaranteed to evaluate to a product of the 
+desired type. If it isn't, then the compiler will halt with a 
+`nest-fail` crash. Proper Hoon code will include a cast for each
+gate.
 
-Usually we use molds purely in the first sense: as an abstract
-definition of a noun.  Don't actually call a mold unless you're
-actually validating untrusted foreign data.  As a beginner,
-hopefully you aren't!
+The `^` family of runes is for casting. In particular, `^-` allows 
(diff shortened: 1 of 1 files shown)


keatondunsford:
Looks better than what we have.

docs/hoon/basic.md
```diff
@@ -7,85 +7,125 @@ title: Basic types
...
-Officially, we don't use the word "type" in Hoon.  But if we had
-to say "type," it would mean `span`.
+A Hoon `type` is a set of nouns. There is no syntax for directly 
+defining the data type of a Hoon expression. Instead, the compiler 
+infers the type of a Hoon expression by calculating the range of 
+possible values of that expression.
+
+Working with a type in Hoon requires the use of a `mold`. Each 
+type is defined by a mold. To define a custom type, you must create 
+a custom mold.
```
joemfb: A think this language is over constrained. While it's certainly the typical pattern, "requires" and "must" aren't justified, IMO.

docs/hoon/basic.md
```diff
@@ -7,85 +7,125 @@ title: Basic types
...
-Usually we use molds purely in the first sense: as an abstract
-definition of a noun.  Don't actually call a mold unless you're
-actually validating untrusted foreign data.  As a beginner,
-hopefully you aren't!
+## `type`: a set of types
 
-## `span`: a set of nouns
+Below is the mold for `type`, written in Hoon. This mold's range is 
+the set of all types. (Don't let yourself get too dizzy thinking 
+about it! Strictly speaking, the range is of trees of type-*labels*
```
joemfb: I don't find this speaking to be enough strictly

joshuareagan: "range is of trees of type-labels"

If this isn't strict enough, then there must be a counterexample... i.e., a product of ++type that isn't such a tree.  I don't think that's the case.

joshuareagan: "range is of trees of type-labels"

If this isn't strict enough, then there must be a counterexample... i.e., either a product of ++type that isn't such a tree, or such a tree (well-formed) that cannot be a product of ++type
e. I don't think that's the case.

docs/hoon/basic.md
joshuareagan: Earnest question, how else may we use types?  I hesitated when writing this, because of `^+`, but that only works because there is a mold for the example that `^+` uses.  So while it's a little indirect, a mold is indeed needed there.

docs/hoon/basic.md
joemfb: Is this noun a label?

```
[%core #t/<1.xxy $~> q=[p=%gold #t/<1.xxy $~> r=[p=~ q=~] s=[p=[1 2] q={[p=0 q=[p=[p=~ q=~] q={[p=%$ q=[p=~ q=[%ash p=[%sand p=%ud q=2]]]]}]]}]]]
```

It's the type of the following:

```
=>(~ =>(|.(1) |.(2)))
```

joshuareagan: It's a *tree* of such labels, if I understand it correctly (slightly dubious about the latter).

keatondunsford:
@joshuareagan If you agree with these comments, feel free to make another PR with more changes to this `basic` doc, and I'll merge it!

docs/hoon/basic.md
joemfb: Assuming that noun has the face `typ`,

- `p.typ` is the type of the payload
- `p.q.typ` is the variance
- `q.q.typ` is the type we compiled against
- `r.q.typ` is labeled documentation
- `p.s.q.typ` is the battery
- `q.s.q.typ` is a map of core chapters, wherein each chapter is a pair of documentation and a map from arm name to "foot" (arm genericity and arm hoon).

I feel that stretches the definition of label, as I understand it.

joemfb: There are lots of symbols in that nouns, but the only ones that are labels/tags from the tagged-union which is `type` are 1) the entire noun, 2) the payload, 3), the compilation type. Everywhere else those symbols/labels/tags serve another purpose (for instance, to discriminate between forks in other types).

In short, my point is that a noun that is an instance of `type` may be made up of a lot more than just some labels. Hoon ASTs, Nock formulas, documentation, etc. can all be present.

belisarius222:
Label and parameters, maybe?

docs/hoon/basic.md
joshuareagan: "I feel that stretches the definition of label, as I understand it."

Hence the emphasis on the word "tree"!

joemfb: It's definitely a *tree*, and there are definitely *labels*. My argument is that it is much more than simply a *tree* of *labels*, unless you generalize the definition of label to encompass any reference to or description of something.

joshuareagan: @joemfb, I'm having trouble with your objection to my definition, because it assumes from the start that the definition should be different.  You write that this is a type:

> [%core #t/<1.xxy $~> q=[p=%gold #t/<1.xxy $~> r=[p=~ q=~] s=[p=[1 2] q={[p=0 q=[p=[p=~ q=~] q={[p=%$ q=[p=~ q=[%ash p=[%sand p=%ud q=2]]]]}]]}]]]

But on my account, it isn't.  On my account, it's a tree that tracks a type.  I certainly didn't say that this is a label!

To be fair to you, I did say "tree of labels", and that structure may be too complex to count as a tree of labels.  But in my defense, I am writing for the basic type docs, which are intentionally over-simplified compared to the advanced type docs....  It's fair to say that I can revise the "tree of labels" thing so that it's more accurate.  But what's important on my account is that this tree is an accounting device for the compiler, not itself a type.

This shift is part of my crusade to distinguish the semantics that a Hoon beginner needs to learn vs. the implementation of the Hoon compiler.  This is a good crusade and I'll continue to wage it.

joemfb: I share your goals, and commend you for your efforts. Quality introductory documentation is very important! But I think there some imprecisions here that go too far.

You say my example is not a "type", but "a tree that tracks a type". I'm not sure I can parse that.

Until very recently, my example noun would've been called a "span" -- the inferred type produced by the compiler. "A data structure to track the type of an expression in the compiler" would be a perfectly suitable definition. But "span" has been renamed to "type", and that noun is very much *the type* of a trivial core.

This all goes back to the following:

> Strictly speaking, the range is of trees of type-*labels*

`s/roughly/strictly/` (or something like that), and I'll go away ;)

joshuareagan: Also, we need to distinguish the result that the pretty-printer gives you, and the actual tree used by the compiler.

mattlevan: `s/strictly/roughly` and it should be good. If you really want to be more precise, say "tree that tracks a type and contains labels." The reader can infer that more than labels may exist in the tree, but he's not yet worried about it.

PR #231, opened 2018-02-10
Fixed small layout problem in ++sear doc (4f.md).

jfranklin9000:
*Before*
=

sear-before

*After*
=

sear-after


Fang-:
Good catch, thanks!

PR #232, opened 2018-02-12
Straighten out the Accepts and Produces sections of docs/hoon/library…

jfranklin9000:
…/4m.md.

All of the Accepts sections were empty and the Produces sections had what should have been in Accepts so I moved them there and filled out the Produces sections.

I'm not sure about `++slav`, `++slaw`, `++slay` and `++stab`.

### DIFF
--- a/docs/hoon/library/4m.md
+++ b/docs/hoon/library/4m.md
@@ -13,10 +13,12 @@ Renders a dime `mol` as a cord.
 Accepts
 -------
 
+`mol` is a `++dime`.
+
 Produces
 --------
 
-`mol` is a `++dime`.
+A `cord`.
 
 Source
 ------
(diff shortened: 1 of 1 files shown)


joemfb:
`++slay` parses `txt` (a cord) for any atom, producing `(unit coin)`.

`++slaw` parses `txt` (a cord) for an atom with an aura of `mod`, producing `(unit @)`. (it only works with `++dime`, not any `++coin`)

`++slav` is `(cork slaw need)`; it produces the atom from `++slaw`, crashing if null.

`++stab` parses a cord for a path, producing a path or crashing if it can't.

BTW, this doc consistently says "parse span" where it should say "parse a cord".

jfranklin9000:
I'll incorporate these changes and resubmit.  But "parse span" should be changed to "parse a knot" (except for `++stab` which does indeed parse a `++cord`).

However, since a `++knot` (`@ta`) is a `++cord` (`@t`) should I change them to "cord"?

cgyarvin:
Might as well stick with "cord" for all of them.  But it doesn't matter
that much!

ohAitch:
I mean, they do parse a knot, like if you give them a non-knot cord they'll
error cleanly but anything that's a valid slay/slaw/slav atom-as-text
format is valid @ta

jfranklin9000:
I've incorporated ~master-morzod's input to the PR (except the `++dime` info; didn't know how to express that).  I used `cord` instead of `knot`.

PR #236, opened 2018-03-07
changing `pave` for `scow`

joshuareagan:


ohAitch:
That's fair, like you might want `<count>` but definitely not the explicit `!>` version

joshuareagan:
I remember being told that `< >` isn't really intended for normal use, but is rather a debugging tool of sorts.  Maybe I misunderstood or I'm misremembering, but I think it's syntactically more elegant and easier to remember than `scow`.  Is there any reason not to use the angled brackets here?

ohAitch:
It's like equivalent to the pave version, which is slower than %scow
because it is effectively a dynamically typed print. I think it makes sense
to avoid it in serious code, but in beginning examples I think being in
your face about the concept of %ud (or !> ofc!) adds too much confusion.

joshuareagan:
Thanks, that makes sense.  I'll keep that in mind in the future.

PR #237, opened 2018-03-08
Make example planet + ticket more obviously fake.

Fang-:
This caused some confusion, someone tried running the stated command as-is.

Though it's really not a huge issue, I see no reason not to use a shipname and ticket that is obviously fake.


ohAitch:
I am reminded of https://www.snopes.com/fact-check/the-woolworth-card/

PR #240, opened 2018-04-03
Change to meson

eglaysher:
This is a first draft of changes to the install documentation for meson. It works for brew based systems, but macports and all the other package managers still need to be updated.

### DIFF
--- a/docs/using/install.md
+++ b/docs/using/install.md
@@ -52,7 +52,7 @@ Which can usually be installed with the following one-liners:
 ```
 # macOS (Homebrew) https://treehouse.github.io/installation-guides/mac/homebrew
 # Linuxbrew http://linuxbrew.sh/
-$ brew install autoconf automake cmake gcc git gmp libsigsegv libtool python2 openssl
+$ brew install autoconf automake cmake gcc git gmp libsigsegv libtool meson ninja pkg-config python2 openssl re2c
 
 # macOS (Macports)
 $ sudo port install autoconf automake cmake gmp libsigsegv openssl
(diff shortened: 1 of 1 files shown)


eglaysher:
The [Filesystem Hierarchy Standard](https://wiki.debian.org/FilesystemHierarchyStandard) says that locally compiled binaries go in `/usr/local/bin`. Or: yes, people are using /usr/local for urbit.

mikolajpp:
Well, actually (;-), this is about preventing clashes between in-house software and distro supplied. Since we urbit is actually the only bundle we provide, it could safely go into /usr. /usr/local is for these cases where there could be clashes between distro provided software and locally installed one. I for one install my urbit into /usr.

Regardless of the fact where we put the fact, one thing is missing:
the libraries against which urbit is compiled should also be installed: after all we did provide all the fallback libraries. 

So the proper installation is to actually configure the install path and install using `ninja` install command.

joemfb:
BTW, I just confirmed that you don't have to explicitly `git submodule init` and update, meson will handle that for you.

keatondunsford:
@joemfb That's cool. Probably doesn't hurt to just keep it in the script though, unless you want to take it out.

mikolajpp:
@joemfb you may want to see my PR to your branch.

`~palfun-foslup` needed compiling h2o, so I added meson.build as well as subprojects basing off your branch.

mikolajpp:
Also, for some reason @joemfb your branch has missing flags (funsigned-char) which caused that recent bug.

joemfb:
I was hoping we could just remove the bootstrap script, but I tested on Ubuntu Xenial, and legacy meson does not handle submodules automatically.

keatondunsford:
I updated the Install page, and I was able to boot a fake ~zod with the Meson-PR branch. If we hit any issues, we can deal with them -- but, this all looks good to me. Merging!

keatondunsford:
Wait, don't we want to install Meson through `pip` to get the latest version?

PR #241, opened 2018-04-03
Show users how to find ship name

rmariani:
This mostly matters for those who are running a comet or a moon. Can't hurt to add this.


rmariani:
Typo; committed a better change.

PR #242, opened 2018-04-04
fixed dead links

rmariani:
in a few locations it looks like there were URLs with an old scheme


keatondunsford:
Good catch -- thought I fixed these all.

PR #244, opened 2018-04-05
removed confusing path concatenation

rmariani:
Readers are given mixed signals by trying to start the app from the home desk without putting the `%examples` path in. By following these docs, readers will not have an `app/echo.hoon` under the home desk.

Also, out of nowhere comes `:examples-echo` right after an `:echo` command. I'm not sure why this is. I think my changes make things clearer. I also fixed /twig dead links

### DIFF
--- a/docs/arvo/network.md
+++ b/docs/arvo/network.md
@@ -39,12 +39,12 @@ a value, it prints that value out. To try this out, you have to start
 the app, then you can poke it from the command line with the following
 commands:
 
-    ~fintud-macrep:dojo> |start %echo
+    ~fintud-macrep:dojo> |start %examples %echo
     >=
     ~fintud-macrep:dojo> :echo 5
     [%echo %noun 5]
     >=
-    ~fintud-macrep:dojo> :examples-echo [1 2]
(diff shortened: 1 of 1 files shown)


keatondunsford:
@Fang- @ohAitch Does starting apps on other desks actually work?

Fang-:
Confirmed it works, start with `|start %desk %app` and then poke as `:app ...`.

I think the `:examples-echo` may be a leftover from when that was in `/app/examples/echo.hoon`.

keatondunsford:
It seems like there's a potential namespacing issue here. Shouldn't one be able to poke an app on a specific desk? Perhaps there is an existing Arvo issue for this, but if not, we should file it.

Fang-:
Practically, currently, there's no need. If you try to start an app from two different desks you hit urbit/arvo#542.

In the short-term this should just error. In the long term, I'm not even sure if support was ever intended. Note how you can't specify a desk when poking an app on a foreign ship. But this is a whole different discussion.

keatondunsford:
Mkay. Well, the syntax of this PR seems more explicit, which is better. So, I'm going to merge this.

Thanks @rmariani.

PR #245, opened 2018-04-06
= instead of /

rmariani:


keatondunsford:
This would be even better if:

```
`{^ {{arg=TYPE $~} $~}}`
```

was changed to:

```
`[^ [[arg=TYPE ~] ~]]`
```

which is the latest desired syntax for Hoon 143. There are probably many cases where this old syntax is still used. 

I would personally focus more on trying to understand the content than I would trying to salvage these docs, which probably all need to get rewritten.

PR #253, opened 2018-04-23
Update interface.md

rmariani:
Links were broken

### DIFF
--- a/docs/hall/architecture.md
+++ b/docs/hall/architecture.md
@@ -10,7 +10,7 @@ title: Hall architecture
 
 This document is complemented by Hall's source code, but doesn't require it. Definitions of data structures, code snippets and diagrams will be provided where useful.
 
-Hall's implementation is structured according to the new Gall model. Familiarity with its concepts is assumed. See [the new Gall model](new-gall).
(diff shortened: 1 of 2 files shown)


keatondunsford:
Can you change these to the relative path (like the current link), so that these can still be viewed on localhost? E.g. if I'm viewing a copy of the docs on my local ship, they should point to my local Clay file, rather than redirect me to urbit.org. The current relative link system used here and throughout the docs works, if you set them up that way.

rmariani:
Sure, I just thought that the issue might have been the relative paths

keatondunsford:
Try `[new gall](new-gall))` and `#[Hall's architecture](architecture)`.

keatondunsford:
Cool. Can you also rebase the changes on your branch against origin/master to squash this into a single commit? It's an easy thing that you can Google how to do. Something like:

```
git rebase -i origin/master
```

Then change the left letters to 'f' to squash them and throw away their commit/message history, to boil these down into the first commit to be made to current origin/master.

Let me know if that doesn't make sense.

keatondunsford:
Closed in favor of #254 (helped Rob with this).

PR #256, opened 2018-05-04
simplifying demo.md

joshuareagan:
The current version uses too many new concepts, uses a standard library function when `< >` will do, uses the phonetic rune naming before it's been introduced, and had some overlong explanations.  The goal is just to build newbie confidence that they can understand Hoon fizzbuzz, so we want to lighten the cognitive burden in any way we can.

### DIFF
--- a/docs/hoon/demo.md
+++ b/docs/hoon/demo.md
@@ -7,12 +7,13 @@ title: Demo
 
 # Demo
 
-Here's everyone's favorite whiteboard example, FizzBuzz.  Each
+Here's everyone's favorite whiteboard example, 
+[FizzBuzz](https://en.wikipedia.org/wiki/Fizz_buzz).  Each 
 line is commented with its number:
 
 ```
-|=  end=@ud                                             ::  1
-=/  count=@ud  1                                        ::  2
+|=  end=@                                               ::  1
+=/  count=@  1                                          ::  2
 |-                                                      ::  3
 ^-  (list tape)                                         ::  4
 ?:  =(end count)                                        ::  5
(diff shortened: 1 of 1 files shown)


docs/hoon/demo.md
```diff
@@ -24,65 +25,60 @@ line is commented with its number:
...
+the type `@`, an atom (i.e., an unsigned integer).
 
-Line 2 uses the rune `=/` (pronounced "tis-fas") to declare a variable,
-`count` of the unsigned decimal number type, with initial value `1`.
+Line 2 uses `=/` to declare a variable, `count`, as an atom with an 
+initial value of `1`.
 
-Line 3 uses the rune `|-` ("bar-hep") to begin a loop.
+Line 3 uses `|-` to indicate the starting point of a loop 
+(technically a recursion).
```
joemfb: I'd lose the parenthetical. Technically, it's an immediately invoked `|.`, or one-armed core without a sample. The "recursion" is the re-evaluation of the arm of that core, which occurs on line 15.

docs/hoon/demo.md
```diff
@@ -24,65 +25,60 @@ line is commented with its number:
...
-`end`. This will produce either `%.y` (yes, the number 0) or `%.n`
-(no, 1) and consequently compute the "then" or "else" clause
-respectively.
+Line 5 uses the rune `?:` which is an ordinary "if-then-else" 
+conditional.  If `=(end count)` evaluates as true, then line 6 is 
+evaluated next.  Otherwise line 7 and on is evaluated.
 
-If the values of `count` and `end` are equal, line 6 (and the whole loop)
-returns the value `~`. The list is finished.  
+If the values of `count` and `end` are equal then line 6 returns the
```
joemfb: Here and prior, are we comfortable use "return" and "produce" interchangeably in introductory docs?

docs/hoon/demo.md
joshuareagan: ok, dropped

docs/hoon/demo.md
joshuareagan: I am because this particular doc is meant to be more intuitive than rigorous, but I don't see any reason not to switch to "produce" here.

docs/hoon/demo.md
```diff
@@ -24,65 +25,60 @@ line is commented with its number:
...
-strings. (Of course, if you run this code from the Urbit
-[_Dojo_](../../using/shell) (shell) you will be able to see the
-product.)
+When FizzBuzz is written in non-functional languages the 
+output is generated as a side effect of the program, e.g., with 
+`printf()`.  Hoon is a functional language so technically it 
+has no side effects.  Accordingly, the code above *returns* a 
+list of strings, it doesn't print them. (If you run it from the 
+Urbit shell, [_Dojo_](../../using/shell), you will of course be 
+able to see the product.)
```
joemfb: *returns/produces nitpicking intensifies*

joshuareagan: Tbh I'm not clear on the anti-"return" crusade.  What's actually wrong with it?  Functions return values; it's common, intuitive language.

keatondunsford: Produces is more functional.

joshuareagan: That's just not true.  The word "returns" is ubiquitous and far more common than "produces" in discussions of functions.

https://en.wikipedia.org/wiki/Function_(mathematics)
https://en.wikipedia.org/wiki/Function_composition_%28computer_science%29

joemfb: It could be that "return" suggests the possibility of the absence of a return (as in void functions in an imperative language). Honestly, I've been crusading so long I don't even remember.

@cgyarvin?

joshuareagan: The only reason I'm pushing on this (it's not that big a deal really) is that lots of people are coming from elsewhere having used the word "return" (either in math or software dev).  If we scrupulously avoid the word "return", it suggests that "produce" means something importantly different, when as far as I can tell it doesn't.  If there's some technical reason to avoid it, fine, but technical reasons don't apply as much to this early Hoon outreach doc, which should have intuitive language.

keatondunsford:
Okay, since "produce" was still used elsewhere throughout the doc, I changed the original offensive "return" case back to "produce", and we can bikeshed that later with @cgyarvin if we want.

Additionally, I made a couple copyedits (formatting and grammar).

Lastly -- mostly an opinionated matter of style and orderliness, but something that _really_ grinds my gears when I see it, @joshuareagan @rmariani (recently we've been finding out at the office that public shaming for cleanliness actually works) -- I cleaned up this:

screen shot 2018-05-10 at 11 38 44 am

To this:

screen shot 2018-05-10 at 11 53 59 am

This can be done with a little [Git rebasing](https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history), and it just makes the log so much cleaner and easier to read for everyone as we collaborate. This is surely a desirable art to master!

Merging! Thanks again guys.

keatondunsford:
Hm, you're right, that may have been a little fast on my part @joshuareagan. I assumed that, as reviewer, I had the privilege to make the edits necessary to make this PR mergeable. I viewed the bulk of the "simplification" work being definitely mergeable, whereas this pedagogical issue was more contentious. I'd be happy to continue the conversation about that point in a specific issue or PR, because you're right, it's important.

joshuareagan:
No problem Keaton!  Your actions are understandable and are certainly not being impugned.  It's just a plan for the future.

PR #260, opened 2018-05-08
corrected sieve example

joshuareagan:
`+-  tap` is no longer a gate, old code caused sieve.hoon not to compile.


keatondunsford:
This looks good to me -- just tested this in Dojo, and it works. Good work!

PR #153, opened 2017-02-19
Correct syntax and signature of %~ :open

hoclun-rigsep:

PR #262, opened 2018-05-10
Update doc to correct signature of %~ %cnsg

keatondunsford:
See #153 and urbit/arvo#698.

### DIFF
--- a/docs/hoon/rune/cen/sig.md
+++ b/docs/hoon/rune/cen/sig.md
@@ -7,7 +7,7 @@ title: %~ "censig"
 
 # `%~ "censig"`
 
-`[%cnsg p=wing q=hoon r=hoon]`: call with multi-armed door.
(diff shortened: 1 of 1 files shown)


joshuareagan:
These snippets of pseudo-parsed hoons in the rune docs are a bit strange; many don't reflect the actual products of the parser.  And why not?  Many seem to be modified with the intention of illustrating the correct syntax of the rune in question.

I think there's value in talking about the product of the parser for each rune, and I think there's value in explaining the correct syntax.  But I don't think there's much value in trying to do both at the same time.  The neither-fish-nor-fowl character of most of the pseudo-parsed hoons is a bad thing.

Keaton's revision shows the genuine product of the parser, and I think it's fine to have that somewhere in the rune definition (albeit in a less prominent place in the doc).  So I'm okay with merging this, and any other PR that makes the same sort of revision.  We'll figure out something else for illustrating expression syntax.  Once we have a better rune doc schematic we'll make those changes.

PR #263, opened 2018-05-11
replace gap with ace between foo and |=

deelawn:
it won't work in the dojo with two spaces between foo and |=


joshuareagan:
Looks good, dojo only takes `ace` between face and value.

PR #264, opened 2018-05-12
Updated syntax for 1a

rmariani:


joshuareagan:
In the future, let's have all changes to the technical reference docs (runes or stdlib) read by two reviewers.  This isn't because we don't trust Rob (or each other)---but because there's a lot of nit-picky stuff to get right (or wrong) in these things.  If something is missed in the submission, then hopefully someone else will catch it.  We need to make sure (1) that the info is right, (2) that no outdated terminology is used, to include previous gates that no longer exist, (3) that all the examples given still compile and give the stated answer, and (4) that all old syntax is updated to the new stuff.  This is a lot!  Reviewers should be carefully double-checking all this stuff; attention to detail is important here.

@rmariani Are you able to request reviewers?  If so, default to using me and @keatondunsford so we can get in the habit of reading through these regularly.  If you can't request reviewers, let me know and I'll try to get you permission to do that.

joemfb:
It looks like `+dec` is in here twice.

rmariani:
Yeah, I can request reviewers. I'll keep all that in mind.

docs/hoon/library/1a.md
```diff
@@ -64,7 +64,469 @@ Source
...
+
+An atom.
+
+Source
+------
+
+    ++  div                                                 ::  divide
+      ~/  %div
+      |:  [a=`@`1 b=`@`1]
+      :>  quotient
```
joshuareagan: We need to figure out whether to include doccord lines in the `hoon.hoon` code or not.  Check with the infrastructure team about whether to include these.  Either include them all, or get rid of the `:>  quotient` here.

docs/hoon/library/1a.md
```diff
@@ -64,7 +64,469 @@ Source
...
+    > (div 20 30)
+    0
+
+
+***
+### `++gte`
+
+Greater-than/equal
+
+Tests whether `a` is greater than a number `b`.
```
joshuareagan: "greater than or equal to a....

rmariani:
Alright, I seem to have jumped the gun on this one. Will feel things out before opening.

joshuareagan:
No, it's okay to leave this open.  I'm just going to nitpick a bit so you have an idea of what we're aiming for.

joshuareagan:
My last comment, is: let's add an entry for `++dvr`, since it's in that part of `hoon.hoon`.  It's straightforward division with remainder, so that shouldn't be tough.

docs/hoon/library/1a.md
```diff
@@ -64,7 +64,469 @@ Source
...
       ~/  %dec
-      |=  a/@
+      |=  a=@
+      ~_  leaf+"decrement-underflow"
+      ?<  =(0 a)
+      =+  b=0
+      |-  ^-  @
+      ?:  =(a +(b))  b
+      $(b +(b))
+
```
belisarius222: Blank lines should be replaced with `::`

docs/hoon/library/1a.md
belisarius222: Doccords are going away, to be replaced with leading comments, like this:
```
::  +some-arm: one-line description, all lower-case, no trailing period
::
++  some-arm  ...
::  +another-arm: one-line description
::
::    Long-form comments have four leading spaces and
::    use complete sentences. One blank comment line follows.
::
++  another-arm
  ::  some-face: a face declaration
  ::
  =/  some-face=@  42
  ...
```

belisarius222: For more details on this style, see the [ford-turbo branch](https://github.com/urbit/arvo/blob/ford-turbo/lib/ford-turbo.hoon).

docs/hoon/library/1a.md
```diff
@@ -64,7 +64,469 @@ Source
...
+`a` is an atom.
+
+`b` is an atom.
+
+Produces
+--------
+
+A boolean.
+
+Source
```
belisarius222: We shouldn't duplicate the source; instead, prepend documentation to the actual source.

belisarius222: Oh. I had a moment of delirium and for some reason thought this was in hoon.hoon. I think we do want to eventually move to a literate programming style where these docs would be generated from comments in hoon.hoon itself, but I guess that isn't happening yet.

My apologies for spouting irrelevant noise at you guys.

joshuareagan: Don't be too hard on yourself, this is definitely relevant noise!

docs/hoon/library/1a.md
```diff
@@ -64,7 +64,469 @@ Source
...
+
+
+Examples
+--------
+
+    > (gth 'd' 'c')
+    %.y
+    > (gth ~h1 ~m61)
+    %.n
+
```
keatondunsford: These are good examples that show the flexibility of atoms, but there should be a third and fourth -- the first ones to read -- that just use raw atoms. Something like:

```
> (gth 42 41)
%.y

> (gth 42 43)
%.n

> (gth 'd' 'c')
%.y

> (gth ~h1 ~m61)
%.n
```

rmariani:
Good ideas, I will make those changes.

—
~lodleb-ritrul
https://urbit.org

docs/hoon/library/1a.md
```diff
@@ -25,23 +25,37 @@ An atom.
...
     1.337
 
+    > `@`'a'                                     ::`'a'` is an atom represented as a different data type.
+    97                                             We cast it to the aura `@` to see what its integer representation is.
+    > (add 'a' 4)                                
+    101
+    (add 'a' 'b')
+    195  
+
+    > (add 0xc 0xf)                              ::`0xc` and `0xf` are of the hexadecimal soft atom type (`@ux`).
```
joshuareagan: "soft atom type" is too imprecise a term for stdlib docs.  We should assume a basic familiarity with Hoon here; no explaining auras in the `add` gate reference.

docs/hoon/library/1a.md
```diff
@@ -25,23 +25,37 @@ An atom.
...
+
     > (add 1.333 (mul 2 2))
     1.337
 
+    > `@`'a'                                     ::`'a'` is an atom represented as a different data type.
+    97                                             We cast it to the aura `@` to see what its integer representation is.
+    > (add 'a' 4)                                
+    101
+    (add 'a' 'b')
+    195
```
joshuareagan: Don't include the cord addition example.  This isn't a normal use case for the gate.

The hex addition is nice though.

docs/hoon/library/1a.md
joshuareagan: Also, short comments only for in-line comments; something like 4-5 words is probably already pushing it

docs/hoon/library/1a.md
```diff
@@ -62,33 +76,43 @@ An atom.
...
     ! exit
 
+    > (dec 'b')
+    97
+    > `@t`(dec 'b')
+    'a'
+
+    > (dec 0xc)                                  
+    11
+    > `@ux`(dec 0xc)
```
joshuareagan: spacing

docs/hoon/library/1a.md
```diff
@@ -62,33 +76,43 @@ An atom.
...
 --------
 
     > (dec 7)
     6
+
     > (dec 0)
     ! decrement-underflow
     ! exit
 
+    > (dec 'b')
```
joshuareagan: Let's not multiply examples too much.  We should drop the first cord example, maybe the second too.  I like the hex example, but that should be enough to clue people in on using `dec` with other auras.

docs/hoon/library/1a.md
```diff
@@ -105,35 +129,111 @@ An atom.
...
+    > `@`~sibtel-tallyd
+    7.303.014
+    > `@`~zod                                    ::the above operation returned an error because `~zod` is a representation of `0`.
+    0
+
+    > `@p`2                                      ::this casts `2` to `@p` to produce the ship name which represents that atom.
+    ~bud
+    > (div ~sibtel-tallyd ~bud)
+    > 3.651.507
+
```
joshuareagan: What are these `@p` examples supposed to show?

docs/hoon/library/1a.md
```diff
@@ -62,33 +76,43 @@ An atom.
...
+    0xb
 
 
 ***
 ### `++div`
 
 Divide
 
-Computes `a` divided by `b`.
+Computes `a` divided by `b`. Only a whole number, without the remainder, is produced.
```
joshuareagan: We can be less wordy:  Computes `a` divided by `b` without remainder.

docs/hoon/library/1a.md
```diff
@@ -105,35 +129,111 @@ An atom.
...
+    [p=0 q=0]
+
 
 ***
 ### `++gte`
 
 Greater-than/equal
 
-Tests whether `a` is greater than a number `b`.
+Tests whether a number `a` is greater than or equal a number `b`.
```
joshuareagan: "or equal *to*"

We probably shouldn't say "number" here, because this doesn't work for all Hoon "numbers" in general.  (e.g., try `(lth .12 .~11)` in the dojo... it'll return %.y).  Let's just say `a` and `b`, and let the input section specify "atom".

docs/hoon/library/1a.md
```diff
@@ -145,28 +245,41 @@ Accepts
 Produces
 --------
 
-A boolean.
+A flag. Our version of a boolean, except that `0` is yes/true and `1` is no/false.
```
joshuareagan: Not the place for defining "flag"

docs/hoon/library/1a.md
```diff
@@ -145,28 +245,41 @@ Accepts
...
     %.n
 
+    > `@`'h'                                     ::`h` is an atom represented as a `cord`, one of many different soft atom types.
+    104                                            ::We can cast it to the mold `@` to see what its integer representation is.
+    > `@`'b'
+    98
+    > (gte 'h' 'b')
+    %.y
+
+    > (gte '0xc' '0xf')
```
joshuareagan: Why hexadecimals as cords?

rmariani:
Are you against the casting examples? Or just the code-comments explaining
them?
—
~lodleb-ritrul
https://urbit.org

joshuareagan:
If by "casting examples" you mean:

```
> (add 0xc 0xf)
27

> `@ux`(add 0xc 0xf)
0x1b
```

...I like that one.  Adding hexadecimals and keeping them as `@ux` is a normal and useful thing to do, so it's good to illustrate that.

An example checking whether `'bob'` is less than `'Sally'` seems not very valuable to me.  The example lines showing a cast of atoms with various auras to vanilla atoms also seem superfluous.

rmariani:
Changes made

joshuareagan:
Looks good

PR #269, opened 2018-05-23
Add rune documentation for `=,`

joshuareagan:
Basic page addition for a rune that isn't yet documented

### DIFF
--- a/docs/hoon/rune/tis/com.md
+++ b/docs/hoon/rune/tis/com.md
@@ -7,11 +7,12 @@ title: =, "tiscom"
 
 # `=, "tiscom"` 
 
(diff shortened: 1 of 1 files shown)


belisarius222:
`=,` isn't just for cores and arms. It also applies to generic nouns and their faces:
```
=/  foo  [bar=2 baz=[3 qux=4]]
=,  foo
[bar qux.baz]  ::  [2 4]
```

belisarius222:
I think it would be good to include an example using faces instead of arms, too, but the text looks better now.

joshuareagan:
Whoops, didn't mean to merge this.  I was only trying to update my PR.  Noob mistake!

PR #270, opened 2018-05-23
adding example to illicitly added =, rune doc

joshuareagan:
I mistakenly merged the previous PR for the new `=,` rune page.  Complaints about this new rune page can go here.  I also added an example at @belisarius222 's request.


belisarius222:
LGTM

PR #271, opened 2018-05-23
Update 1c.md

rmariani:
Updated syntax, fixed code, explained more. Also used `#` headers to make subheads smaller.

@joshuareagan @keatondunsford

### DIFF
--- a/docs/hoon/library/1c.md
+++ b/docs/hoon/library/1c.md
@@ -9,137 +9,201 @@ navhome: /docs/
 
 Blocksize
 
-Atom representing a blocksize, by convention expressed as a power of 2.
+Atom representing block size. A block of size `a` has a bitwidth of `2^a`.
 
-Source
-------
+#### Source
 
-    ++  bloq  @                                            ::  blockclass
+    ++  bloq  @                                         
 
-Examples
---------
+#### Examples
 
-    > :: ++met measures how many bloqs long an atom is
-    > (met 3 256)
+    ::++met measures how many a-bloqs long an atom is. In other words, how
+    many bloqs of size a would be needed to represent b.
+
+    > (met 3 256)                      ::one 3-bloq has a bidwidth of 8, which can represent 0-255
     2
-    > :: 256 is 2 bloqs of 2^3
+
+    > (met 3 255)                      
+    1
+
+    > (met 3 65.535)                   ::two 3-bloqs together have a bidwidth of 16, which can represent 0-65,535                  
+    > 2              
+
+    > (met 2 65.536)                   
+    > 3
+
+    > (met 0 4)
+    3
+
+    > (met 1 4)
+    2
+
+    > (met 5 9.999.999.999)
+    2
+
+    > (met 2 0xb5)
+    2
+
+#### Discussion
+
+You can think of `bloq` as a numeral system that can represent values of
+`a^2^2`, where `a` is the size of a given block. That's because a block of size
+`a` is `a^2` bits wide, and a binary number that is `b` bits wide can represent
+`b^2` values.
+
+A `bloq` of 0 has a bitwidth of 1. (2^0 bits; can represent 2 values.)
+
+A `bloq` of 1 has a bitwidth of 2. (2^1 bits; can represent 4 values.)
+
+A `bloq` of 2 has a bitwidth of 4. (2^2 bits; can represent 16 values.)
+
+A `bloq` of 3 has a bitwidth of 8. (2^3 bits; can represent 256 values.)
+
+And so on.
 
 
 ***
 
 ### `++each`
 
-Mold of fork between two
+Mold of fork between two types
 
-mold generator: produces a discriminated fork between two types.
+A mold generator. Produces a discriminated fork between two types, defaulting to
+`a`.
 
-Source
-------
+#### Source
 
         ++  each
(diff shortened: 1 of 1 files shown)


docs/hoon/library/1c.md
```diff
@@ -9,243 +9,295 @@ navhome: /docs/
...
 
-Examples
---------
+#### Examples
 
-    > :: ++met measures how many bloqs long an atom is
-    > (met 3 256)
+    ::++met measures how many a-bloqs long an atom is. In other words, how
+    many bloqs of size a would be needed to represent b.
+
```
joshuareagan: Neatness issue---before we decided to add 'discussion' sections, this sort of comment got thrown in to the examples.  I think it would be better to remove lengthy comments from examples and add that commentary to 'discussion'.  Very brief comments seem fine in the examples, but long ones look sloppy.  The longest ones require scrolling right to read the whole thing, which is bad UX as well.

docs/hoon/library/1c.md
```diff
@@ -9,243 +9,295 @@ navhome: /docs/
...
+    > (met 3 256)                      ::one 3-bloq has a bidwidth of 8, which can represent 0-255
+    2
+
+    > (met 3 255)                      
+    1
+
+    > (met 3 65.535)                   ::two 3-bloqs together have a bidwidth of 16, which can represent 0-65,535                  
+    > 2              
+
```
joshuareagan: Some examples have a `>` before the return value and some don't.  You should remove these because they aren't consistent with what people see in the dojo.

docs/hoon/library/1c.md
```diff
@@ -9,243 +9,295 @@ navhome: /docs/
...
+    > (met 2 0xb5)
+    2
+
+#### Discussion
+
+You can think of `bloq` as a numeral system that can represent values of
+`a^2^2`, where `a` is the size of a given block. That's because a block of size
+`a` is `a^2` bits wide, and a binary number that is `b` bits wide can represent
+`b^2` values.
+
```
joshuareagan: My brain may be a bit fuzzy at the moment, but I find this explanation a little tough to follow.

docs/hoon/library/1c.md
```diff
@@ -9,243 +9,295 @@ navhome: /docs/
...
     > ? gate
-      <1.ybc {* <110.jyx 1.ztu $151>}>
-    <1.ybc {* <110.jyx 1.ztu $151>}>
+      <1.rdl {* <31.ohr 1.jmk $143>}>
+    <1.rdl {* <31.ohr 1.jmk $143>}>
 
     > (`gate`|=(a/* [a 'b']) 1)
-    [1 'b']
+    [1 98]
+
```
joshuareagan: Change the old `a/*` to `a=*`.  I also think you should add this example to the end:

```
> (`gate`|=(a=@ [a 'b']) 1)
nest-fail
```

It shows that failing to use the right sample type will get you into trouble.

docs/hoon/library/1c.md
```diff
@@ -9,243 +9,295 @@ navhome: /docs/
...
+#### Discussion
 
+A `gate` is analogous to a function in other programming languages. We created
+this new jargon because other constructs in Urbit are "functions" in the
+mathematical sense.
```
joshuareagan: I think we should mention that `gate` normalizes to an iron gate.  @belisarius222  Ted had an excellent adventure this weekend because the docs aren't clear about this.

docs/hoon/library/1c.md
```diff
@@ -9,243 +9,295 @@ navhome: /docs/
...
+    > *(quip @ _'hi')
     [~ 'hi']
 
+    > `(quip @ @ux)`[[1 2 3 4 ~] 0x10]
+    [~[1 2 3 4] 0x10]
+
+#### Discussion
 
+A common pattern in Hoon code is to return a `list` of changes, along with a
+new state. You'll often see `quip` it used in Gall apps.
```
joshuareagan: This prose is a little 'caveman' for my taste.

docs/hoon/library/1c.md
rmariani: I'm not sure what an iron gate is. How should it be explained here?

joshuareagan: Sorry, I missed this.  Understanding iron gates requires understanding how Hoon's type system handles polymorphism, and that's a huge topic---I don't think we want to say too much about that here.  The main point to get across is that people shouldn't worry that casts like this don't work, and that's okay:

```
> `gate`|=(@ 15)
nest-fail
```

That's because `gate` isn't for all gates, just iron ones.  I remember being confused when I couldn't cast my gates to `gate`!

rmariani: Ok, thank you for the explanation. Do you think an example like that would be sufficient?

joshuareagan: We should definitely also say that `gate` normalizes to an iron gate for those who know about polymorphism.  This example is a relatively accessible illustration of one consequence of that fact.

joemfb: Nitpick: `+gate` is a `$-` is a `$_` -- in other words, it only "normalizes" to a constant example. (I'm not sure if "normalize" is still being used to describe what happens when you call a "mold".)

PR #272, opened 2018-05-24
Update 2b

rmariani:
Updated syntax, made headers better, added `gulch`, alphabetized

@joshuareagan @keatondunsford

### DIFF
--- a/docs/hoon/library/2b.md
+++ b/docs/hoon/library/2b.md
@@ -4,28 +4,132 @@ navhome: /docs/
 
 
 # 2b: list logic
-Reverse
+### `++fand`
 
+All indices in `list`
+
+Produces the indices of all occurrences of `list` `nedl` in
+`list` `hstk` as a `list` of atoms.
+
+#### Accepts
+
+`nedl` is a list.
+
+`hstk` is a list.
+
+#### Produces
+
+A `list`.
+
+#### Source
+
+    ++  fand                                                ::  all indices
+      ~/  %fand
+      |=  [nedl=(list) hstk=(list)]
+      =|  i=@ud
+      =|  fnd=(list @ud)
+      |-  ^+  fnd
+      =+  [n=nedl h=hstk]
+      |-
+      ?:  |(?=($~ n) ?=($~ h))
+        (flop fnd)
+      ?:  =(i.n i.h)
+        ?~  t.n
+          ^$(i +(i), hstk +.hstk, fnd [i fnd])
+        $(n t.n, h t.h)
+      ^$(i +(i), hstk +.hstk)
+
+
+#### Examples
+
+    > (fand [3]~ ~[1 2 3])
+    ~[2]
+
+    > (fand [4]~ ~[1 2 3])
+    ~
+
+    > (fand ['a']~ "cbabab")
+    ~[2 4]
+
+    > (fand "ba" "cbabab")
+    ~[1 3]
+
+
+
+***
+### `++find`
+
+First index in `list`
+
+Produces the index of the first occurrence of `list` `nedl` in
+`list` `hstk` as the `unit` of an atom.
+
+#### Accepts
+
+`nedl` is a list.
+
+`hstk` is a list.
+
+#### Produces
+
+The `unit` of an atom.
+
+#### Source
+
+    ++  find                                                ::  first index
+      ~/  %find
+      |=  [nedl=(list) hstk=(list)]
+      =|  i=@ud
+      |-   ^-  (unit @ud)
+      =+  [n=nedl h=hstk]
+      |-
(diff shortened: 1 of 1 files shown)


docs/hoon/library/2b.md
```diff
@@ -4,28 +4,132 @@ navhome: /docs/
...
 
 # 2b: list logic
-Reverse
+### `++fand`
+
+All indices in `list`
+
+Produces the indices of all occurrences of `list` `nedl` in
+`list` `hstk` as a `list` of atoms.
+
```
joshuareagan: The "All indices in `list`" part looks out of place, and redundant with the fuller explanation just below.

Also, I'm pretty sure this next complaint is my fault not yours---but the `list` `nedl` and `list` `hstk` don't come out looking right.  I think we should just drop the `list` in each case.  We define `nedl` and `hstk` as lists immediately below, so this shouldn't cause any confusion.

docs/hoon/library/2b.md
```diff
@@ -4,28 +4,132 @@ navhome: /docs/
...
+    > (fand [4]~ ~[1 2 3])
+    ~
+
+    > (fand ['a']~ "cbabab")
+    ~[2 4]
+
+    > (fand "ba" "cbabab")
+    ~[1 3]
+
```
joshuareagan: Is there any particular reason you used `[blah]~` and not `~[blah]` in these examples?  These are equivalent when `blah` is an atom but not otherwise.  My habit has been to prefer putting the `~` in front for lists, whereas putting it at the end ensures the result is always taken as a one-item list.  But I don't know where my preference came from.  To be clear, I'm not making a criticism, I'm just wondering about what the standard is.  @joemfb @belisarius222 @keatondunsford ?

docs/hoon/library/2b.md
```diff
@@ -4,28 +4,132 @@ navhome: /docs/
...
+
+***
+### `++find`
+
+First index in `list`
+
+Produces the index of the first occurrence of `list` `nedl` in
+`list` `hstk` as the `unit` of an atom.
+
```
joshuareagan: Same thing with `list` here... they look weird and nothing is lost by dropping them.

docs/hoon/library/2b.md
```diff
@@ -35,46 +139,75 @@ Source
...
         --
       a
 
 
-Examples
---------
+#### Examples
 
     > lyst
     [i=1 t=[i=97 t=[i=2 t=[i=98 t=[i=[~ u=10] t=~]]]]]
```
joshuareagan: The `homo` examples could use a little more spacing.

docs/hoon/library/2b.md
```diff
@@ -991,31 +902,27 @@ is the continuation of the RNG.
...
 Gate to list, with state
 
-Accepts a `++list` `a` and a gate `b`. `c` is internal state, initially
+Accepts a `list` `a` and a gate `b`. `c` is internal state, initially
 derived by *bunting* the tail of the sample of gate `b`, instead of
 being passed in explicitly as in `++spin`. Produces a list with the
-gate applied to each element of the original list. `b` is called with a tuple
--- the head is an element of `a` and the tail is the state `c`, and should
+gate applied to each element of the original list. `b` is called with a #### tuple the head is an element of `a` and the tail is the state `c`, and should
 produce a tuple of the transformed element and the (potentially modified)
```
joshuareagan: A #### tuple?  Is this a mistake?

docs/hoon/library/2b.md
joemfb: > ... putting `~` at the end ensures the result is always taken as a one-item list ...

Exactly. The difference between `~[blah]` and `~[blah]` is not due to `blah` being an atom -- it's just the number of hoons inside the brackets:

```
> [1 2]~
[[1 2] ~]
> ~[1 2]
[1 2 ~]
> =+(a=[1 2] ~[a])
[[1 2] ~]
> =+(a=[1 2] [a]~)
[[1 2] ~]
```

I don't have any particular preference for these examples. These expressions for null-terminated tuples are confusing at first, then pretty easy to remember.

joshuareagan: "The difference between ~[blah] and ~[blah] is not due to blah being an atom -- it's just the number of hoons inside the brackets:"

`[(add 12 (mul 2 3))]~`

This has multiple Hoons in it, but because it evaluates to an atom it's the same as:

`~[(add 12 (mul 2 3))]`

joshuareagan: You're correct that my initial criterion is wrong:

```
[[12 14]]~
```

is the same as

```
~[[12 14]]
```

rmariani:
Ok, changes made.

PR #274, opened 2018-05-25
Update Ubuntu install process based on feedback from 2 users

kirkins:
I'm making this PR to suggest the fix that helped me with my problem https://github.com/urbit/urbit/issues/984

Another user has chimed in today that the same fix worked for him.


eglaysher:
Thank you!

PR #276, opened 2018-05-29
Updated 2e

rmariani:
Updated syntax and examples. Also added documentation for `++muk` and `++mum`.

@joshuareagan @keatondunsford @vvisigoth


joemfb:
`eny` is no longer 256 bits, but 512. I noticed this here, but I'm sure it applies throughout the docs.

rmariani:
Thanks, will update the docs accordingly
—
~lodleb-ritrul
https://urbit.org

PR #277, opened 2018-05-29
Update 2q

rmariani:
Updated syntax, headers, and examples, and added entries for `++knot` and `++tang`.

@joshuareagan @keatondunsford @vvisigoth

### DIFF
--- a/docs/hoon/library/2q.md
+++ b/docs/hoon/library/2q.md
@@ -8,7 +8,7 @@ navhome: /docs/
 
 Character
 
-A single character. Aura `@tD` designates a single unicode byte. All parsers
+A single character. Aura `@tD` designates a single UTF-8 byte. All parsers
 consume `tape`s, which are lists of `char`.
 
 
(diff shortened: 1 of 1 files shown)


docs/hoon/library/2q.md
```diff
@@ -32,150 +31,213 @@ Examples
...
+
+An atom type that only permits ASCII characters.
+
+#### Source
+
+++  knot  @ta
+
+#### Examples
+
+    > (knot 105)
```
joemfb: This should be a cast, not a clam (or whatever it's called now).

rmariani:
Ok, fixed all examples that are incorrect in that way.

docs/hoon/library/2q.md
```diff
@@ -6,23 +6,22 @@ navhome: /docs/
 # 2q: molds and mold builders
 ### `++char`
 
-Char
+Character
 
-A single character. odor `@tD` designates a single unicode byte. All
-parsers consume `++tape`s, which are lists of `++char`.
+A single character. Aura `@tD` designates a single unicode byte. All parsers
```
joemfb: "unicode byte" is a little imprecise, this should probably say UTF-8.

PR #281, opened 2018-06-04
Update 2f

rmariani:
Added `++lor` entry, added descriptions on other entries where they were missing, added examples and update syntax

@joshuareagan @keatondunsford

### DIFF
--- a/docs/hoon/library/2f.md
+++ b/docs/hoon/library/2f.md
@@ -8,28 +8,23 @@ navhome: /docs/
 
 Alphabetic order
 
-Computes whether `a` and `b` are in alphabetical order, producing a
-boolean.
+Computes whether `a` and `b` are in alphabetical order, producing a flag.
 
-Accepts
--------
+#### Accepts
 
 `a` is a noun.
 
 `b` is a noun.
 
-Produces
---------
+#### Produces
 
-A boolean atom.
+A flag.
 
+#### Source
 
-Source
-------
-
-    ++  aor                                                 ::  a-order
+    ++  aor                                                 
       ~/  %aor
-      |=  {a/* b/*}
+      |=  [a=* b=*]
       ^-  ?
       ?:  =(a b)  &
       ?.  ?=(@ a)
@@ -44,24 +39,32 @@ Source
         $(a (rsh 3 1 a), b (rsh 3 1 b))
       (lth c d)
 
-
-Examples
---------
+#### Examples
 
     > (aor 'a' 'b')
     %.y
+
     > (aor 'b' 'a')
     %.n
+
+    > (aor 'a' 'a')
+    %.y
+
     > (aor "foo" "bar")
     %.n
+
     > (aor "bar" "foo")
     %.y
+
     > (aor "abcdefz" "abcdefa")
     %.n
+
     > (aor "abcdefa" "abcdefz")
     %.y
+
     > (aor 10.000 17.000)
     %.y
+
     > (aor 10 9)
     %.n
 
@@ -72,27 +75,23 @@ Examples
 
 Numeric order
 
-Computes whether `a` and `b` are in ascending numeric order, producing a
-boolean.
+Computes whether `a` and `b` are in ascending numeric order, producing a flag.
 
-Accepts
--------
+#### Accepts
 
 `a` is a noun.
 
 `b` is a noun.
 
-Produces
---------
+#### Produces
 
-A boolean atom.
+A flag.
 
-Source
-------
+#### Source
 
-    ++  dor                                                 ::  d-order
+    ++  dor                                               
       ~/  %dor
-      |=  {a/* b/*}
+      |=  [a=* b=*]
       ^-  ?
       ?:  =(a b)  &
       ?.  ?=(@ a)
@@ -104,20 +103,24 @@ Source
       (lth a b)
 
 
-Examples
---------
+#### Examples
 
     > (dor 1 2)
     %.y
+
     > (dor 2 1)
     %.n
+
     > (dor ~[1 2 3] ~[1 2 4])
     %.y
+
     > (dor ~[1 2 4] ~[1 2 3])
     %.n
-    > (dor (limo ~[99 100 10.000]) ~[99 101 10.000])
+
+    > (dor `(list @)`~[99 100 10.000] ~[99 101 10.000])
     %.y
-    > (dor ~[99 101 10.999] (limo ~[99 100 10.000]))
+
+    > (dor ~[99 101 10.999] `(list @)`~[99 100 10.000])
     %.n
 
 
@@ -127,26 +130,26 @@ Examples
 
 Hash order
 
-XX revisit
+Computes whether the products of `a` and `b` hashed with `mug` are in
+ascending numeric order, producing a flag.
 
-Accepts
--------
+[`mug`](../2e) is the the 31-bit nonzero FNV-1a hash algorithm.
+
+#### Accepts
 
 `a` is a noun.
 
 `b` is a noun.
 
-Produces
---------
+#### Produces
 
-A boolean atom.
+A flag.
 
-Source
-------
+#### Source
 
-    ++  gor                                                 ::  g-order
+    ++  gor                                               
       ~/  %gor
-      |=  {a/* b/*}
+      |=  [a=* b=*]
       ^-  ?
       =+  [c=(mug a) d=(mug b)]
       ?:  =(c d)
@@ -154,29 +157,39 @@ Source
       (lth c d)
 
 
-Examples
---------
+#### Examples
 
     > (gor 'd' 'c')
     %.y
+
     > 'd'
     'd'
+
     > 'c'
+    'c'
+
     > `@ud`'d'
     100
+
     > `@ud`'c'
     99
-    > (mug 'd')
+
+    > (mug 'd')                               
     1.628.185.714
+
     > (mug 'c')
     1.712.073.811
+
     > (gor 'd' 'c')
     %.y
+
     > (gor 'c' 'd')
     %.n
+
     > (gor "foo" "bar")
     %.n
-    > (gor (some 10) (limo [1 2 3 ~]))
+
+    > (gor (some 10) `(list @)`[1 2 3 ~])
     %.n
 
 
@@ -184,31 +197,33 @@ Examples
 ***
 ### `++hor`
 
-Hash order
+Head/tail hash order
 
-XX revisit
+If `a` and `b` are cells, `hor` checks to see if their heads are equal. If their
+heads are equal, `gor` is computed on their tails, producing a flag. If their
+heads are not equal, `gor` is computed on their heads, producing a flag.
 
-Recursive hash comparator gate.
+If `a` and `b` are atoms, they are computed as samples of `gor`, producing a
+flag.
 
-Accepts
--------
+If `a` is an atom and `b` is not, then `%.y` is produced. If `b` is an atom and
+`a` is not, `%.n` is produced.
+
+#### Accepts
 
 `a` is a noun.
 
 `b` is a noun.
 
-Produces
---------
+#### Produces
 
-A boolean atom. 
+A flag.
 
+#### Source
 
-Source
-------
-
-    ++  hor                                                 ::  h-order
+    ++  hor                                            
       ~/  %hor
-      |=  {a/* b/*}
+      |=  [a=* b=*]
       ^-  ?
       ?:  ?=(@ a)
         ?.  ?=(@ b)  &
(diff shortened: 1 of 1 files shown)


joemfb:
It'd be really nice if the various hash-orders had some explanation of why they exist. Sets use `+hor`, maps use `+gor`, and queues use `+vor` -- but I have no idea why.

I realize this is not a trivial request ... ;)

belisarius222:
I'll give this a shot, since I'm already looking into treaps for Ford caching today. @cgyarvin you might want to take a look at this to make sure I got this right.

Each ordering function is intended to be used as a comparator for sorting. It takes two arbitrary nouns `a` and `b` as input and produces `&` if `a` should come before `b`, or `|` if `b` should come before `a`. In particular, some of these functions provide probabilistic balancing for Urbit's treap data structures: sets, maps, and queues.

`+aor` is alphabetical order. It treats each atom in the tree as a character and checks highest bit first.

`+dor` compares numeric values. If `a` and `b` are atoms, `dor` is equivalent to `+lte`. If they're cells, it recurses on the heads, and then if the heads are the same it checks the tails. If `a` is an atom and `b` is a cell, produce `&`. If `a` is a cell and `b` is an atom, produce `|`. Intuitively, for cells, if each of the elements are equal, `dor` says the 'shorter' one should come first.

`+gor` compares the short hashes `(mug a)` and `(mug b)`, falling back to `+dor` if the mugs are equal.

`+hor` runs `gor` if both `a` and `b` are atoms. Otherwise, if one is a cell and the other is an atom, it will say the 'shorter' one comes first, using the same logic as `+dor`. If they're both cells, use `gor` on the heads, or if the heads are equal, run `gor` on the tails.

`+lor` is not used anywhere in Arvo. It seems to be the same check as `+dor ` except it doesn't check the case where `a` is a cell and `b` is an atom. I suspect this might be a bug.

`+vor` compares double-hashed `(mug (mug a))` and `(mug (mug b))`, dropping down to `+dor` if they're equal.

The idea of a treap is that it keeps itself approximately balanced both vertically and horizontally. It does this by behaving like a binary search tree in terms of horizontal ordering, and then the vertical ordering is uncorrelated to that horizontal ordering.

For a given node and its children (n, l, r), using horizontal ordering l <= n <= r must hold. There are multiple valid tree shapes for any set of nodes, though, so to keep it balanced, the treap also enforces a vertical ordering. 

This vertical ordering should be uncorrelated to the horizontal ordering; otherwise the treap will become unbalanced. The limiting case uses the same vertical ordering as horizontal ordering, which produces a list. At that point, we've lost all the performance benefits of a treap.

Since hashing (hopefully) destroys correlation, double-mugging should be mostly uncorrelated with single-mugging, which is why we single-mug in one direction and double-mug in the other. Since the horizontal ordering is used for lookups, we use the faster single-mugging along that dimension. It should also be noted that the mug of a noun is calculated lazily in the process of testing equality, and it's memoized at the C level.

Sets (`+set`, `+in`) use `+hor` for their horizontal order and `+vor` for their vertical order. See `+apt:in` to see how a set verifies its correctness. I think `hor` is short for 'horizontal order' and `vor` is short for 'vertical order.' 

Maps (`+map`, `+by`) only look at the keys (the head of the key-value pair elements) for ordering. They use `+gor` on the key for horizontal order and `+vor` on the key for vertical order.

Queues (`+qeu`, `+to`) only use `+vor`, for vertical ordering. I think the horizontal ordering is the insertion order of the elements; i.e. a FIFO queue. It seems to me that we might be better off using `+gor` for queues, since it would be faster, and even a single-mugging should be uncorrelated with the insertion order in general.

`+hor` is a wrapper around `+gor` that first runs `+gor` on the heads of `a` and `b`, whereas `+gor` hashes the whole nouns and falls back to value-by-value equality comparison in the case of hash collision. I'm having trouble understanding why maps use `+gor` and sets use `+hor`. Maybe it has to do with expecting the keys in a map to be an atom and having no such expectation of set elements?

cgyarvin:
I have all the same questions you do :-)

rmariani:
Looks awesome to me. Thanks Ted.
—
~lodleb-ritrul
https://urbit.org

joemfb:
This is fantastic! Thanks for taking the time.

Should I open a tracking issue on urbit/arvo to see if the set/map treap ordering can be unified? This has been discussed a few times in the past, not sure of the level of interest.

belisarius222:
Sounds like a good idea. There should probably also be an issue to remove
`+lor`, unless someone is attached to its presence.


—
~rovnys-ricfer
https://urbit.org

ohAitch:
lor is my fault, and should be removed

the map vs set ordering is greviously bugged, in that the horizontal order
is the same so a map pretty-prints the same as the set of its k-v pairs,
but the vertical order is incompatible so the two aren't actually .= to
each other. This should be resolved by either adding an equivalent
head-first clause to set-vor, or just removing the special case and making
maps and sets more transparently distinct.

Queues are in fact FIFO, and only have vertical ordering; if you want them
to be faster, I'd redefine vor to be single-mug and hor to be double-mug,
and not just structure queues completely differently from other treaps.
That said, I suspect the mug overhead to be small-o, and absolutely all
urbit computations to be dominated by pointer-walking and allocation: if
profiling shows >5% increase, it might be worth it, but I think profiling
would instead show the number of mugs to be noise compared to other known
performance issues.

belisarius222:
Oh, so sets use `+hor` so that their horizontal ordering matches the
horizontal ordering of maps. Since `+hor` checks the head first, it would
behave like maps using `+gor` on the key (the head) -- unless there's a
hash collision! Then they'd diverge: sets would mug the tails, whereas maps
would do a full equality check on the heads. I wonder if that's happened
yet ...

I don't feel as though the internals of maps and sets should be constrained
to having anything to do with each other at all. If they happen to use the
same ordering or different ordering, that shouldn't be anybody's business.
The programmer should be relying on the invariants provided by the set and
map interfaces, as opposed to relying on implementation details to convert
data from one form to another.

If we agree on that, then I think that implies we should use `+gor` as
horizontal ordering for both maps and sets (only on the key for maps) and
delete `+hor` and `+lor`.

Sounds like we should also change `+qeu` to use `+gor` instead of `+vor`.
Even if it's not a big performance difference, there's no reason to make
extra work for the interpreter in our standard library functions.

I believe any of these changes would require either a breach or a
relatively heroic system-wide state migration; otherwise we'd break those
precious invariants.


—
~rovnys-ricfer
https://urbit.org

joemfb:
Nothing relies on set and map using mutually consistent hash orders (although I did made that mistake in the past, see urbit/arvo#268).

I do think that it would be to nice to have a map be the same shape as a set of the keys, especially if we're going to make changes in this area.

ohAitch:
I'm not sure what you mean about hash collisions:

      ?:  =(-.a -.b)
        (gor +.a +.b)
      (gor -.a -.b)

++gor in turn drops down to ++dor if the hashes match, you'll never have conflicting `(gor a b)` and `(gor b a)` unless `=(a b)`.

++vor stands for "vertical order", ++hor stands for "horizontal order", please don't use the horizontal order as a vertical order unless you actually have performance numbers that say
a) qeu is being used extensively, such that the extra-mug overhead is having any noticeable effect
b) set/map use `++hor` significantly more than `++vor`, such that having them be inconsistent is worthwhile

> I don't feel as though the internals of maps and sets should be constrained
to having anything to do with each other at all.

They are constrained by urbit being a system meant to be completely understandable by a single user, and "there is a consistent treap structure that is used for maps/sets/queues" is easier to completely understand than "these three datastructures are superficially similar but you need to memorize completely unrelated hashing schema for each". It should certainly be *possible* to write hoon programs using only the invariants, but underneath, things being simpler is preferable to them being complicated.

Changing the shapes of datastructures with jets will want a breach, yes.

ohAitch:
(note that "a map is shaped like a set of keys, a set is shaped like a qeu someone happened to put the elements into in horizontal order" is consistent; as is "a map is a set with no duplicate element-heads, and uses head-only variants of the horizontal and vertical order to make this clear in the implementation". The current map situation is obviously the worst of both worlds)

belisarius222:
Oh you're right @ohAitch, I was wrong about the hash collisions causing
divergent behavior.

please don't use the horizontal order as a vertical order

So much for lapidary naming's mantra that "interpretation is defined by the
formula and nothing else." If being named `vor` means we're not allowed to
use it anywhere other than as a vertical treap ordering, let's rename it.

I can see two possible reasons for wanting map and set to share tree shape
or orderings: performance, and limiting the number of concepts required to
understand Urbit.

As for number of concepts, map and set will never be hashed exactly the
same way. Maps will have only their keys hashed, while sets will have the
whole element hashed. This means you'll still have to learn two different
hashing schemes, even though they both use mugs. By the time someone has
absorbed the concepts of treaps, binary search trees, and probabilistic
balancing, having a couple of slightly different ordering schemes is
definitely not the hard part, especially if they had a coherent rationale
and a modicum of documentation. Take it from me; I just spent the past two
days doing exactly this.

We will never "make this clear in the implementation." We will make this
clear by documenting the fact that these are treaps and documenting our
justifications for our choices of orderings.

Sharing a tree shape would matter for performance if we could use a set as
a map in-place, or vice-versa. That would save memory and CPU by obviating
the need for a copy operation. However, there's no guarantee that a set
will become a valid map, because a set could store values with equal heads
and differing tails, which would become a map with duplicate keys. We could
treat a map as a set of key-value pairs, but we wouldn't be able to convert
back.

Does this matter, though? I think the reason we're even considering this is
because we don't have an iterator / sequence / lazy list type, so every
time we map some operation over some container, we either have to hand-roll
stream fusion with specialized ad-hoc recursion (which we usually don't),
or we copy the whole data structure. I'd rather fix this problem in a
general way than special-case a container type to support a one-way
transformation to one other container type.

ohAitch:
By "don't use horizontal order as vertical order", I mostly mean that if for sets the code is `?:((vor new-element n.a) insert-here $(... recur downwards))`, the qeu should use the same function for the same purpose, regardless of name. Which is a *should* not a must, by RFC 2119: if the performance case is compelling, it is probably worth changing, but the burden of proof is on doing the confusing thing.

Like, is there another reason *for* making them different, other than "fewer concepts is too constraining"? I am not opposed to having more general iterators, but I don't think those are scheduled to exist soon enough to be a consideration

ohAitch:
horizontally, sets do not *currently* have their whole element hashed: they have their heads(/atoms) hashed, and their tails only if the heads are equal

vertically, they do hash the whole element.

this inconsistency caused urbit/arvo#268, and is what I was referring to earlier with the "greviously bugged" printing-identically-but-different-vertically behaviour. The fix is to make vertical order do the same head-first thing, or horizontal order not do the head-first thing.

belisarius222:
Ah ok, I see what you're saying about the vertical order.

The reason to make them different is that the simplest ordering that solves the problem for sets is `+gor`, not `+hor`.

I found it confusing that the implementation of a data structure was based on a rather subtle interaction with some other data structure, as opposed to being the most obvious solution for that data structure in isolation. I think we should implement a set as a set, a map as a map, and a queue as a queue. They share a `+tree` structure and some aspects of their orderings, but they're not the same data structure. Trying to get these things to be artificially more similar to each other is part of how we got into this mess in the first place.

Something along these lines could be interesting (the types aren't right, I know; it's pseudohoon):
```
++  treap  [val-type=mold hor=$([* *] ?) vor=$-([* *] ?)]
...
+=  map  |*  val-type  `treap`[val-type hor=|=(^ (compare-mugs +<- +<+)) vor=compare-double-mugs]
+=  set  |*  val-type  `treap`[val-type hor=compare-mugs vor=compare-double-mugs]
```

ohAitch:
Hmm, only two(and a half?) instances probably isn't worth the abstraction overhead, but I can see that being useful if we grow like five more similar tree containers(the %jael static-keyset map seems relevant here).

I'm on board with the "hor elements for sets, hor keys for maps" resolution definitely ^-^

joshuareagan:
Rob, with a little help from Ted, would you be willing to add a bit to this doc in order to fulfill Joe's initial request?

rmariani:
Of course.

rmariani:
Alright, here's my latest hack at it.

cgyarvin:
Wut?  Remember that a map is a search structure.  If I’m searching for a key but don’t have the value, how do I search a tree ordered by [key value]?

Maybe I don’t understand the proposal...

ohAitch:
When I last looked at the code, sets were ordered by "key"(head of element)
first, and the tail as a tie breaker.

But only horizontally, not vertically. And while I appreciate the elegance
of the code's clear original intent, and am all for Ted implementing it for
real, the important part of the ~proposal~ as I currently remember it is to
fix the current(?) egregious state of "maps and sets of the same pairs
pretty-print identically, but aren't .="

cgyarvin:
I had another idea for fixing this — but I get it.  A tricky change to implement, but I also want to switch to Murmur32 mug... also tricky...

belisarius222:
This proposal would use a map-compatible search order in sets, by comparing
heads (keys) before looking at tails (values). Because a map can never have
duplicate keys, this scheme allows you to convert any valid map into a
valid set -- but not necessarily the reverse.

I don't know anything about the interaction between treap ordering and
pretty-printing.


—
~rovnys-ricfer
https://urbit.org

cgyarvin:
And you don’t need to know anything!

Pretty-printing remains in the Tao.  As Lao-tze says: those who know, don’t talk.  Those who talk, don’t know. I’d talk, but I don’t know...

ohAitch:
The interaction is that because horizontal ordering *already* acts as
described, and the pretty-printer ignores vertical ordering, constructing a
map and set in the dojo makes them look like they're hashed
identically(expectation), while in practice they're structured
differently(reality)

joshuareagan:
fwiw this pretty-printer issue has fooled me in the past

cgyarvin:
There are other ways to handle the ordering difference.  But I’m reminded that originally it was designed to work as you describe...

eglaysher:
I'm unsure how you'd go about making the map ordering depend on the value; we have ways of just modifying the value in place (`+jab:by`) which would otherwise require changing the tree structure if the ordering depended on the value.

joemfb:
I would just apply the single horizontal hash-order function to the keys, as we do now. The difference would be that it would be same horizontal hash-order used on map keys and set elements.

joemfb:
To be more specific, I think the current `+vor` (double-mug order) should remain the vertical order, and the current `+gor` (mug order) should be our exclusive horizontal order.

joemfb:
These changes have now been staged on the linked PRs.

PR #284, opened 2018-06-08
Update 2k

rmariani:
Adding the missing entry `+-nip:to`, added discussion sections, filled in missing bodies of entries, updated source code, updated syntax, added examples

Thanks to Ted for helping with the discussion of the tree.

### DIFF
--- a/docs/hoon/library/2k.md
+++ b/docs/hoon/library/2k.md
@@ -8,38 +8,41 @@ navhome: /docs/
 
 Queue operations
 
-Container arm for queue operation arms. The contained arms inherit its
-sample `++qeu` `a`.
+Container arm for queue operation arms. The contained arms inherit `a`, its
+sample queue.
 
-Accepts
--------
+#### Accepts
 
-`a` is a queue, ++qeu.
+`a` is a tree.
 
-Source
-------
+#### Source
 
-    ++  to                                                  ::  queue engine
-      |/  a/(qeu)
+    ++  to                                                 
+      |_  a=(tree)
 
+#### Discussion
+
+`tree` is the generic tree-shape mold. Maps, sets, and queues all share that
+shape, but they put their elements in different places and have different
+interfaces. `to` is the interface core for `qeu`; it's not generic across all
+the different containers that use `tree`.
+
+***
 ### `+-bal:to`
 
 Balance
 
 Vertically rebalances queue `a`.
 
-Accepts
--------
+#### Accepts
 
 `a` is a queue.
 
-Produces
---------
+#### Produces
 
 A queue.
 
-Source
-------
+#### Source
 
       +-  bal
         |-  ^+  a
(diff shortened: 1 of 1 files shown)


docs/hoon/library/2k.md
```diff
@@ -8,38 +8,41 @@ navhome: /docs/
...
+#### Source
 
-    ++  to                                                  ::  queue engine
-      |/  a/(qeu)
+    ++  to                                                 
+      |_  a=(tree)
 
+#### Discussion
+
+`tree` is the generic tree-shape mold. Maps, sets, and queues all share that
```
belisarius222: `+tree` is the generic mold builder for trees where each node looks like either `~` or `[n=<some value> l=tree r=tree]`.

docs/hoon/library/2k.md
```diff
@@ -8,38 +8,41 @@ navhome: /docs/
...
+shape, but they put their elements in different places and have different
```
belisarius222: The places where the elements can go are the same; they're just ordered differently. I'd say something more like:
`but they store their elements in different orders and have different interfaces`.

docs/hoon/library/2k.md
```diff
@@ -49,128 +52,119 @@ Source
...
+    > =a (~(gas to `(qeu @)`~) [1 2 3 ~])
+    > =b [4 5 6 ~]
+    > `(qeu)`(~(gas to a) b)
+    {6 5 4 3 2 1}
 
+***
 ### `+-get:to`
 
-Pop
+Head-tail pair
```
belisarius222: I don't understand what 'head' means here. I assume it means the element that was inserted first? I think we should document that this is a FIFO queue, as opposed to some other kind of queue.

docs/hoon/library/2k.md
```diff
@@ -180,164 +174,192 @@ Source
...
-    > a
+
+    > `(qeu)`a
     {6 5 4 3 2 1}
 
-------------------------------------------------------------------------
+***
+###  `+-nip:to`
+
+Removes the root of queue `a`, producing the resulting queue.
```
belisarius222: What significance does the root element have? I wonder if this is really just part of the private interface for the queue to rebalance itself, as opposed to something intended to be called publicly. If so, we should mention that in the documentation.

docs/hoon/library/2k.md
```diff
@@ -180,164 +174,192 @@ Source
...
 
-------------------------------------------------------------------------
-
+***
 ### `+-tap:to`
 
 Queue to list
 
-Produces `++queue` `a` as a `++list` from front to back.
+Produces queue `a` as a list from front to back.
```
belisarius222: Does this mean it produces the elements in FIFO order, or the reverse? the "front" and "back" concepts don't seem totally clear to me.

docs/hoon/library/2k.md
```diff
@@ -180,164 +174,192 @@ Source
...
 
 A queue.
 
-Source
-------
+#### Source
 
-      +-  nap                                               ::  removes head
+      +-  nap                                              
         ?>  ?=(^ a)
```
belisarius222: Looks like this errors out on an empty queue. That should be documented.

rmariani: Interesting, will do so

msutherl:
How's this PR doing?

rmariani:
I'm going to finally go through it and make it Good

PR #287, opened 2018-06-11
Update 2h

rmariani:
Updated syntax, fixed examples, added missing entries and added missing descriptions.

### DIFF
--- a/docs/hoon/library/2h.md
+++ b/docs/hoon/library/2h.md
@@ -255,7 +255,7 @@ A set.
 
 
 ***
-###`+-dif:in`
+### `+-dif:in`
 
 Difference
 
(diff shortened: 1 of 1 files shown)


docs/hoon/library/2h.md
```diff
@@ -152,141 +237,189 @@ Source
...
     {'bonita' 'madeleine' 'daniel'}
+
     > (~(del in b) 'susan')
     {'bonita' 'madeleine' 'daniel' 'john'}
 
-------------------------------------------------------------------------
 
+***
+###`+-dif:in`
+
```
joshuareagan: Get a space between `###` and `+-dif:in`.  It doesn't render correctly without it.

rmariani:
Good catch. Fixed.

PR #288, opened 2018-06-15
Update 3e

rmariani:
AES encryption was removed from the standard library.


joemfb:
AES and ed25519 were moved to zuse.

rmariani:
Noted, thanks

PR #289, opened 2018-06-18
Update 2c

rmariani:
Added `+fe` and the arms contained therein.

### DIFF
--- a/docs/hoon/library/2c.md
+++ b/docs/hoon/library/2c.md
@@ -746,3 +746,434 @@ An atom.
 
 
 ***
+### `++fe`                                                 
+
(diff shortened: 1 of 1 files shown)


docs/hoon/library/2c.md
```diff
@@ -745,4 +745,435 @@ An atom.
...
+
+    > (~(dif fe 2) 63 64)
+    15
+
+***
+### `++inv`
+
+Inverse
+
+Inverts the order of the mowdular field.
```
joemfb: typo here

rmariani: thanks

PR #292, opened 2018-06-27
Update 2o

rmariani:
Cleaned up formatting, fixed examples, updated syntax

### DIFF
--- a/docs/hoon/library/2o.md
+++ b/docs/hoon/library/2o.md
@@ -2,150 +2,172 @@
 navhome: /docs/
 ---
 
-
 # 2o: normalizing containers
 ### `++jar`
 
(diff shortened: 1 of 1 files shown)


docs/hoon/library/2o.md
```diff
@@ -2,150 +2,172 @@
 navhome: /docs/
 ---
 
-
 # 2o: normalizing containers
 ### `++jar`
 
+Tile generator. A `jar` is a `map` of `list`.
```
joemfb: I think this should be "mold generator", to match the others. "tile" is a soon-to-be-obsolete internal compile-time representation of a mold.

rmariani:
good catch

PR #294, opened 2018-07-03
how to use a different version of meson

TJamesCorcoran:
Some notes for people frustrated with a bad version of meson.


ohAitch:
((can urbit be switched back to  make  , which isn't an 0.x project and
thus doesn't have breaking changes on minor revisions, yet))

galenwp:
I believe there's another thread on this topic from a while ago.

PR #296, opened 2018-07-04
fix bug in Hoon implementation of Nock

joshuareagan:
it didn't compute correctly for certain formulas


ohAitch:
yikes, that [excerpt from `++mink`](https://github.com/urbit/arvo/blob/master/sys/hoon.hoon#L5443-L5517) sure wasn't correctly excerpting ++mink

PR #300, opened 2018-07-09
New section on using/admin.md on how to create simple self signed SSL…

ramrunner:
Although i was informed that urbit is moving towards the use of let's encrypt certificates. it seems that at least for localhost, self-signed certificates could still be useful. The section can be used in the future to incorporate the certbot/let's encrypt approach.

### DIFF
--- a/docs/using/admin.md
+++ b/docs/using/admin.md
@@ -149,3 +149,17 @@ Urbit network.
 
 When this happens, back up any files you'd like to save, shut down your
 Urbit and recreate it (as if you were starting for the first time).
+
+## Creating SSL certificates
+
+It is possible to create SSL certificates so that you serve files through your ship's webserver using https.
+To do that you, would need a machine that has OpenSSL installed.
+In order to create a new private key and a self-signed certificate for your localhost, valid for a year, you can issue the following commands:
+
+```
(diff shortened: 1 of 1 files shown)


rmariani:
Thanks! I'd like to requisition the expertise of @joemfb before I merge.

docs/using/admin.md
```diff
@@ -149,3 +149,17 @@ Urbit network.
...
+## Creating SSL certificates
+
+It is possible to create SSL certificates so that you serve files through your ship's webserver using https.
+To do that you, would need a machine that has OpenSSL installed.
+In order to create a new private key and a self-signed certificate for your localhost, valid for a year, you can issue the following commands:
+
+```
+$ openssl req -x509 -sha256 -newkey rsa:4096 -keyout privatekey.pem -out certificate.pem -subj '/CN=localhost' -days 365
+$ mkdir -p your-urbit/.urb/tls
+$ mv privatekey.pem certificate.pem your-urbit/.urb/tls
```
joemfb: The key should be `private.pem`. See https://github.com/urbit/urbit/blob/master/vere/http.c#L693

joemfb:
Should this maybe go in the Web section of the admin page? Or maybe on the separate web-specific page? Not sure who's making those calls.

ramrunner:
thanks for the review and sorry for the wrong key filename. i squashed the new commits together into one. i'm not changing sections for now.

PR #302, opened 2018-07-13
Fixed 2h examples

rmariani:
Some examples here showed a different result than they would when tested in the Dojo

### DIFF
--- a/docs/hoon/library/2h.md
+++ b/docs/hoon/library/2h.md
@@ -137,6 +137,7 @@ A flag.
     %.y
 
     > =a (sy [1 2 3 4 5 6 ~])
+    > a
     [n=6 l={5 4} r={1 3 2}]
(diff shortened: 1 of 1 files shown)


joemfb:
It strikes me as suboptimal to duplicate these type-related misprints in the docs, as opposed to fixing them in `hoon.hoon`. Some well-placed  `^-  (set _?>(?=(^ a) n.a))` casts (or similar) inside `+in` should resolve these printing issues.

joshuareagan:
I think we have to document the Hoon we have, not the one we'll have at some point in the future.  We had someone in /urbit-meta confused because the docs didn't match his dojo result.  I'm pretty sure @cgyarvin is making changes to the pretty-printer, and it also seems suboptimal to fix dead code.  We can (and will) revisit this when the pretty-printer is fixed up.

PR #311, opened 2018-08-06
Update 4a

rmariani:
Arranged, formatted, and updated. Some were previously marked as being prefix operations when they were the opposite, and vice versa


joshuareagan:
Good catch.  It's slightly dubious to use `@t` examples when the gate calls for `@tas` atoms, but I can't imagine anyone being confused by this.

PR #315, opened 2018-08-20
Write 3b

rmariani:
Wrote documentation for the previously undocumented 3b

### DIFF
--- a/docs/hoon/library/3b.md
+++ b/docs/hoon/library/3b.md
@@ -123,7 +123,7 @@ infinite exponent range (`$i`). Default value is `%d`.
       |%
 
 ***
-### `++^rou:fl`
+### `++rou:fl`
 
 Round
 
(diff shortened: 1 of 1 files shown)


joshuareagan:
There's a lot of good work here!  That said, I'm not crazy about some of these headers, e.g., `++^rou:fl`.  This is pretty far removed from what we see in the code, and by itself it doesn't make much sense either.  I'd rather just have us cite the arm name, e.g., `++  rou`.  Context makes clear which door this arm is under.

cgyarvin:
It’s a good stab but new Hoon will be much better at these kinds of names/paths.

rmariani:
I suppose so. I just wanted to minimize confusion from stuff like a reader
ctrl+f-ing and finding different results

—
~lodleb-ritrul
https://urbit.org

rmariani:
Alrighty, this new version lacks the offending `^`s

rmariani:
This has a merge conflict and isn't in udon

PR #317, opened 2018-08-24
Update glossary

rmariani:
Added entries (most notably on Arvo + vanes),

### DIFF
--- a/docs/about/glossary.md
+++ b/docs/about/glossary.md
@@ -391,19 +391,18 @@ information in speech. [`Runes`](#-rune) are composed of two glyphs.
 Below is a complete list of glyphs with their associated pronunciations.
 
 ```
-ace [1 space]   gal <               pal (
-bar |           gap [>1 space, nl]  par )
-bas \           gar >               sel [
-buc $           hax #               mic ;
-cab _           hep -               ser ]
-cen %           kel {               sig ~
-col :           ker }               soq '
-com ,           ket ^               tar *
-doq "           lus +               tec `
-dot .           pam &               tis =
-fas /           pat @               wut ?
-zap !
-
+ace  [1 space]       hep  -            rob  }
+ban  >               ket  ^            say  '
+bar  |               lac  [            sig  ~
+bat  \               led  <            tar  *
+bus  $               lit  (            tec  `
+cab  _               lob  {            tis  =
+cen  %               lus  +            vat  @
+col  :               mic  ;            wut  ?
+com  ,               net  /            yel  "
+dot  .               pad  &            zap  !
+gap  [>1 space, nl]  rac  ]     
+hax  #               rit  )
 ```
 
 ### Hall
--- a/docs/about/glossary.md
+++ b/docs/about/glossary.md
@@ -391,17 +391,19 @@ information in speech. [`Runes`](#-rune) are composed of two glyphs.
 Below is a complete list of glyphs with their associated pronunciations.
 
 ```
-ace  space      gal  <          per  )
-bar  |          gar  >          sel  [
-bas  \          hax  #          mic  ;
-buc  $          hep  -          ser  ]
-cab  _          kel  {          sig  ~
-cen  %          ker  }          soq  '
-col  :          ket  ^          tar  *
-com  ,          lus  +          tec  `
-doq  "          pam  &          tis  =
-dot  .          pat  @          wut  ?
-fas  /          pel  (          zap  !
+ace [1 space]   gal <               pal (
+bar |           gap [>1 space, nl]  par )
+bas \           gar >               sel [
+buc $           hax #               mic ;
+cab _           hep -               ser ]
+cen %           kel {               sig ~
+col :           ker }               soq '
+com ,           ket ^               tar *
+doq "           lus +               tec `
+dot .           pam &               tis =
+fas /           pat @               wut ?
+zap !
+
 ```
 
 ### Hall
--- a/docs/about/glossary.md
+++ b/docs/about/glossary.md
@@ -8,191 +8,445 @@ title: Glossary
 # Glossary
 
 Urbit is renowned for its exotic terminology. Here's a simple overview
-from the strange words in.
+from the strange words within.
 
 As Dijkstra put it: "The purpose of abstraction is not to be vague, but
 to create a new semantic level in which one can be absolutely precise."
 
-## Ships
+### application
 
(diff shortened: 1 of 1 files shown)


docs/about/glossary.md
```diff
@@ -8,191 +8,445 @@ title: Glossary
...
+bas  \          hax  #          mic  ;
+buc  $          hep  -          ser  ]
+cab  _          kel  {          sig  ~
+cen  %          ker  }          soq  '
+col  :          ket  ^          tar  *
+com  ,          lus  +          tec  `
+doq  "          pam  &          tis  =
+dot  .          pat  @          wut  ?
+fas  /          pel  (          zap  !
+```
```
joshuareagan: Is this chart canon?  What happened to `gap`?

Fang-:
More incoming changes are mentioned [here](https://fora.urbit.org/general/posts/~2018.6.29..01.08.03..c4af~/). I'm still against most of these, and as such consider them non-final. (New hoon hasn't released yet, after all.) The pronouncables listed here should probably just match those on `/hoon/syntax.md` for now.

joshuareagan:
The new Hoon release is imminent and the pronunciation changes are almost entirely a separate matter.  The only connection is the way parsed expressions are tracked by the Hoon compiler, and no one who is looking up stuff in Hoon is going to be confused if we post the updated vocalizations now.

That said, I don't know how final these changes are.  @cgyarvin ?

cgyarvin:
I’m happy to explain my motivations but this is a classic bikeshed moment.  Mostly the new pronunciations (all found in the h.h parser) are either vocally faster (eg, fas to net — try saying each ten times as fast as you can) or eliminate confusion (sem, which sounds like cen, becomes mic).

Search and replace in the docs is easy, so I’m not too worried either way.  I hate doing this to people, but I’d rather do it to people who are at least getting paid to suffer... and of course I’ve already done it to myself.

joshuareagan:
Good enough for me.

Rob, use the names from this piece of code:

https://github.com/cgyarvin/arvo/blob/research-tome/sys/hoon.hoon#L5126

Alphabetize them as well please.

rmariani:
Aye aye.

—
~lodleb-ritrul
https://urbit.org

joshuareagan:
Actually I needed it too, so here it is:

```
ace  [1 space]       hep  -            rob  }
ban  >               ket  ^            rit  ) 
bar  |               lac  [            say  '
bat  \               led  <            sig  ~
bus  $               lit  (            tar  *
cab  _               lob  {            tec  `
cen  %               lus  +            tis  =
col  :               mic  ;            vat  @
com  ,               net  /            wut  ?
dot  .               pad  &            yel  "
gap  [>1 space, nl]  rac  ]            zap  !
hax  #                
```

ohAitch:
Shortness is good but IMO "net" is pretty indefensible, how exactly would you propose pronouncing `http://netrunner.net/site`

(general objection: existing english words are fine iff they already mean the relevant symbol, i.e. `bar dot gap tis at wut zap`. `ace` gets a pass for being an abbreviation and infrequently explicit, `hax` is questionable but cute and also invalid in live `.hoon` files, `pad` and `led` are highly questionable, `tar` is okay maybe)

rmariani:
The spacing was a little confusing in yours, and it wasn't quite
alphabetized (rob and rit), so here's my edit:


ace  [1 space]       hep  -            rob  }
ban  >               ket  ^            say  '
bar  |               lac  [            sig  ~
bat  \               led  <            tar  *
bus  $               lit  (            tec  `
cab  _               lob  {            tis  =
cen  %               lus  +            vat  @
col  :               mic  ;            wut  ?
com  ,               net  /            yel  "
dot  .               pad  &            zap  ~
gap  [>1 space, nl]  rac  ]
hax  #               rit  )



—
~lodleb-ritrul
https://urbit.org

rmariani:
The line-break was a little confusing in yours, and it wasn't quite alphabetized (rob and rit), so here's my edit:


```
ace  [1 space]       hep  -            rob  }
ban  >               ket  ^            say  '
bar  |               lac  [            sig  ~
bat  \               led  <            tar  *
bus  $               lit  (            tec  `
cab  _               lob  {            tis  =
cen  %               lus  +            vat  @
col  :               mic  ;            wut  ?
com  ,               net  /            yel  "
dot  .               pad  &            zap  ~
gap  [>1 space, nl]  rac  ]        
hax  #               rit  )
```

ohAitch:
Also shadowing a carrier name seems impolite to say the least, which I think none of the existing syllables do; no matter how endearing pronouncing unix paths beginning with "signet" may be.

ohAitch:
Indefensible, and consequently, undefended. But presumably still getting
pushed next breach?

baudtack:
I, for one, strongly dislike both the net and bat versions of / and \

Sort of again bike-shedding but really not a fan. I much prefer paired names for those things as it makes it much easier to remember them. Bat and net break the association you get with fas and bas with forward slash and backslash. Is it really worth changing them because they are faster to say? Is that seriously a concern when discussing code? How fast you can say the runes?

cgyarvin:
Natural languages are full of broken patterns.  Well, ours is anyway.  It’s aesthetic.

The speed and feeling of speech is very important to how we perceive it — this (prosody in general) is a big part of why prosody works.  Also, try saying /usr/local/bin/vim 10x over with net vs fas.

Yes, I remain BD for this release.  But more Sulla than Augustus.  Whatever I leave screwed up the Republic will get a chance to fix in its lumbering Republican way — or, more likely, the converse.

ohAitch:
I mean yes, in practice even with "net" that's going to be pronounced "usr
local bin vim", possibly with longer pauses if the path segments themselves
are multi-word via dashes. (As will the URL I listed, "https net runner dot
net slash site" in non-urbit parlance)

cgyarvin:
Of course.  But there are plenty of cases where you do say it.

ohAitch:
And I agree that's a worthy goal. Just not worth, when you're trying to be
that explicit, having a separator confusable for a not-uncommon directory
name, TLD, galaxy name, etc

Are there no syllables with similar prosodic characteristics that are not
already taken by English?

cgyarvin:
Everything's taken by something.  And the association between "net" and "/"
is more than accidental -- "/" is a familiar symbol in Unix and the
Internet, so it actually goes well with a familiar sound.  And you can say
"net" three times in the time it takes to say "slash."

But why -- what color do *you* think we should paint the shed?

PR #325, opened 2018-12-26
Separate Fedora from RHEL instructions

jcmdln:
When compiling urbit on Fedora 29, I installed the packages that were listed and ran the included build scripts without error. This PR splits Fedora from the RHEL/CentOS instructions.


rmariani:
I will work this into a PR that accounts for the new docs structure.

PR #326, opened 2019-01-02
Clean up deps

mrdomino:
We don't need automake, autoconf, cmake, exuberant-ctags (did we ever?), libtool, re2c, or ragel any longer. Also, we do depend on zlib.

Note that ninja will print a warning about re2c on those platforms that need to build it from source, but it's fine since we aren't making changes to ninja's *.in.cc files.

I have not yet tested all the platforms (I did test Ubuntu/Debian and MacPorts OS X). I also cowardly refrained from changing the deps line for Fedora since maybe the Python build needs some of that stuff.

### DIFF
--- a/docs/using/install.md
+++ b/docs/using/install.md
@@ -31,24 +31,19 @@ with us at [urbit.org/stream](https://urbit.org/stream).
 Urbit depends on:
 
 ```
-autoconf
-automake
 C compiler (gcc or clang)
-cmake
 curses
 git
 gmp
 libcurl
 libsigsegv
-libtool
 libuv
 meson
 ninja
 openssl
 pkg-config
 python2
-ragel
-re2c
+zlib
 ```
 
 ### Instructions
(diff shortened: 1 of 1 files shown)


joemfb:
@mrdomino, could you make PRs to urbit/urbit and urbit/arvo to remove these unnecessary dependences from `.travis.yml`? Might also be a good sanity-check that they are indeed unused.

ohAitch:
autoconf is libuv iirc; which suggests possibly splitting out a dependency
list for "clean" builds(the official submodules), while keeping the "I
trust whatever versions my distro happens to provide" installation
instructions correct.

re2c is from a defunct regex jet, if that's gone probably just safe to
remove yeah

mrdomino:
Looks like we're building libuv with meson these days though.

rmariani:
Thanks for the useful intel! This page doesn't exist anymore, however.

joemfb:
Rob, I think these updates need to be forward-ported to https://urbit.org/docs/getting-started/. The changes themselves are important.

ohAitch:
Oh, was this not a "merged out of band to the moved file" PR closure?

rmariani:
Yeah I am doing that manually @ohAitch @joemfb. 

Are the actual depedencies any different now? Like in: 

```
sudo apt-get install autoconf automake cmake exuberant-ctags g++ git libcurl4-gnutls-dev libgmp3-dev libncurses5-dev libsigsegv-dev libssl-dev libtool make openssl pkg-config python python3 python3-pip ragel re2c zlib1g-dev
```

joemfb:
Yes, the actual dependencies are different. The values in this PR are correct. Thanks!

PR #340, opened 2019-01-17
Remove explicit meson version from install instructions

Fang-:
Current release-candidate fails to build (in libent) with meson 0.29, but builds fine with latest (0.49).

Maybe we want to pin to 0.49 instead of going with "whatever is latest"? But we already weren't specifying a specific version in the macOS or Arch install instructions, so...

### DIFF
--- a/getting-started.udon
+++ b/getting-started.udon
@@ -177,7 +177,7 @@ urbit
 sudo apt-get update
 sudo apt-get install autoconf automake cmake exuberant-ctags g++ git libcurl4-gnutls-dev libgmp3-dev libncurses5-dev libsigsegv-dev libssl-dev libtool make openssl pkg-config python python3 python3-pip ragel re2c zlib1g-dev
 sudo -H pip3 install --upgrade pip
-sudo -H pip3 install meson==0.29
+sudo -H pip3 install meson
 
 # we need ninja 1.5.1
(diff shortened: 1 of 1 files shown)


getting-started.udon
```diff
@@ -222,8 +222,7 @@ popd
 
 curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py
 sudo /usr/local/bin/python3 get-pip.py
-# downgrade meson to avoid dependencies that redhat can't meet
-sudo -H /usr/local/bin/pip3 install meson==0.29
```
joemfb: Based on that comment, it seems like this might not work.

joemfb:
I think this change might invalidate the Redhat installation instructions. I don't know if it's important to support legacy-meson anymore -- LTS Ubuntu (which I think was the catalyst) has been upgraded.

Fang-:
Yeah, the redhat comment was noted offline. Not sure what to do about it though. It would need to be able to run latest meson to be able to build latest urbit... Supporting legace meson seems very undesirable.

ohAitch:
Meson delenda etc

joemfb:
I think we either need to restore support for legacy-meson, or remove instructions that will be wrong without it. And if we remove legacy-meson, we should remove it from `meson.build` as well, there's quite a bit of code that goes into it.

PR #346, opened 2019-01-18
More text fixes to make getting-started match bridge.

jfranklin9000:


jfranklin9000:
One nit: `https://github.com/urbit/bridge/releases` runs Bridge on port 5000
but `https://urbit.org/docs/getting-started/` has a screenshot of it running
on port 3000.

PR #358, opened 2019-01-26
Fix confusing my/by map typo

courajs:


rmariani:
Thanks for catching this.

PR #365, opened 2019-01-30
Adds links to runes.udon

rmariani:
Also adds descriptions.


rmariani:
Agreed. I made the change.

PR #368, opened 2019-02-01
updates old-zuse names

rmariani:
I erred on the side of being overly specific within vane documentation: I replaced things like `++ankh` with `++ankh:clay`, and so forth


baudtack:
LGTM

PR #369, opened 2019-02-01
adds 'decorum' section

rmariani:
Gently massaged from the google doc

### DIFF
--- a/introduction/contributing.udon
+++ b/introduction/contributing.udon
@@ -13,51 +13,31 @@ If you're interested in contributing, but don't know where to start, feel free t
 
 ## Code guidelines
 
-Our official guide to contributing can be found
-[here](https://github.com/urbit/urbit/blob/master/CONTRIBUTING.md),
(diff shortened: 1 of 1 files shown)


baudtack:
Maybe it's just me but maybe it would be worth making clear more than once that these are only rules that affect the tlon run parts of the network and that not only will tlon not police anything else, but we're literally incapable of doing so.

ohAitch:
Decorum request: don't mix hard-wrapped and non-hard-wrapped lines in the
same document :)

rmariani:
I'll make that clearer @baudtack

ohAitch:
The "incapable of doing so" part?

rmariani:
Yeah, upon re-reading it, I think that making it any clearer would be unnecessary and forced

rmariani:
@ohAitch I fixed the wrapping :)

rmariani:
Ok, I added something to drive the point home.

PR #370, opened 2019-02-01
adds 'next' link at bottom of lessons

rmariani:
Also did some minor reordering


baudtack:
Did  you test that these links work? They look like they should but...

rmariani:
They are all accurate

baudtack:
lgtm

PR #372, opened 2019-02-02
Remove meson version from bash install script

hoclun-rigsep:
I couldn't build on Debian 9 with current script because the meson version was so old.


ohAitch:
Aw I saw "remove meson" and got excited ):

rmariani:
Thanks for catching that, @hoclun-rigsep

PR #374, opened 2019-02-03
Point out that stack traces can be activated in talk.

jfranklin9000:


ohAitch:
"(e.g. stack traces)" maybe?

jfranklin9000:
OK, I changed it.

PR #384, opened 2019-02-08
Fix Gall text/hyperlink in glossary.

jtobin:
This fixes a typo in the Gall section that had referred to Ford, and also updates the "more info" link to match the newer scheme used throughout the docs.


rmariani:
Thanks!

PR #387, opened 2019-02-08
fixes typos in bus.udon

rmariani:
I just took out the `?(@f0 @f1)` because it seemed like confusing gibberish (is `@f` even an aura?)


ohAitch:
It is even an aura, and I think ?(%.y %.n) is literally what you get when
you type spear(or dojo "?") a loobean

joemfb:
`@f` is an aura. Like `@n` (`~`), it's a particularly strange aura. `?` is `?(%& %|)`, which is `?(_`@f`0 _`@f`1), if that were allowed to be run, but it nest-fails. I think all casts to `@f` are disabled. I'm not sure what the docs implications of this should be. It's definitely a known rough corner.

PR #399, opened 2019-02-21
Restore proper pronunciation guide

baudtack:
Now that ~sorreg is gone, let us exorcise the last reglossing that everyone hates

### DIFF
--- a/learn/arvo/hoon/hoon-syntax.udon
+++ b/learn/arvo/hoon/hoon-syntax.udon
@@ -36,18 +36,18 @@ Hoon source files are composed almost entirely of the [printable ASCII character
 Hoon code makes heavy use of non-alphanumeric symbols.  Reading off code aloud using the proper names of ASCII symbols is tedious, so we've mapped syllables to symbols:
 
 ```
-ace  [1 space]       hep  -            rob  }
-ban  >               ket  ^            say  '
-bar  |               lac  [            sig  ~
-bat  \               led  <            tar  *
-bus  $               lit  (            tec  `
-cab  _               lob  {            tis  =
-cen  %               lus  +            vat  @
-col  :               mic  ;            wut  ?
-com  ,               net  /            yel  "
-dot  .               pad  &            zap  !
-gap  [>1 space, nl]  rac  ]
-hax  #               rit  )
+ace [1 space]   gal <               pal (
+bar |           gap [>1 space, nl]  par )
+bas \           gar >               sel [
+buc $           hax #               mic ;
+cab _           hep -               ser ]
+cen %           kel {               sig ~
+col :           ker }               soq '
+com ,           ket ^               tar *
+doq "           lus +               tec `
+dot .           pam &               tis =
+fas /           pat @               wut ?
+zap !
 ```
 
 Learning and using these names of non-alphanumeric symbols is entirely optional; you can become an expert Hoon programmer without them.  Many Hoon programmers find them helpful, however, particularly when talking to someone else who knows them.
--- a/learn/arvo/hoon/hoon-syntax.udon
+++ b/learn/arvo/hoon/hoon-syntax.udon
@@ -44,13 +44,15 @@ cab _           hep -               ser ]
 cen %           kel {               sig ~
 col :           ker }               soq '
 com ,           ket ^               tar *
-doq "           lus +               tec `
+doq "           lus +               tic `
 dot .           pam &               tis =
 fas /           pat @               wut ?
 zap !
 ```
 
-Learning and using these names of non-alphanumeric symbols is entirely optional; you can become an expert Hoon programmer without them.  Many Hoon programmers find them helpful, however, particularly when talking to someone else who knows them.
+Learning and using these names of non-alphanumeric symbols is entirely optional; you can become an expert Hoon programmer without them.  Many Hoon programmers find them helpful, however, par
+
+ularly when talking to someone else who knows them.
 
 Note that the list includes two separate whitespace forms: `ace` for a single space; `gap` is either 2+ spaces or a newline.  In Hoon, the only whitespace significance is the distinction between `ace` and `gap` -- i.e., the distinction between one space and more than one.
 
--- a/reference/hoon-expressions/rune/buc.udon
+++ b/reference/hoon-expressions/rune/buc.udon
@@ -1,11 +1,11 @@
 :-  :~  navhome/'/docs/'
         next/'true'
         sort/'12'
-        title/'Structures $ ("bus")'
+        title/'Structures $ ("buc")'
     ==
 ;>
 
-# Structures $ ("bus")
+# Structures $ ("buc")
 
 The `$` family of runes is used for defining custom types.  Strictly speaking,
 these runes are used to produce 'structures'.  A structure is a compile-time
@@ -61,7 +61,7 @@ Irregular: `*` makes `%noun`, `^` makes `%cell`, `?` makes
 
 ;=
   ;h3
-    ;div(id "buscab"): $_ "buscab"
+    ;div(id "buccab"): $_ "buccab"
   ==
 ==
 
@@ -101,7 +101,7 @@ and produces its *example* instead.
 
 ;=
   ;h3
-    ;div(id "buscen"): $% "buscen"
+    ;div(id "buccen"): $% "buccen"
   ==
 ==
 
@@ -136,7 +136,7 @@ be an infinite loop!
 
 ;=
   ;h3
-    ;div(id "buscol"): $: "buscol"
+    ;div(id "buccol"): $: "buccol"
   ==
 ==
 
@@ -174,7 +174,7 @@ Irregular (structure mode): `[a b c]` is `$:(a b c)`.
 
 ;=
   ;h3
-    ;div(id "bushep"): $- "bushep"
+    ;div(id "buchep"): $- "buchep"
   ==
 ==
 
@@ -193,7 +193,7 @@ Regular: *2-fixed*.
 
 ##### Discussion
 
-Since a `$-` reduces to a [`$_`](#buscab), it is not useful for normalizing, just for typechecking.  In particular, the existence of `$-`s does *not* let us send gates or other cores over the network!
+Since a `$-` reduces to a [`$_`](#buccab), it is not useful for normalizing, just for typechecking.  In particular, the existence of `$-`s does *not* let us send gates or other cores over the faswork!
 
 ##### Examples
 
@@ -206,7 +206,7 @@ Since a `$-` reduces to a [`$_`](#buscab), it is not useful for normalizing, jus
 
 ;=
   ;h3
-    ;div(id "busket"): $^ "busket"
+    ;div(id "bucket"): $^ "bucket"
   ==
 ==
 
@@ -244,7 +244,7 @@ Regular: *2-fixed*.
 
 ;=
   ;h3
-    ;div(id "bussig"): $~ "bussig"
+    ;div(id "bucsig"): $~ "bucsig"
   ==
 ==
 
@@ -311,7 +311,7 @@ Using `$~`:
 
 ;=
   ;h3
-    ;div(id "busvat"): $@ "busvat"
+    ;div(id "bucvat"): $@ "bucvat"
   ==
 ==
 
@@ -353,7 +353,7 @@ Example:
 
 ;=
   ;h3
-    ;div(id "bustis"): $= "bustis"
+    ;div(id "buctis"): $= "buctis"
   ==
 ==
 
@@ -392,7 +392,7 @@ p=%foo
 
 ;=
   ;h3
-    ;div(id "buswut"): $? "buswut"
+    ;div(id "bucwut"): $? "bucwut"
   ==
 ==
 
@@ -417,8 +417,8 @@ Irregular: `?(%foo %baz)` is `$?(%foo %baz)`.
 ##### Discussion
 
 For a union of atoms, a `$?` is fine.  For more complex nouns,
-always try to use a [`$%`](#buscen), [`$@`](#busvat) or
-[`$^`](#bushep), at least if you expect your structure to be used as a
+always try to use a [`$%`](#buccen), [`$@`](#bucpat) or
+[`$^`](#buchep), at least if you expect your structure to be used as a
 normalizer.
 
 ##### Examples
--- a/reference/hoon-expressions/rune/sig.udon
+++ b/reference/hoon-expressions/rune/sig.udon
@@ -179,11 +179,11 @@ wing `q`.  We assume the parent is already registered -- as it
 must be, if (a) we registered it on creation, (b) the child was
 created by an arm defined on the parent.
 
-(Cores are actually managed by their formula/bastery.  Any
+(Cores are actually managed by their formula/battery.  Any
 function call will create a new core with a new sample, but
-basteries are constant.  But it is not sufficient to match the
-bastery -- matching the semantics constrains the payload as well,
-since the semantics of a bastery may depend on any parent core
+batteries are constant.  But it is not sufficient to match the
+battery -- matching the semantics constrains the payload as well,
+since the semantics of a battery may depend on any parent core
 and/or payload constant.)
 
 The purpose of registration is always performance-related.  It
--- a/reference/hoon-expressions/rune/sig.udon
+++ b/reference/hoon-expressions/rune/sig.udon
@@ -86,7 +86,7 @@ ford: build failed
 
 ;=
   ;h3
-    ;div(id "sigbus"): ~$ "sigbus"
+    ;div(id "sigbuc"): ~$ "sigbuc"
   ==
 ==
 
@@ -179,11 +179,11 @@ wing `q`.  We assume the parent is already registered -- as it
 must be, if (a) we registered it on creation, (b) the child was
 created by an arm defined on the parent.
 
-(Cores are actually managed by their formula/battery.  Any
+(Cores are actually managed by their formula/bastery.  Any
 function call will create a new core with a new sample, but
-batteries are constant.  But it is not sufficient to match the
-battery -- matching the semantics constrains the payload as well,
-since the semantics of a battery may depend on any parent core
+basteries are constant.  But it is not sufficient to match the
+bastery -- matching the semantics constrains the payload as well,
+since the semantics of a bastery may depend on any parent core
 and/or payload constant.)
 
 The purpose of registration is always performance-related.  It
@@ -318,7 +318,7 @@ This should work fine:
 
 ;=
   ;h3
-    ;div(id "signet"): ~/ "signet"
+    ;div(id "sigfas"): ~/ "signet"
   ==
 ==
 
--- a/reference/glossary.udon
+++ b/reference/glossary.udon
@@ -420,18 +420,18 @@ information in speech. [`Runes`](#rune) are composed of two glyphs.
 Below is a complete list of glyphs with their associated pronunciations.
 
 ```
-ace  [1 space]       hep  -            rob  }
-ban  >               ket  ^            say  '
-bar  |               lac  [            sig  ~
-bat  \               led  <            tar  *
-bus  $               lit  (            tec  `
-cab  _               lob  {            tis  =
-cen  %               lus  +            vat  @
-col  :               mic  ;            wut  ?
-com  ,               net  /            yel  "
-dot  .               pad  &            zap  !
-gap  [>1 space, nl]  rac  ]
-hax  #               rit  )
+ace [1 space]   gal <               pal (
+bar |           gap [>1 space, nl]  par )
+bas \           gar >               sel [
+buc $           hax #               mic ;
+cab _           hep -               ser ]
+cen %           kel {               sig ~
+col :           ker }               soq '
+com ,           ket ^               tar *
+doq "           lus +               tec `
+dot .           pam &               tis =
+fas /           pat @               wut ?
+zap !
 ```
 
 ### Hall
--- a/reference/glossary.udon
+++ b/reference/glossary.udon
@@ -428,7 +428,7 @@ cab _           hep -               ser ]
 cen %           kel {               sig ~
 col :           ker }               soq '
 com ,           ket ^               tar *
-doq "           lus +               tec `
+doq "           lus +               tic `
 dot .           pam &               tis =
 fas /           pat @               wut ?
 zap !
--- a/reference/hoon-expressions/rune.udon
+++ b/reference/hoon-expressions/rune.udon
@@ -56,7 +56,7 @@ Runes used for making function calls in Hoon.
 
 Runes that let us adjust types without violating type constraints.
 
-### [`$ bus` (Structures)](bus)
+### [`$ buc` (Structures)](buc)
 
 Runes used for defining custom types.
 
--- a/reference/hoon-expressions/rune/lus.udon
+++ b/reference/hoon-expressions/rune/lus.udon
@@ -69,7 +69,7 @@ Notice that `p.q` has the label `%numbers`.  Contrast with:
 
 ;=
   ;h3
-    ;div(id "lusbus"): +$ "lusbus"
+    ;div(id "lusbuc"): +$ "lusbuc"
   ==
 ==
 
--- a/reference/hoon-expressions/rune/mic.udon
+++ b/reference/hoon-expressions/rune/mic.udon
@@ -176,7 +176,7 @@ Fails because not a fixpoint:
 
 ;=
   ;h3
-    ;div(id "micnet"): ;/ "micnet"
+    ;div(id "micfas"): ;/ "micfas"
   ==
 ==
 
--- a/reference/hoon-expressions/rune/tis.udon
+++ b/reference/hoon-expressions/rune/tis.udon
@@ -381,7 +381,7 @@ Regular: *3-fixed*.
 
 ;=
   ;h3
-    ;div(id "tisnet"): =/ "tisnet"
+    ;div(id "tisfas"): =/ "tisfas"
   ==
 ==
 
@@ -418,7 +418,7 @@ Regular: *3-fixed*.
 ##### Discussion
 
 `p` can be either a name or a name=type.  If it's just a name,
-`=/` ("tisnet") "declares a type-inferred variable."  If it has a type, `=/`
+`=/` ("tisfas") "declares a type-inferred variable."  If it has a type, `=/`
 "declares a type-checked variable."
 
 ##### Examples


rmariani:
Only a few of these people took issue with, `buc` being the obvious offender. I see that you kept `mic`, though, so we're on the same page.

baudtack:
the only ones that really  annoy me are buc, gar, gal, fas, and bas.

cgyarvin:
Okay to change these of course, but if you’re going to bikeshed, you need to bikeshed multilaterally I think!  No way around it.

Some things are good, some are bad, some are ok but there are better things I didn’t think of.  It’s worth one final collective design pass, as is the glossary in general.  Think of it as a well constrained team building exercise.  Clearly can never be done again!

Everyone has an opinion on this subject.  A conversation (which I’ll stay out of) is not a waste of time, although it’s always ok to ask me what the hell I was thinking.

And no one person should ever have the power I once had (or thought I had)... or if they do, they have to wrest it from all others by sheer will and charisma...

msutherl:
@rmariani is it pronounced "mick" or like "mic." as in microphone?

ohAitch:
:D thank you thank you thank you, this is a PR I was procrastinating on a
lot, up to the grudgingly keeping mic bc it was solving an actual problem

lmk if any assistance would be helpful on the hoon.hoon / jet side of
things, the network is still doing regular breaches so it shouldn't be too
hard yet

Re: mick vs mic, the former, the latter is spelled "myc" ;)

ohAitch:
Hm okay not exactly the same, pal par were pel per before the reglossing:
https://github.com/urbit/arvo/blob/master@%7B2019-1-1%7D/sys/hoon.hoon#L4768

Fang-:
> grudgingly keeping mic bc it was solving an actual problem

Let us hear it! I dislike mic and think sem is way easier. If we go with the former, "mike"/"myk" is the better pronunciation.

I always thought `` ` `` was "tic", not "tec". Any reason that shouldn't be the case?

No strong opinions on pal par vs pel per, just want to note that people have been mistakenly remembering them as pal par since forever. `a` and `e` sounds might both be a little bit overloaded though. It's probably fine, but lit rit wouldn't be *terrible*.

cgyarvin:
One factor worth paying attention to is the sonic weight of the syllable — basically just how long it takes to say.  I didn’t think about this the first time around and regretted it.

To test this, say the syllable five times in a row as fast as you can.  Try with myk/mihc, par/rit, etc etc.  Makes a huge difference...

ohAitch:
The actual problem is sem/cen; which you can try to tackle from the cen
angle instead but mic is a lot more straightforward. And it's not myc,
seMIColon doesn't rhyme with "semi"

Re: sonic weight, perperper is definitely a lot easier than ritritrit, I
keep fumbling the tr.

At any rate I disagree that an extensive discussion would be helpful here,
though it would certainly be entertaining: everyone is going to have
different opinions on details, and in the meantime the 2019Q1 version would
stick around in the kernel. "Don't randomly change things" is the only
effective anti-bikeshedding measure; and given the new version has landed
poorly, "the system people have been working with for years" is the closest
of approximation of "fundamentally not changed" that we have. (Except for
sem/cen, which people while working with the thing for years have stumbled
on continuously, and is the kind of judicious narrow polish that being
not-yet-fully-frozen is for)

cgyarvin:
It’s definitely not real bikeshedding unless Anton is involved

Fang-:
> And it's not myc, seMIColon doesn't rhyme with "semi"

This is news to me. I thought both were valid. But I'm also not a native speaker! Maybe the three-letter equivalent of "meek" is more appropriate? But yeah, "mic" seems fine here then.

> To test this, say the syllable five times in a row as fast as you can.

But that only tests the syllable against itself, not against an actual context. I maintain that this exercise, and pronunciation speed in general (rather than clarity and ease (though ease often leads to speed)) are poor metrics.  
Really any pronunciation metrics are going to be different for people whose native tongue is not English, but given the context designing for an English-speaking if not -native audience is best.

cgyarvin:
Then try saying any Unix or urbit path with net versus fas.

Weight of a syllable is a very real thing and the feeling of heaviness/clumsiness it gives can be quite strong I think.

baudtack:
My  argument for  restoring most of these is that their function is to aid in communication about hoon. When the rune pronunciations are hard to remember and have very little helpful pronunciation to hang your thoughts off of, they are useless. It should not be any harder  to learn about hoon than it already is and the reglossing muddied the water for these. The most important part is are they unique from each other and do the pronunciation have some hook you can use to remember them. buc -> bus removes the  context of $ being the sign for USD, colloquially referred to as a buck. While net has a stronger relationship to / than fas does it also breaks the symmetry of fas, bas. vat where you don't pronounce the v is just a non sequitur, at least to me. There may be some reasoning behind it but even if there is I don't see it as a  substantive improvement  over the old pat.


I have no feelings about pal par vs pel per. I used pal and par because I just grabbed the first old table I could find because editing that ascii table is a pain in the ass. Both versions are equally nonsense but they maintain the symmetry so I am happy with either.

philipcmonk:
I'm on board with mic.  cen/sem was a real problem.

Gal/gar are drastically better than ban/led because they convey the symmetry, and we use that symmetry to hang other concepts on (like `=<` versus `=>`)

I learned pel and per, but I kind of like pal and par because they're easier to say -- per in particular doesn't sound like a rune, it sounds like a small word in in a phrase, so if you say it alone, people don't understand you meant the rune and you have to repeat it.

Net also sounds like a word.  Fas clearly indicates a rune, and so hearing a unix path makes sense.

As far as I'm aware backtick has always been tec.

jtobin:
Just to officially weigh in: I think either scheme, or mix thereof, is *perfectly fine*.  Both schemes simply do a great job of solving the pronunciation issue.



What is important is that the names stop changing.  So I would be content sticking with the old names -- which have been around longer, and, by my above rationale, didn't need changing anyway -- or the new ones, which are at-present official, and thus don't need to be changed either.  Flip a coin and go with the result! 😄

eglaysher:
The most important metric of a syllable is how mnemonic it is. Mnemonic learning needs to be prioritized so people can learn things in the first place.

In as much as sem/cen has a confusion issue, the only reason mic is acceptable is because it is seMIColon. I have no idea where to mentally hang 'net' as a slash, while I get fas as Forward slASh. Changing buc to bus breaks the mnemonic with the USD. Etc.

Arguments of the form "it's easier to say x several times fast than y" don't sway me when x is harder to memorize than y in the first place.

ohAitch:
I think curtis just pronounces the letter "t" more quickly than me - timing things in audacity, I can actually get 4.7 "fas"es in a second, and only 4.5 "net"s; 7 vs 6.5 if I lower my standards on [enunciation](https://en.wikipedia.org/wiki/T-glottalization). Which is close enough to not be worth changing it _back_ over weight, but as mentioned above distinctness and memorability is more important here: obviously a system where all the glyph mneumonics were just "a" would be both easier to say and somewhat useless.

/ in particular is a _separator_, so being able to pick it out from surrounding sounds is important. fas is made of "f" and "s" which are very distinctive if you've ever seen them on a spectrogram; and a vowel that doesn't occur frequently in unix paths, combining well with pat which also acts as a separator. net is very indistinct - perhaps this informs the \~perceived\~ speaking time?

But ultimately all of the linguistic digression doesn't matter, fas was here first and people like it more - I expect if it had been "net" the whole time the above would still not be worth changing it. (And we wouldn't have ended up with a ~net, which as a galaxy name uh, occurs in unix paths)

cgyarvin:
Bear in mind that the hoon type has to match...

ohAitch:
This is why my first comment included "lmk if any assistance would be helpful on the hoon.hoon / jet side of things, the network is still doing regular breaches so it shouldn't be too hard yet" yes

tho I'm open to the exercise of "how do we do this sort of migration for real" tbh

learn/arvo/hoon/hoon-syntax.udon
```diff
@@ -36,18 +36,18 @@ Hoon source files are composed almost entirely of the [printable ASCII character
...
 
 ```
-ace  [1 space]       hep  -            rob  }
-ban  >               ket  ^            say  '
-bar  |               lac  [            sig  ~
-bat  \               led  <            tar  *
-bus  $               lit  (            tec  `
-cab  _               lob  {            tis  =
-cen  %               lus  +            vat  @
-col  :               mic  ;            wut  ?
```
ohAitch: though I will note stylistically that "mic" is alphabetically after "lus", not "sel"

ohAitch:
We should probably also explicitly notate that the official paren pronunciation is pel pər sel sər kel kər; I guess that might not go well with ascii purism?

reference/hoon-expressions/rune/buc.udon
```diff
@@ -311,7 +311,7 @@ Using `$~`:
 
 ;=
   ;h3
-    ;div(id "busvat"): $@ "busvat"
+    ;div(id "bucvat"): $@ "bucvat"
```
pilfer-pandex: Shouldn't this be `pat` instead of `vat`, to be consistent with other changes?

reference/hoon-expressions/rune/sig.udon
```diff
@@ -179,11 +179,11 @@ wing `q`.  We assume the parent is already registered -- as it
 must be, if (a) we registered it on creation, (b) the child was
 created by an arm defined on the parent.
 
-(Cores are actually managed by their formula/battery.  Any
+(Cores are actually managed by their formula/bastery.  Any
```
pilfer-pandex: Typo?

reference/hoon-expressions/rune/sig.udon
```diff
@@ -179,11 +179,11 @@ wing `q`.  We assume the parent is already registered -- as it
...
 function call will create a new core with a new sample, but
-batteries are constant.  But it is not sufficient to match the
-battery -- matching the semantics constrains the payload as well,
-since the semantics of a battery may depend on any parent core
+basteries are constant.  But it is not sufficient to match the
+bastery -- matching the semantics constrains the payload as well,
```
pilfer-pandex: It's a place to go to make sure your chicken is super-moist.

pilfer-pandex:
To make it easier to evaluate this proposal, I've compiled a table:

| Symbol | Curtisian Revisionist Name | Kerschnerian Restorationist Name |
| ---- | ---- | ---- |
| ` ` | ace | ace |
| `>` | ban | gar |
| `\|` | bar | bar |
| `\` | bat | bas |
| `$` | bus | buc |
| `_` | cab | cab |
| `%` | cen | cen |
| `:` | col | col |
| `,` | com | com |
| `.` | dot | dot |
| `  ` | gap | gap |
| `#` | hax | hax |
| `-` | hep | hep |
| `^` | ket | ket |
| `[` | lac | sel |
| `<` | led | gal |
| `(` | lit | pal |
| `{` | lob | kel |
| `+` | lus | lus |
| `;` | mic | sem |
| `/` | net | fas |
| `&` | pad | pam |
| `]` | rac | ser |
| `)` | rit | par |
| `}` | rob | ker |
| `'` | say | soq |
| `~` | sig | sig |
| `*` | tar | tar |
| `` ` `` | tec | tec |
| `=` | tis | tis |
| `@` | vat | pat |
| `?` | wut | wut |
| `"` | yel | doq |
| `!` | zap | zap |


And just the changes, brackets:

| S | C | K |
| ---- | ---- | ---- |
| `(` | lit | pal |
| `)` | rit | par |
| `[` | lac | sel |
| `]` | rac | ser |
| `{` | lob | kel |
| `}` | rob | ker |
| `<` | led | gal |
| `>` | ban | gar |

Nonbracket changes:

| S | C | K |
| ---- | ---- | ---- |
| `\` | bat | bas |
| `$` | bus | buc |
| `;` | mic | sem |
| `/` | net | fas |
| `\` | bat | bas |
| `&` | pad | pam |
| `@` | vat | pat |
| `'` | say | soq |
| `"` | yel | doq |

reference/hoon-expressions/rune/sig.udon
baudtack: The perils of using sed.

pilfer-pandex:
Other design considerations potentially worth considering:

- Avoid names that differ only in the voicing of a single consonant, especially the final consonant. So having both (revisionist) pad as `&` and (restorationist) pat as `@` simultaneously in the final nomenclature would be unacceptable
- Consider what happens when consonants run into each other in the names of runes. So for example, consider `$:` "buccol"; this has only one consonant in the middle rather than the usual two, which IMHO is unaesthetic. More problematic are cases where you have consonants that differ only in voicing next to each other, because then one can assimilate the other: e.g. `\"` "batdoq" is probably realized as "batok".

It's telling that many of these examples rely on mixing names from the two proposals.

ohAitch:
Two notes on the restorationist table:
1) the restorationist parens are pel/per, not pal/par, for consistency with
the other brackets
2) the proposed restoration is not 100% pure; that is, the revisionist
sem->mic is retained

Really confused where you got a "restorationist" list that misses both of
these tbh: the commit in this PR has 2), and the literal
hoon.hoon-two-months-ago I linked earlier has 1)

baudtack:
@pilfer-pandex thanks for building that! Much easier to see now. Also will probably convert the existing table to an actual table as long as udon supports it. Not sure if it does or not.

philipcmonk:
> * Consider what happens when consonants run into each other in the names of runes.

This is somewhat mitigated in my speech by pronouncing digraphs with each syllable stressed (ie as two words with a slight pause between them, not a one-word trochee).  I can easily distinguish "buc col" from "buc ol" because of the pause.

pilfer-pandex:
I like the new brackets, except for "ban" which comes out of nowhere. I guess it's to avoid the word "red"? If so, I'd actually be ok with the symmetry-breaking:

| Symbol | Pandex's Synthesis |
| --- | --- |
| `(` | lit |
| `)` | rit |
| `[` | lac |
| `]` | rac |
| `{` | lob |
| `}` | rob |
| `<` | gal |
| `>` | gar |

An argument for this is that `<>` function differently as symbols in hoon from the rest.

For the other symbols, I would do:

| S  | P
| --- | ---
| `$` | bus |
| `;` | mic |
| `/` | fas |
| `\` | bas |
| `&` | pam |
| `@` | pat |
| `'` | say |
| `"` | yel |

I feel more strongly about `;/\&@` than the rest.

I kind of like bus because it's what I learned, sounds better to me in some rune names than buc does, and is less confused with a curse word. The mnemonic for it is that it looks like an s. But this is not a hill I want to die on.

I pronounce "mic" with an [/ɪ/](https://en.wikipedia.org/wiki/Near-close_front_unrounded_vowel). "Mic" is cool because it's used in "make" runes. The previous confusion was not good.

As a further change to a thing both proposals agree with, per @Fang-:

| Symbol | Old | Pandex
| --- | --- | ---
| `` ` `` | tec | tic

Finally, I must confess I am a bit squeamish about the name "sig." I've thought of "nul" and "sys" as alternatives, but I'm not in love with either. This might be too late to change though.

ohAitch:
I think

> Avoid names that differ only in the voicing of a single consonan

is too important to allow the new parens tbh. The old ones have the
wonderful property that while they look very visually similar due to
differing by one letter in text, when spoken they also differ by vowel;
here the vowel is being used to redundantly discriminate type of paren,
while direction will be impossible to distinguish in some accents.

The new quotes are cute, but similarly phonetic similarity seems more
useful than conceptual similarity; and single vs double is a lot better
mnemonic than " being "louder" somehow

Along those lines, may I propose tiq as the official spelling of "tic"?
It's kind of a quote, but kind of not ^-^

pilfer-pandex:
Anton, /l/ and /r/ don't differ in voicing — they're both voiced.

cgyarvin:
You’ll notice that < and > are not mostly used as symmetric pairs and are the only paired characters used in runes.

Also notice: some languages, like Japanese, have l/r confusion.  I decided not to worry about this, because Pearl Harbor.  But that was a long time ago and FDR was up to some shady stuff too, so another idea was an alternate pronunciation of r, like z.

ohAitch:
Obvious trolls aside, they are in fact both liquid constants even in 100%
canonical English. Not as bad a m/n by any means, but definitely not worth
removing an existing distinguishing factor from.

"Used in runes" seems like kind of a non sequitur, like yes brackets are
irregular form twigs and not regular form twigs? I would not describe
irregular forms as particularly rarely used so as to merit their
de-mnemoniciziation; and for that matter bas is even less rune-shaped, that
doesn't mean it's good to break the parallel with fas.

0x70b1a5:
Japanese learners of English will deal with L/R pronunciation in many other parts of their lives; I'm sure one more won't hurt that much. Besides, futzing with the blissful intuition of `gal` / `gar` for a sub-sub-subsection of our userbase renders only minimal marginal utilitarian points to our rationalist punch cards, and the practice sets a poor precedent.

cgyarvin:
Good analysis.  I mean, this wasn’t my main reason for the gal/gar change — but it’s also why all these other l/r issues don’t bother me that much.  

At the same time, substituting a consonant creates a regional dialect, which is also a perfectly legitimate and even aesthetic outcome.  Variation is lovely!

ohAitch:
To be clear, the blissful gal/gar is the version in this PR, and led/ban is
the version currently roaming the network. Good, but somewhat misleading,
analysis.

0x70b1a5:
Comment is intended as endorsement of PR.

bsima:
Having learned both the restorationist version in this PR and curtis's revisionist version, I'm actually in favor of the latter. I find it more mnemonic, for example:

- `;` kinda looks like a microphone
- `]` if you rotate it kinda looks like a rack, hence `rac`.
- symmetric pairs like `lit`/`rit` are easy to remember because the first letter are the same as `left`/`right`. With `pal`/`par` etc I found it less obvious which is left/right.
- `/` as `net` is useful when reading paths because it sounds like `next`, e.g. `/home/user/src` is read almost like `next home next user next src`, as if you're traversing into the tree.
- `say` and `yel` are both strings, used for text, which is much more mnemonic than `soq`/`doq` which I couldn't memorize at all.

By "learned" i mean i actually used a spaced-repetition program (org-drill) to commit them to memory and I found the curtisonian version easier to memorize and pronounce.

baudtack:
I've tried to bring together a set that the majority of people in this thread  seem to be in favor of. I'm sure this could be discussed to death (which the cynical part of me suspects was @cgyarvin 's point in commenting on it at all..) But we should solve this once  and for all. Please indicate via up and down thumb emotes if you can live with this set. Not if it's perfect, but if you can live with it.

| S | K |
| ---- | ---- |
| `(` | pal |
| `)` | par |
| `[` | sel |
| `]` | ser |
| `{` | lob |
| `}` | rob |
| `<` | gal |
| `>` | gar |
| `$` | buc |
| `;` | mic |
| `/` | fas |
| `\` | bas |
| `&` | pam |
| `@` | pat |
| `'` | say |
| `"` | yel |

pilfer-pandex:
We're voting?

baudtack:
> We're voting?

We're taking a poll to see what happens. we really have no good mechanisms in place currently for deciding things by consensus  instead of by executive fiat which we need to start developing.

Fang-:
>say and yel

I can live with this, but... what the heck man, your list was almost Actually Good.

>we really have no good mechanisms in place currently for deciding things by consensus instead of by executive fiat which we need to start developing.

Do we? Pretty sure we don't want to take votes on language decisions. I wouldn't trust my own vote, after all! This is why we have @pilfer-pandex take ownership of the language, right?

pilfer-pandex:
I can exercise authority here if people really want me to. That said, there are other hills that are more important for me to die on.

baudtack:
> @pilfer-pandex take ownership of the language, right?

Urbit and by extension hoon is supposed to be community property.  Let's act like it or admit that it isn't.

cgyarvin:
I would suggest that it’s probably too easy to vote on an entire slate — it’s just giving power to whoever picks the slate.  Also, not sure about counting heads.

Bear in mind that if Hoon does not succeed, none of this matters; and if it does, the amount of energy invested in a few early adopters undoing their bindings, arguing about design details one by one, etc, is very very small compared to what people will put into it.

ohAitch:
I believe I mentioned that being that point of commenting at all in my
first response :P

What fang said wrt say/yel; in general my principle here is "actually the
old set, with the exception of mic". I can live with pal/par.

(And I don't think *anyone* has expressed support for mixed old and new
parens, so I'm not sure how pal par sel ser /lob rob/ happened?)

ohAitch:
Like in general, the PR said "restore", I think a reasonable reaction to
seeing the PR would have been "oh good they're being changed back, I don't
need to worry about it any more", and trying to get into "well actually
some people think some of the changes are okay and they can't agree which
ones but we should keep them anyway" undermines this expectation. If
someone wants to pipe up in support for sem over mic, I would be interested
in seeing their arguments on why we might want a completely waterproof
restoration, but otherwise my position is the synthesized set is worse than
the PR as pushed(tips hat to @cgyarvin), which in turn slightly errs on
pel/per but is itself livable.

ohAitch:
(And similarly going by <s>voting</s> nonbinding resolutions, which as the
lone dissent I unsurprisingly think we shouldn't, the OP got 5 hooray
reacts and no downvotes, which is both more emphatic and more numerous than
4 upvotes and 1 downvote)

0x70b1a5:
Re: taking ownership vs quibbling: Restoration is a mindset and accepts little compromise. The bard once said, "A drop of bikeshedding makes an entire discussion bikeshedding, but a drop of authority ends all debate."

msutherl:
As a complete beginner with some familiarity with the "K" style, I decided to try reading some of `hoon.hoon` with the "C" and "K" styles to see what I prefer. I also made some uninformed judgements based on certain proposed syllables that sound weird to me or evoke strong distracting mental imagery.

Some things I feel strongly about:

- Symbols with a left and right form that are often matched are much easier to say and remember when they lead with an "L" or "R". For this reason, I favor many of Curtis' suggestions.
- Symbols that have a left and right version, but are not often matched, I prefer with unmatched names. For instance "led" and "ban" seem weird and unintuitive at first, but they're actually more mnemonic than the "K" versions and using different vowels differentiates them from the matched pairs.
- A number of suggestions evoke similar words or distracting mental imagery for me. I'm aware that I'm especially sensitive to this, but it really bothers me. "mic" I read as "mike" or "microphone" impulsively. "Rob" and "lob" are good solutions to the problem at hand, but one is a name and the other is a word, and both have a kind of evocative mouth-feel that rubs me the wrong way. "vat" reminds me of vats, which are often gross. "Soq" sounds like sock, which is gross. "Doq" feels strange, and looks like "dog," but I guess I could get used to it.

Some things I don't feel strongly about:

- I suggested "lek" and "rek," which are perhaps easily confused with "lac" and "rac." I couldn't think of something better, and I see why "lob" and "rob" are good solutions technically. "Lob" and "rob" sound weird and incongruent when spoken in context. I spent some time with this line: `=+  a==>(v {p.gen})`, in which I must say "lit v lob p dot gen rob rit". Compare with "lit v rek p dot gen rek rit".
- "say" and "yel" are kind of funny and introduce a new mnemonic style, but I guess it's fine
- "buc" seems silly to me, as the only semantic mnemonic, but it's tempting to keep

"M" are my suggestions, with pros and cons of "M" suggestions to the right.

S | C | K | M | Pro (M) | Con (M)
-- | -- | -- | -- | -- | --
` ` | ace | ace | ace | 
`` \| `` | bar | bar | bar | 
`\` | bat | bas | bas | clear mnemonic | matched
`/` | net | fas | fas | clear mnemonic | "net" flows better, matched
`$` | bus | buc | buc | clear mnemonic | "buc" is silly
`_` | cab | cab | cab | 
`%` | cen | cen | cen | 
`:` | col | col | col | 
`,` | com | com | com | 
`.` | dot | dot | dot | 
` ` | gap | gap | gap | 
`#` | hax | hax | hax | 
`-` | hep | hep | hep | 
`^` | ket | ket | ket | 
`>` | ban | gar | ban | unmatched, clearer mnemonic | "gar" is cool
`<` | led | gal | led | unmatched, clearer mnemonic | none
`[` | lac | sel | lac | leading "L", clearer mnemonic | none
`]` | rac | ser | rac | leading "R", clearer mnemonic | none
`(` | lit | pal | lit | leading "L", fast | none
`)` | rit | par | rit | leading "R", fast | none
`{` | lob | kel | lek | leading "L", "lob" sounds bad | sounds like "lac"
`}` | rob | ker | rek | leading "R", "rob" sounds bad | sounds like "rac"
`+` | lus | lus | lus | 
`;` | mic | sem | sem | "mic" similar to "mike", sounds bad | not sure?
`&` | pad | pam | pam | clear mnemonic | not sure?
`'` | say | soq | say | "soq" is weird | not sure?
`~` | sig | sig | sig | 
`*` | tar | tar | tar | 
`` ` `` | tec | tec | tic | clear mnemonic | not sure?
`=` | tis | tis | tis | 
`@` | vat | pat | pat | "vat" is weird | not sure?
`?` | wut | wut | wut | 
`"` | yel | doq | yel | "doq" is weird | not sure?
`!` | zap | zap | zap |

ohAitch:
- con of sem: cen. This was the motivation for touching the syllabary in
the first place, and then everything got out of hand
- con of K parens: pel per sel ser kel ker convey a clear intuitive
category of paren-y-ness. (pal par, to reiterate, are neither C nor K)
- con of unmatched gal gar: they don't actually appear in runes that don't
have a matching version, iirc.
- con of say yel: are completely phonetically distinct, impossible to
connect without thinking about them as words instead of symbols. Meanwhile
soq/doq/tiq (quote, quote, almost quote) are doing a similar sort of "slant
rhyme" as pel/sel/kel/gal (paren, paren, paren, occasional paren)
- buc as the only semantic mneumonic: bar, dot, wut, zap. It's the only
*disputed* semantic mnemonic

Re mnemonic clarity, > "GreAteR", [ "cell". I'm willing to conced ban for
"bracket, angled" and lac for "left bracket" are not *less* clear
necessarily, but they're definitely not *more* clear.

pilfer-pandex:
Thank you all for your suggestions. @belisarius222 had something he wanted to contribute to this discussion. After that, I think it needs to be wrapped up.

belisarius222:
I remain a staunch Kerschnerian Restorationist. 

I actually tried using the Curtisian pronunciations for a couple months a while ago, and found the overall experience to be worse. Repeating a syllable over and over is not an actual use case, so it isn't a good test. How the syllables connect to one another and to English syllables is a much more important characteristic, and I think even that is less important than mnemonics. I also don't think speed of speech is as important as ease of speech, which might correlate but aren't the same thing.

Having spent months saying "bus" instead of "buc," I have concluded that "buc" is much better, and I will defend this position as required.

My major concern here is that the parentheses, brackets, curly braces, and quotes pretty much never get spoken out loud. To me, this is clear evidence that they need a redesign. The rest of the characters are pretty much fine as-is. I've tried some other alternatives, in addition to the Curtisian suite, and it's actually hard to devise a system for these that doesn't suck in some way.

Other than the runes that to me clearly need to be replaced, I think the weakest are hep and cab. Hep  is mostly used when speaking kebab-case variable names and Unix command-line arguments, and it doesn't feel or sound good in either of those contexts. Its mnemonic connection to "hyphen" is also quite weak, so people often forget it and mix it up with other things.

Cab is decent as a syllable, but I've never thought of a mnemonic for it, and it's always struck me as a bit clumsy for that reason.

Mic doesn't feel wonderful, but it's very important to distinguish it from sem, which was the worst problem with the syllables by far. It's also really tough to come up with something better than mic. My only suggestion would be 'mec', but I think mic is fine. I think we'll want to try out pronouncing it as 'meek,' 'meck', 'mick,' or 'mike' to taste, and I think they'll all be understandable. German has something like this with "zwei" and "zwo," and Mandarin has "nar" vs "nali." I don't think those cause major problems.

I do find gar and gal also get mixed up quite a bit, presumably because they sound so similar. Despite sharing a symmetry, no one mixes up =- and =+, but people mix up => and =< all the time. That being said, the utter lack of symmetry with led and ban also doesn't feel right. They're not usually matched up with each other, but they're often semantically symmetric: =< is a backward =>, and ?> is a negated ?<. Of the two, led is better than ban, but I don't think they really work that well. Maybe 'led' and 'gar', with mnemonics 'less than' and 'greater than', would work.

Despite my concerns with these syllables, they get used and they're not horrible. Long-term I think we'll want to replace (), [], {}, ', and ", but the rest should probably stay as-is.

ohAitch:
My experience is that
- parens are too common to usefully pronounce(every wide rune), and are
generally implied by the opening rune or gate name, and a trailing
pause(+usually-fixed arity), respectively
- curly braces are rare(assuming Kerscherian type syntax), but I don't
think I've heard anyone circumlocute with "interpolate" or anything
- "sel a b", "sel a b c" are actually spoken that way iirc; the closing ser
is also usually a pause
- along similar lines, I think say and yel are okay opening quotes, but
terrible closing quotes? "soq-a-soq" feels symmetric esp with vowel
adjustments, but when you say "say 'whatever'" in English this is strongly
established syntax for a right-binding operator. So "wutcol a say b say c"
seems likely to first parse as a lispy ?:(a 'b 'c) and not ?:(a 'b' c).
(Similar for yel)

ohAitch:
Wait I take it back about them okay opening quotes, as the owner of ~fyr I
would like to request a formal resolution on whether "y" is a vowel or a
consonant. (In this PR I've definitely been pronouncing it as short i, and
having everything be CVC seems important for establishing "is this person
trying to say a glyph" probabilities)

cgyarvin:
Pronouncing "fyr" as "fire", not "fir", keeps it away from the "schwa trap"
in which all vowels sound the same.

ohAitch:
Sorry, shorter i, "saj" "jel" might convey the pronunciation of y as a
consonant better. Agree that ~fyr is fire :D

The vowel concern is not theoretical, (yell '"sail" is the hoon templating
system') is a bit of a nightmare under the revisionist quotes.

rmariani:
My only input here is that `buc` is, outside of taste and opinions, superior to `bus`.

0x70b1a5:
(I thought "fyr" was "fear".)

ohAitch:
I'll take it! But I think the official urbit spelling there would have been
"fier"

pilfer-pandex:
Based on extensive discussion, I have decided on the following "conservative modified restorationist" slate:

| Symbol | Pronunciation  |
| ---- | ---- |
| `(` | pal |
| `)` | par |
| `[` | sel |
| `]` | ser |
| `{` | kel |
| `}` | ker |
| `<` | gal |
| `>` | gar |
| `$` | buc |
| `;` | mic |
| `/` | fas |
| `\` | bas |
| `&` | pam |
| `@` | pat |
| `'` | soq |
| `"` | doq |
| `\``| tic |

Thank you all for your input. No decision here will make everyone happy, but I expect that this is one we can all be okay with. At the very least, we can all agree that it is more okay than further delay, since we have students in Hoon School who must learn a pronunciation. @baudtack will be appearing here shortly with a commit implementing this change, which I will accept.

wcable:
(I'm not sure a commit is necessary, afaict this is the original PR verbatim)

ohAitch:
(I'm not sure a commit is necessary, afaict this is the original PR verbatim)

On 3/13/19, Anton Dyudin <notifications@github.com> wrote:
> (I'm not sure a commit is necessary, afaict this is the original PR
> verbatim)
>
> --
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub:
> https://github.com/urbit/docs/pull/399#issuecomment-472671140


-- 
Here lies a toppled god —
His fall was not a small one.
We did but build his pedestal,
A narrow and tall one.

pilfer-pandex:
No, it also changes tec to tic.

ohAitch:
Ah yep, you're right! Feel kinda bad about having that in a "restore" PR
since it didn't exist in either version, otoh neither did the parens; we
agreed on a result and that's the important thing.

PR #400, opened 2019-02-21
updates getting-started.udon

rmariani:
This improves the document. Gets rid of some weird stuff where we assume prior knowledge of the wallet generator. Also boils down the "Bridge" section and makes the setting of the networking seed clearer.


msutherl:
@rmariani I want to change what we say about Landscape here, but also eager to merge updates to Getting Started

PR #406, opened 2019-02-27
remove title text from image markdown in learn/arvo/arvo-internals/ha…

jfranklin9000:
…ll.udon (temporary?)

The udon parser doesn't handle title text markdown in links or images.
(It doesn't handle images at all but see https://github.com/urbit/arvo/pull/1085.)

So this patch gets rid of the title text.  If/when udon handles title
text these can be re-added.

This patch assumes the udon image patch will land.  If it doesn't
the image markdown should be changed to link markdown.


rmariani:
Thanks so much for the parser update. I'm sure it will be integrated promptly, but I will close this PR for now and make another one that fixes it under the current Sail-only image regime.

rmariani:
https://github.com/urbit/arvo/pull/1085 is merged, so this looks good

PR #413, opened 2019-03-02
adds list.udon to workbook

rmariani:
This adds a Workbook problem that walks total noobs through the basics of Hoon via an example. Some details are glossed over.


rmariani:
Going to make a few edits actually

PR #414, opened 2019-03-03
fixed standard library sidebar text for section 5d

jfranklin9000:


rmariani:
https://github.com/urbit/docs/pull/416#issuecomment-469356911

PR #415, opened 2019-03-03
moved some arm documentation from 4k to 4i to match hoon.hoon

jfranklin9000:
This also contains the #414 changes.

PR #416, opened 2019-03-04
added a page of standard library arms, listed alphabetically

jfranklin9000:
Work in progress.  It will be complete in a day or two.

This also contains the #414 and #415 changes.


rmariani:
We're doing a restructuring of this (and other) sections of the docks, so I am going to put this on ice for now. The restructuring should be merged soon.

rmariani:
Reopening because the restructuring that I mentioning is going to be further down the line

jfranklin9000:
I've finished adding all the arms.  I can get rid of most of the tecs if https://github.com/urbit/arvo/pull/1091 lands, and all of them if https://github.com/urbit/arvo/pull/1096 lands.

rmariani:
I think the links should be in the scheme of `(../library/4j/)` rather than `(/library/4j/)`

jfranklin9000:
OK, I'll take a look.  I guess it would be a good time to actually build all the docs locally and test..

jfranklin9000:
`alphabetical-arms.udon` is in the same directory as `library/`, not in `library/`, so it should actually be `(./library/4j/)` (or `../reference/library/4j/`).  I've changed it to `(./library/4j/)`.  Good catch.

If https://github.com/urbit/arvo/pull/1096 lands I'll remove the tecs.

rmariani:
Looks good. Thanks for your hard work!

PR #421, opened 2019-03-06
Rob edits

rmariani:
Adds Bridge info to azimuth.udon, which is the destination of a URL printed on paper wallets (via redirect)

Also adds an explanation of the address-space hierarchy in a new document.


rmariani:
@baudtack this is high-ish priority due to us wanting the URL to work

cgyarvin:
Don’t forget urbit.org/docs/pki

rmariani:
Yeah I redirected in there because that explanation already seemed sufficient for me

I think I will add a link to the Azimuth repo though...

PR #423, opened 2019-03-06
Update getting-started.udon

majcosta:
makes sure pacman is using latest repos
added --needed to pacman so existing packages aren't reinstalled
installed pip via pacman instead of curl which is tidier
added meson to pacman install line
separated building tests from final executable install to minimize superuser usage and not leave root-owned files lying around


rmariani:
Thank you

PR #430, opened 2019-03-12
Updated docs for booting a moon.

jfranklin9000:
I tested `-G` and `-k` flags, and both ways of creating `<keyfile>`.


galenwp:
And you were able to successfully boot a moon? Been meaning to test this.

jfranklin9000:
Yes, with both `-k` and `-G` options, and both ways of creating a keyfile for the `-k` option.

galenwp:
Nice work. Thank you!

PR #433, opened 2019-03-13
clarifies bridge instructions

rmariani:
This will make setting networking keys a bit less confusing


ohAitch:
.urb/chk?

rmariani:
This is what I get for trying to use vim like a real hacker

PR #459, opened 2019-04-08
faswork -> network

ryjm:
overzealous search and replace maybe?


baudtack:
That was indeed what happened. Thanks for the fix.

PR #463, opened 2019-04-10
Linebreak

rmariani:
Fixes a rogue newline


msutherl:
LGTM. @rmariani you should get rid of all the double-spaces in `learn/hoon/hoon-tutorial/hoon-syntax.udon` as well.

PR #466, opened 2019-04-17
rearranges getting-started.udon so it's easier to follow

rmariani:
People reported that they found the getting-started page hard to follow. This makes it more straightforwardly a map to our getting-started guides.

### DIFF
--- a/getting-started.udon
+++ b/getting-started.udon
@@ -7,7 +7,7 @@
 
 # Getting Started
 
-Welcome to Urbit, a personal server that makes you the owner of your digital life. Urbit isn't ready for serious use yet, but it definitely works. Follow this step-by-step guide to get an Azimuth point, install Urbit, and join the community.
+Welcome to Urbit, a personal server that makes you the owner of your digital life. Urbit isn't ready for serious use yet, but it's available to try out. Follow this step-by-step guide to get an Azimuth point, install Urbit, and join the community.
 
 ### [Step 1: Using Bridge](using-bridge)
 
@@ -31,7 +31,7 @@ You've already encountered some new terminology on this page, and you'll encount
 
 - Arvo is the name of the Urbit operating system.
 - An instance of the Arvo operating system is called a _ship_.
-- The network linking ships on this network is called "the Arvo network."
-- Azimuth is the name of the Urbit identity layer.
-- If you want to get on the Arvo network, an Azimuth-based identity called a _point_.
-- Urbit is the project as the whole, which includes all of these elements and more.
+- Ships communicate with one another over "the Arvo network."
+- Azimuth is an Ethereum-based PKI used to give each ship a unique identity on the Arvo network.
+- An Azimuth identity is called a _point_. You'll need a point to use the Arvo network.
+- Urbit refers to the project as a whole, including Arvo, Azimuth, and other system components.
--- a/getting-started.udon
+++ b/getting-started.udon
@@ -7,36 +7,31 @@
 
 # Getting Started
 
-Pages in this section are guides to getting you started with Urbit.
+Pages in this section are step-by-step instructions to getting you started with Urbit. Follow the guides below in the order that they are listed.
 
-But first, the absolute basics. Arvo is the name of the Urbit operating system and the name of network. Azimuth is the name of the Urbit identity layer. Urbit is the project as the whole.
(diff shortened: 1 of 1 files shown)


rmariani:
@msutherl curious about your thoughts

msutherl:
@rmariani left some comments. Seems like an iteration in a positive direction overall.

rmariani:
Alright, too your suggestions into account @msutherl

getting-started.udon
```diff
@@ -7,36 +7,31 @@
 
 # Getting Started
 
-Pages in this section are guides to getting you started with Urbit.
+Welcome to Urbit, a personal server that makes you the owner of your digital life. Urbit isn't ready for serious use yet, but it definitely works. Follow this step-by-step guide to get an Azimuth point, install Urbit, and join the community.
```
msutherl: This feels a little off to me still.

I wonder if the first sentence should stay vague. "Makes you the owner of your digital life" sets up the expectation that there's actually software in existence to do that. Right now Urbit doesn't actually provide that experience. How about just "Welcome to Urbit, a personal server."

"Urbit isn't ready for serious use yet, but it definitely works" also worries me – I'm not sure if we can say "definitely works" and mean it. "Urbit isn't ready for serious use yet, but if you're curious you can try it out. Follow this step-by-step guide to get an Azimuth point, install Urbit, and talk to other users on the network."

getting-started.udon
```diff
@@ -7,36 +7,31 @@
...
-### [Booting a Ship](booting-a-ship)
-
-This guide will walk you through the process of getting your ship operational and on the Arvo network.
-
-### [Creating a Development Ship](creating-a-development-ship)
-
-If you just want to do development work with Arvo, and don't need to be on the Urbit network, this section is for you.
+- Arvo is the name of the Urbit operating system.
+- An instance of the Arvo operating system is called a _ship_.
+- The network linking ships on this network is called "the Arvo network."
```
msutherl: This is tautological :) Maybe "Ships can talk to each other over 'the Arvo network.'"

getting-started.udon
```diff
@@ -7,36 +7,31 @@
...
+- Azimuth is the name of the Urbit identity layer.
```
msutherl: Maybe "Azimuth is an Ethereum-based PKI used to give each ship a unique identity on the Arvo network"

getting-started.udon
rmariani: How about "but it works as alpha-stage software."

getting-started.udon
```diff
@@ -7,36 +7,31 @@
...
+- If you want to get on the Arvo network, an Azimuth-based identity called a _point_.
```
msutherl: "An Azimuth identity is called a 'point.' You'll need a point to use the Arvo network.

getting-started.udon
```diff
@@ -7,36 +7,31 @@
...
+- Urbit is the project as the whole, which includes all of these elements and more.
```
msutherl: Typo. Maybe "'Urbit' refers to both the `urbit` program that runs Arvo and Arvo's interpreter, Vere, and the entire project including both Arvo and Azimuth."

getting-started.udon
msutherl: We are [pre-alpha](https://en.wikipedia.org/wiki/Software_release_life_cycle#Pre-alpha), but I wouldn't use this terminology at all.

getting-started.udon
rmariani: This seems like an inordinate info-dump

getting-started.udon
rmariani: I think this is a matter of debate, and "alpha" properly carries the connotation of "rudimentary"

getting-started.udon
msutherl: "Urbit refers to the project as a whole including Arvo, Azimuth, and other system components."

msutherl: Mainly "elements" is vague and "and more!" I'd avoid

getting-started.udon
rmariani: Alright, I am just going to drop the terminology

msutherl:
@rmariani we'll have to create a new PR for this after `zola-port` is merged

PR #470, opened 2019-04-22
better exposes azimuth concept sub-pages

rmariani:


concepts/azimuth.udon
```diff
@@ -22,3 +22,15 @@ Urbit is a project, not a single computer system. It has three components: Arvo,
...
+
+## Further Reading
+
+If you want to know more about Azimuth, check out the pages below.
+
+### [Why Urbit Uses a Blockchain](/docs/concepts/azimuth/blockchain)
+
+An explanation for why Ethereum makes sense for Azimuth.
+
+### [The Urbit HD Wallet](/docs/concepts/azimuth/blockchain)
```
Fang-: These are two identical links. Probably unintended.

rmariani: Nice catch! fixed

PR #471, opened 2019-04-26
State that the keyfile must not have a trailing newline in Booting a …

jfranklin9000:
…Ship section.


Fang-:
imo, seeing as how urbit/urbit#1166 should be a simple fix, I'd rather just kick someone into making that fix, rather than adding temporary requirements to the docs.

ohAitch:
In particular, if the keyfile is plain text, conforming to the UNIX
standard "sequence of lines terminated by a newline" seems good - I think
at some point there was even discussion of enforcing that in clay, most
text editors certainly will.

rmariani:
Let's wait for the fix.

PR #472, opened 2019-04-26
Fix weird word error

Fang-:
idk what kind of find-and-replace was run to mess this up (^:

Critical readers will complain that "but you can also do computation on Bitcoin!", but that's a topic for a different PR.


rmariani:
Nice catch. LGTM

Issue #474, opened 2019-04-26
Urbit HD wallet page doesn't explain networking keys


Fang-:
It talks about ownership and proxies. But the diagram shown includes networking keys, which aren't talked about at all.

We should add a brief paragraph explaining that the Urbit HD wallet structure makes it easy to rotate public keys (for use with Arvo) also.

PR #475, opened 2019-04-30
adds information about networking keys

rmariani:
Addresses https://github.com/urbit/docs/issues/474

### DIFF
--- a/concepts/azimuth/wallet.udon
+++ b/concepts/azimuth/wallet.udon
@@ -21,7 +21,7 @@ are also derived from the wallet.
 Urbit HD wallets are composed of the following items, which are each assigned to
 their own individual Ethereum key-pairs.
 
-#### Master Ticket
+## Master Ticket
 
 Think of your master ticket like a very high-value password. The master ticket
 is the secret code from which all of your other keys are derived. Technically,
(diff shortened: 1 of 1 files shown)


concepts/azimuth/wallet.udon
```diff
@@ -21,40 +21,56 @@ are also derived from the wallet.
...
 child points, voting, and setting networking keys without jeopardizing the keys
 you've designated with ownership rights. Setting proxy rights is optional, but
 it is recommended for on-chain actions you will execute more frequently.
 
-- *Management Proxy*
+### Management Proxy
 
-  Can configure or set Arvo networking keys and conduct sponsorship related
-  operations.
+Can configure or set Arvo networking keys and conduct sponsorship-related
```
Fang-: "configure or set"? Either is fine.

concepts/azimuth/wallet.udon
```diff
@@ -21,40 +21,56 @@ are also derived from the wallet.
...
+operations.
 
-- *Voting Proxy*
+#### Networking Keys
 
-  Galaxies only. Galaxies are the part of the galactic senate, and this means
-  they can cast votes on new proposals including changes to the Ecliptic.
+Networking keys -- encryption and authentication key-pairs -- are used
+for actually interacting on the Arvo network. They are derived from the keyfile
+that you use to boot a ship.
```
Fang-: Github doesn't support formal multi-line suggestions, so have this instead:

Networking keys -- a pair of encryption and authentication keys -- are used for communicating on the Arvo network. They are contained in the keyfile that you use to boot a ship.

concepts/azimuth/wallet.udon
```diff
@@ -21,40 +21,56 @@ are also derived from the wallet.
...
 
-- *Spawn Proxy*
+Unlike proxies, networking keys are not Ethereum wallets. They are just a public
+key that is the point's state on the blockchain, combined with a private key
+that is stored somewhere on your machine.
+
+The HD Wallet structure makes networking keys easy to change if the need arises.
+If, say, your keys are compromised, you only need to access your management proxy
```
Fang-:
```suggestion
If you want or need to rotate your keys, you only need to access your management proxy
```

concepts/azimuth/wallet.udon
```diff
@@ -21,40 +21,56 @@ are also derived from the wallet.
...
+to reset them. Once you do that, download a new keyfile and boot your ship with
+it.
```
Fang-: Suggestion: After you've done that, download a new keyfile and add it to your ship.

Or something along those lines. But you really only care about the *contents* of the file, which Bridge also displays to you in addition to the download button. You need the contents, as a string, for use with [`|rekey`](https://github.com/urbit/arvo/blob/master/gen/hood/rekey.hoon). ie `|rekey '0wmy.cool.key'`.

concepts/azimuth/wallet.udon
```diff
@@ -21,40 +21,56 @@ are also derived from the wallet.
...
+
+### Voting Proxy
```
Fang-: Unrelated, but I'd probably put this *after* the spawn proxy, because it's even less likely to be relevant for the reader.

rmariani:
done.

Fang-:
Can I see the changes, please? ie can you push them? (:

rmariani:
Whoops, thought I did

concepts/azimuth/wallet.udon
```diff
@@ -21,40 +21,55 @@ are also derived from the wallet.
...
-  Can configure or set Arvo networking keys and conduct sponsorship related
-  operations.
+Can set Arvo networking keys and conduct sponsorship-related operations.
 
-- *Voting Proxy*
+#### Networking Keys
 
-  Galaxies only. Galaxies are the part of the galactic senate, and this means
-  they can cast votes on new proposals including changes to the Ecliptic.
+Networking keys -- encryption and authentication key-pairs -- are used
```
Fang-: I believe "key pairs" would be more correct.

rmariani: Not sure I agree here. It's used elsewhere as a compound noun (keypair), and the hyphen similarly reduces ambiguity. "Pair of keys" is the indivisible concept here. Compare "old-key pair" to "old key-pair." If we write it as "old key pair," it's unclear which one of these ideas is the one we want to convey.

Fang-: Eh, fair enough. I'd probably go with "keypairs" then for consistency, but do what you think is best here.

msutherl: Surprised to see "key pair" is possibly more common, but "keypair" is also accepted https://en.wiktionary.org/wiki/keypair

PR #479, opened 2019-05-09
don't bold(udon)/italicize(md) `|reset` or `|start` in Hood section o…

jfranklin9000:
…f using/shell.md

I don't know why we would bold or italic here so I deleted it.


msutherl:
LGTM

PR #480, opened 2019-05-09
convert sail syntax to markdown/html in md files

jfranklin9000:
I used an empty `<span>` placed on the line above the "target" to embed an `id` to be linked to.  This is to separate the markdown from the html required for `id`s.

I replaced `<a>` tags of the form `<a id="some-id"/>` (`<a>` tags are not self-closing) with `<span>`s.

There were two places with `id="ixor"` in `learn/arvo/eyre.md` so I renamed the first one `view-ixor` but I can't tell if that's the right thing to do.

I deleted `(class "first child")` when converting `using/shell.md`.  I don't know if that's what we want.

### DIFF
--- a/learn/arvo/eyre.md
+++ b/learn/arvo/eyre.md
@@ -47,7 +47,8 @@ Runtime
 Let us follow the loading of a simple cli app, as it bounces from
 browser to server to browser and back.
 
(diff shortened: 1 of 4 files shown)


rmariani:
Thanks!

PR #484, opened 2019-05-10
use %28 and %29 for parens within markdown link urls

jfranklin9000:
fixes https://github.com/urbit/docs/issues/481


rmariani:
handled by another PR

PR #486, opened 2019-05-13
Update Bridge instructions to match latest release

Fang-:
As seen on [bridge/releases](https://github.com/urbit/bridge/releases).

Feels kinda hazardous to have instructions in two places for this very reason. But, they're unlikely to change again for a while, so eh.


rmariani:
righteous

PR #469, opened 2019-04-19
Update star operator instructions

msutherl:

PR #487, opened 2019-05-14
Update star operator instructions

msutherl:
Supersedes https://github.com/urbit/docs/pull/469

### DIFF
--- a/getting-started/operating-a-star.md
+++ b/getting-started/operating-a-star.md
@@ -27,17 +27,16 @@ Below are some handy operations for the use and maintenance of your star.
 
 ### DNS Proxying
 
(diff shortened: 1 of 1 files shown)


joemfb:
@msutherl, this doesn't appear to be correct -- `:dns|ip` is not obsolete. Do you know where this information came from?

I still owe you and the docs team a review-of / update-to this document, and maybe an additional document for galaxies. The biggest missing piece appears to be port management.

msutherl:
@joemfb I can't recall who told me that, but someone did – and I recall when I last set up a star, I didn't need to run `:dns|ip`. What do you think these docs should say?

PR #493, opened 2019-05-17
workbook update

baudtack:


msutherl:
Reverted. This broke the build.

rmariani:
It must be because of the presence of a .udon file.

@baudtack, can you make recursion.udon into recursion.md?

learn/hoon/hoon-tutorial/recursion.udon
```diff
@@ -0,0 +1,200 @@
+:-  :~  navhome/'/docs/'
```
msutherl: @baudtack this file needs TOML metadata – we don't use sail anymore. See other documents in the repo for a template, and note the "weight" key, which specifies the sort order in the sidebar

PR #496, opened 2019-05-20
Update Bridge action description

Fang-:
As per urbit/bridge#141.


msutherl:
LGTM

msutherl:
@Fang- I'll merge this right away since it's not any more wrong (before https://github.com/urbit/bridge/pull/141 is merged and released) than what's there currently

PR #504, opened 2019-05-22
two examples and a bash script

baudtack:


msutherl:
@baudtack can you

1. Add comments to the weights.sh script
2. Drop it in a `scripts/` folder (I'd prefer in the urbit.org repo, but whichever you prefer)

msutherl:
Hm, actually I'm worried that creating a folder will add a section to the website build, so maybe test first.

PR #507, opened 2019-05-23
Remove list headers, delete some spurious span tags

msutherl:
Headers in lists are no longer supported by our styles. This PR changes a couple of tables of contents to standard lists.

Also includes deletion of a bunch of random <span> tags I came across and remove colons at the end of some headers in the Glossary.

### DIFF
--- a/reference/glossary.md
+++ b/reference/glossary.md
@@ -9,12 +9,10 @@ from the strange words within.
 As Dijkstra put it: "The purpose of abstraction is not to be vague, but
 to create a new semantic level in which one can be absolutely precise."
 
-<span id="application"></span>
 ### application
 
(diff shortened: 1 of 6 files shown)


reference/glossary.md
```diff
@@ -25,14 +23,14 @@ enclosing [core](#core) itself as the subject.
 There are two kinds of arms: _dry_ and _wet_. Most arms are
 dry.
 
-- ##### dry:
+##### dry
```
rmariani: Don't we want to nix the h5?

msutherl: I didn't take the liberty to change the header depth in this document. Would prefer it be merged ASAP and we can fix that in a separate batch.

PR #508, opened 2019-05-23
changes

baudtack:


msutherl:
@baudtack @rmariani this PR broke the build, reverted in https://github.com/urbit/urbit.org/commit/e63a0ce192f3ed17dbddbfebcbfad9012f80fc7c

baudtack:
CI needs to be set up on this.

msutherl:
Should be possible with the current configuration to make a branch in urbit.org that references this branch and runs CI.

Alternatively, you can just check out urbit.org@master locally, point the submodule at this commit, and run `zola serve`.

baudtack:
If making commits is going to automatically deploy and break the site, we absolutely must have CI setup and CI needs to block on merges into urbit.org@master.

msutherl:
Only pushing to urbit.org@master automatically deploys, so the gating action is a PR in the urbit.org repo. That said, if you merge breaking changes into docs@master, somebody might accidentally deploy them.

I do think the best solution here is to merge the repos, but there are laborious ways around it.

> CI needs to be set up on this.

I imagine it's possible to set up CI that takes into account submodules, but we're using Netlify's built-in CI for lack of available time to configure Travis, so c'est la vie.

Fang-:
> I imagine it's possible to set up CI that takes into account submodules, but we're using Netlify's built-in CI for lack of available time to configure Travis, so c'est la vie.

Just chiming in to share my thought, in line with @baudtack's, that merging the repos is probably too drastic and long-term-impactful a change to justify like this.

PR #514, opened 2019-05-30
Add Landscape subscription instructions to Messaging

msutherl:


using/messaging.md
```diff
@@ -65,6 +67,18 @@ You'll see your message printed below messages from others that came before it:
...
 ```
 
+### Subscribing to a Channel in Landscape
+
+In Landscape, you can always navigate to a channel by opening the menu, or using the shortcut `cmd-k`, then typing `go ~dopzod/urbit-help`. However, if you want to see messages in your inbox, you'll need to subscribe.
+
+Let's try. From the Talk terminal, run:
+
+```
+;source %inbox ~dopzod/urbit-help
```
Fang-: `;join` is shorthand for `;source %inbox`, and also sets presence, which this does not.

Fang-:
At some point we should just separate Landscape and CLI docs.

using/messaging.md
msutherl: Oh, cool – it looks like `;join` subscribes you in Landscape now. IIRC, that wasn't the case before, and when I asked @loganallenc about it this morning, he said you can follow up `;join` with `;source %inbox` to subscribe in Landscape. I thought I'd also tested it to confirm, but looks like I was wrong – `;join` does indeed subscribe you in Landscape. I'll change this.

msutherl:
> At some point we should just separate Landscape and CLI docs.

Landscape will become Modulo chat soon – not sure what they're planning on for documentation. This was meant as a temporary hack to prevent some confusion.

using/messaging.md
Fang-: Honestly the real fix is to have a "join" button in Landscape that just sends a `%source` action to the local hall, which is all that `;join` [does](https://github.com/urbit/arvo/blob/master/app/talk.hoon#L1199) anyway. This used to be in at some point, not sure why it was removed.

_cries in hall_

msutherl: @Fang- I want that feature back as well – it was removed for _some_ reason

using/messaging.md
```diff
@@ -27,14 +28,14 @@ take a look at its [documentation](./docs/learn/arvo/hall.md).
 
 ## Quickstart
 
-The most common use of talk right now is as a single-instance of Slack: one main
-channel (`~dopzod/urbit-help`) and direct messages. Everyone is more than welcome
+The most common use of talk right now is as a single IRC or Slack channel (`~dopzod/urbit-help`) and direct messages. Everyone is more than welcome
```
rmariani: I think better language would be

```
The most common uses of Talk right now are communicating over a public chat channel called `~dopzod/urbit-help` and sending direct messages.
```

msutherl:
@rmariani just a heads-up, I'll return to this in a day or two

PR #518, opened 2019-06-08
Clarify context for accepting transfer

Fang-:
To try and combat confusion found in urbit/bridge#145.


shrugs:
LGTM. would like to include nudges like that in the UI somewhere as well

msutherl:
@Fang- ditto @shrugs

PR #519, opened 2019-06-13
Update libsigsegv-2.11-7.fc30.x86_64 URL locations

matildepark:
In the AWS installation instructions, the two `libsigsegv-2.11-7.fc30.x86_64` URLs no longer work — I've updated them with their current locations. The package version is newer, but urbit still compiles, so I am assuming no breaking errors.

## For future reference

It seems like Fedora will just change the URLs as the packages update, so you may want to establish a local (urbit.org) location for the packages to prevent future breaking links.

### DIFF
--- a/getting-started/installing-urbit.md
+++ b/getting-started/installing-urbit.md
@@ -168,10 +168,10 @@ sudo env "PATH=$PATH" pip3 install meson
 # we need libsigsegv
 #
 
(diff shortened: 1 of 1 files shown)


rmariani:
Thanks Matilde!

matildepark:
Good to hear @jtobin, I just try to prevent other people from hitting the same roadblocks when I encounter them. Might make some other PRs in a similar vein later on.

PR #523, opened 2019-06-18
Fixing broken links in the glossary

matildepark:
Various links give a 404 on visit, so I amended them to their current locations.


rmariani:
Excellent

PR #524, opened 2019-06-18
1.1.1: Removing line numbers so that the example runs

matildepark:
In section 1.1.1, if you do as the tutorial says, and save the code and run it *as it's provided*, you'll receive this error:

```
ford: %hood: syntax error at [8 63] in /~zod/home/0/gen/list/hoon
ford: %core on /~zod/home/0/gen/list/hoon failed:
```

Which is less than ideal for a newcomer interactively engaging with Hoon. Removing the line number comments fixes this. On a cursory skim, I don't see line numbers on any other example besides this first one, so it seems more consistent, too.


joemfb:
I think this is caused by urbit/arvo#307. See also urbit/arvo#968.

ohAitch:
This sounds like a hoon/ford parser bug? Like, the example *should* compile
with the comments.

matildepark:
Not sure how I missed the reference to the line numbers and comments *directly below the example*. OK, I see — would be better to address the parser bug.

PR #525, opened 2019-06-18
1.1: Removing redundant information, clarifying flow

matildepark:
Perhaps a more controversial PR? The first part of the tutorial series repeats information found elsewhere in the documentation — and it even links to those parts at the start. It would be much clearer to just continue from those posts as a starting point. 

In addition the usage of 'ship' and 'comet' is inconsistent (and refers to a planet as a 'comet,' potentially confusing a new user after instructing that a comet has a long address), so I've made it consistently a *ship*.

Finally the last part talks as if the 'noun' section is immediately following, but then goes into generators, which is jarring for the onboarding flow. I've clarified that what's coming *first* is a simple program, and afterward we will discuss what nouns are.

Edit: PR in sum

- Eliminates comet setup information from the start of the Hoon tutorial
- Makes the wording consistently 'ship' in that section
- Changes flow into the generators paragraph by amending the "Noun" paragraph in 1.1 to better clarify the onboarding flow.
- Mentions the three paths to creating an urbit on Getting Started
- Migrates the comet information to "Using" section, beside the fakezod setup information
- Keeps the Hoon tutorial clean and recommend users use a fakezod to develop.

### DIFF
--- a/getting-started/_index.md
+++ b/getting-started/_index.md
@@ -5,13 +5,13 @@ sort_by = "weight"
 template = "sections/docs/chapters.html"
 +++
 
-Welcome to Urbit, a personal server. Urbit isn't ready for serious use yet, but it's available to try out. If you have an Azimuth point, you can follow this step-by-step guide to get an Azimuth point, install Urbit, and join the community.
+Welcome to Urbit, a personal server. Urbit isn't ready for serious use yet, but it's available to try out. If you have an Azimuth point, you can follow this step-by-step guide to get your point's _keyfile_, which you need to boot your ship. Then you can install Urbit and join the community.
 
 If you don't have an Azimuth point, you can always [create a comet](./docs/using/creating-a-comet.md) to join the network before committing to a point, or [create a development ship](./docs/using/creating-a-development-ship.md) and start using Urbit offline first.
 
 ### [Step 1: Using Bridge](./docs/getting-started/using-bridge.md)
 
-In order to join the Urbit network, you need an Azimuth point, which acts like an identity and a password for your Urbit ship. Bridge is our client for managing Azimuth points. This guide tells you how to use Bridge to get your point's _keyfile_, which you need to boot your ship.
+In order to join the Urbit network, you need an Azimuth point, which acts like an identity and a password for your Urbit ship. Bridge is our client for managing Azimuth points, and lets you get your keyfile and claim the point.
 
 ### [Step 2: Installing Urbit](./docs/getting-started/installing-urbit.md)
 
--- a/getting-started/_index.md
+++ b/getting-started/_index.md
@@ -5,7 +5,9 @@ sort_by = "weight"
 template = "sections/docs/chapters.html"
 +++
 
-Welcome to Urbit, a personal server. Urbit isn't ready for serious use yet, but it's available to try out. Follow this step-by-step guide to get an Azimuth point, install Urbit, and join the community.
+Welcome to Urbit, a personal server. Urbit isn't ready for serious use yet, but it's available to try out. If you have an Azimuth point, you can follow this step-by-step guide to get an Azimuth point, install Urbit, and join the community.
+
+If you don't have an Azimuth point, you can always [create a comet](./docs/using/creating-a-comet.md) to join the network before committing to a point, or [create a development ship](./docs/using/creating-a-development-ship.md) and start using Urbit offline first.
 
 ### [Step 1: Using Bridge](./docs/getting-started/using-bridge.md)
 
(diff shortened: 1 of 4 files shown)


learn/hoon/hoon-tutorial/setup.md
```diff
@@ -17,35 +17,9 @@ An **urbit** is an Urbit virtual computer with persistent state that can connect
...
-http: live (insecure, loopback) on 12321
-~palnul_nocser:dojo>
-```
-
-(There will likely be other messages as well.)
-
-Look at the prompt, `~palnul_nocser:dojo>`.  To the left of the `:` you'll see an abbreviation of the comet name you just booted, `~palnul_nocser` in this case.  To the right is the name of a local application, `dojo`.  **Dojo** is the Urbit command line app; it's also a Hoon [REPL](https://en.wikipedia.org/wiki/Read%E2%80%93eval%E2%80%93print_loop) we'll use to run simple Hoon examples.
-
-To make sure everything is working, type `(add 2 2)` at the prompt and hit return.  Your screen now shows:
+Once you've installed your development ship (or your own ship on the live network), let's try a basic command. Type `(add 2 2)` at the prompt and hit return.  Your screen now shows:
```
msutherl: Perhaps "booted" or "started" rather than "installed"? You might also link to instructions for booting a fakezod, or provide them inline.

matildepark: Amended.

baudtack:
Why exactly is all the information on comets being cut?

matildepark:
@baudtack It predominantly restates information from the other two installation streams ("I have a point and I'm booting for the first time from Bridge" and "I'm creating a fakezod"). 

But, looking through the documentation I can't see any other place we mention comets, now. It might ideally live on [creating a development ship](https://urbit.org/docs/using/creating-a-development-ship/)? I can migrate and reorganise the information there? I don't see why it should live in the Hoon tutorial.

msutherl:
@baudtack makes sense to me – learning Hoon with a comet doesn't seem like a great idea, and it adds a big chunk of text plus ~10 minutes of waiting to the onboarding time.

If you mean 'why are we removing all mention of comets anywhere,' I agree with @matildepark that this information can go elsewhere.

baudtack:
It might not belong here but the intro work flow for urbit must mention comets. A comet is not a development ship and so it wouldn't fit in the development ship section. The comet workflow is a third  branch into a running urbit. It doesn't require a point, but unlike a fakezod it will still connect to the live network.

matildepark:
True, true — it's a valid point in itself.

baudtack:
If this is going to be changed at all, we should **strongly** discourage people from doing hoon development on their live planets as they are almost certainly going to sink them.

baudtack:
Perfect. I was just going to suggest putting it there @matildepark

matildepark:
One more commit, just a second...

matildepark:
Okay, so, the last few commits in sum:

- Mention the three paths to creating an urbit on Getting Started
- Migrate the comet information to "Using", beside the fakezod information
- Keep the Hoon tutorial clean and recommend users use a fakezod to develop.

getting-started/_index.md
```diff
@@ -5,7 +5,9 @@ sort_by = "weight"
 template = "sections/docs/chapters.html"
 +++
 
-Welcome to Urbit, a personal server. Urbit isn't ready for serious use yet, but it's available to try out. Follow this step-by-step guide to get an Azimuth point, install Urbit, and join the community.
+Welcome to Urbit, a personal server. Urbit isn't ready for serious use yet, but it's available to try out. If you have an Azimuth point, you can follow this step-by-step guide to get an Azimuth point, install Urbit, and join the community.
```
Fang-: > If you have an Azimuth point, you can follow this step-by-step guide to get an Azimuth point

learn/hoon/hoon-tutorial/setup.md
```diff
@@ -66,36 +40,15 @@ You just used a function from the Hoon standard library, `add`.  Next, quit Urbi
...
-
-```
-$ urbit mycomet
-[...]
-ames: on localhost, UDP 31337.
-http: live (insecure, public) on 8080
-http: live ("secure", public) on 8443
-http: live (insecure, loopback) on 12321
-~palnul_nocser:dojo>
+urbit palnul_nocser
```
Fang-: `palnul_nocser` is not a valid ship name (it's just a common way to abbreviate comet names), and comets always have user-defined names for their piers/directories, so this might be confusing.

If we're proceeding off the "development ship" instructions, probably just go with `~zod` in the prompt, and `my-fake-zod` as directory name (or whatever we commonly use in examples)?

matildepark: Ah, my mistake.

getting-started/_index.md
matildepark: Amended.

PR #526, opened 2019-06-19
library doc and intro gall doc

baudtack:


rmariani:
Can you make a PR with just libraries.md? I will handle separately the gall doc, and probably made a PR myself after my edits.

learn/hoon/hoon-tutorial/egg-timer.md
```diff
@@ -0,0 +1,120 @@
...
+`ost`  A reference to a chain of events
+`eny`  Entropy
+`now`  The current time of the event
+
+The `door` we've made has two arms `poke-noun` and `wake`. Gall is capable of dispatching `pokes`, or requests, to an app based on the mark of the data given along with that poke. These are sent to the arm with the name that matches the mark of the data. Here we use the generic `noun` mark:
+
+```
+++  poke-noun
+  |=  t=@dr
+  ^+  [*(list effect) +>.$]
```
Fang-: It doesn't actually matter much, but I'm curious: why this over `^-  [(list effect) _+>.$]`?

learn/hoon/hoon-tutorial/egg-timer.md
```diff
@@ -0,0 +1,120 @@
...
+`src`  The ship the request originated with
+`ost`  A reference to a chain of events
+`eny`  Entropy
+`now`  The current time of the event
+
+The `door` we've made has two arms `poke-noun` and `wake`. Gall is capable of dispatching `pokes`, or requests, to an app based on the mark of the data given along with that poke. These are sent to the arm with the name that matches the mark of the data. Here we use the generic `noun` mark:
+
+```
+++  poke-noun
+  |=  t=@dr
```
Fang-: If you're gonna use the `%noun` mark, but take in anything other than a `*`, then you should maybe mention how there's a mismatch there.

learn/hoon/hoon-tutorial/egg-timer.md
```diff
@@ -0,0 +1,120 @@
++++
+title = "2.7.1 Walkthrough: Egg Timer"
```
Fang-: Kinda sad at the title not containing "gall" or "app dev" or whatever. You know, for discoverability.

learn/hoon/hoon-tutorial/egg-timer.md
```diff
@@ -0,0 +1,120 @@
...
+`now`  The current time of the event
+
+The `door` we've made has two arms `poke-noun` and `wake`. Gall is capable of dispatching `pokes`, or requests, to an app based on the mark of the data given along with that poke. These are sent to the arm with the name that matches the mark of the data. Here we use the generic `noun` mark:
+
+```
+++  poke-noun
+  |=  t=@dr
+  ^+  [*(list effect) +>.$]
+  :_  +>.$  :_  ~
+  [ost %wait /(scot %da now) (add now t)]
```
Fang-: Why `/(scot %da now)` instead of a less opaque wire like `/egg-timer`? This (theoretically, I wouldn't encourage it for this tutorial per se) allows you to do `++wake-egg-timer`, or at least have a more readable `wire` in `++wake`. Using _just_ a plain, "random" timestamp for the wire seems ill-advised.

learn/hoon/hoon-tutorial/egg-timer.md
```diff
@@ -0,0 +1,120 @@
...
+
+The final part of this `syscall` is:
+
+```
+(add now t)
+```
+
+`now` is the current time of type `@da`, and `t` was declared as `@dr`. Because they are both atoms, we can add `now` and `t` these two to get an atom that is `t` units of time into the future from `now`. That produced atom can be interpreted as a `@da`.
+
+That's all for our `poke-noun` arm. But what about when the timer goes off? Behn will create a `move` similar to how we created one with `effect`, only this time it will end up being dispatched back to us in the `++wake` arm. Behn always looks for an arm named `++wake` when a timer goes off; it is defined this way in the vane's code. So any So any app that wants to use a timer trigger needs to have an arm called `++wake`.
```
Fang-: Technically incorrect in a couple ways, but I don't think we care much here?

learn/hoon/hoon-tutorial/egg-timer.md
baudtack: No particular reason.

learn/hoon/hoon-tutorial/egg-timer.md
baudtack: Eh. Yeah probably.

learn/hoon/hoon-tutorial/egg-timer.md
baudtack: It did but it got changed. maybe Gall Walkthrough or Gall Egg Timer or something

learn/hoon/hoon-tutorial/egg-timer.md
baudtack: given that we're not dispatching on it, i'm not sure that it matters?  but probably yeah.

learn/hoon/hoon-tutorial/egg-timer.md
baudtack: We absolutely do care. Please elaborate.

learn/hoon/hoon-tutorial/egg-timer.md
Fang-: It does indeed not matter technically, but it's semantically incorrect and kinda bad practice.

learn/hoon/hoon-tutorial/egg-timer.md
Fang-: > Behn will create a `move`
> Behn always looks for an arm named `++wake`
> it is defined this way in the vane's code.

The moves our application sends out don't go directly to the vanes. They first go through Gall, which then sends them off to the vanes (after stripping off the wire, which the vanes don't care about) in the form of vane `task`s.

If/when a vane wants to respond or otherwise talk to our application, it sends a `gift` back up the `duct` (aka call stack), so that ends up back in Gall. Gall then (after tacking the wire back on, because apps do care about it), figures out the appropriate app arm to call based on the `gift` contents, and attempts calling it.

I imagine that goes deeper into Gall/vanes than this particular doc cares about, so being a bit hand-wave-y here is fine. But, as it stands, this may perhaps be a bit misleading.

learn/hoon/hoon-tutorial/egg-timer.md
baudtack: Fair enough. If you use /egg-timer though don't you have a potential problem with say, setting two timers at once for different times?

learn/hoon/hoon-tutorial/egg-timer.md
baudtack: No, I think this does need some modification. I just misunderstood what vanes do in response to requests. Thanks. I'll make a pr.

learn/hoon/hoon-tutorial/egg-timer.md
Fang-: They'll still work if that's what you're asking. Duplicate wires are fine iirc.

PR #531, opened 2019-06-20
Minor technical corrections to egg-timer walkthrough

baudtack:


rmariani:
Oh I didn't see that you were also a reviewer :^)

Fang-:
_Oh_, I didn't see he tagged us both! Generally fair to merge after one review, but I always let the PR author decide when to merge to avoid these kinds of scenes. Whatever, all good!

PR #533, opened 2019-06-26
Some mark conversion documentation

matildepark:
Marks are currently fairly undocumented (removed with the `layout` page if I recall), but seem like a not insignificant aspect of creating applications. ~~That said, I sense it's in a transition phase, but in this interim period, it should still have some representation somewhere.~~ It appears that `&cord` conversions are now gone, as are `&md` conversions, but we can still convert Udon and `&hymn` and stack those conversions, too.

This PR, then, explains that mark conversions are written destination-first; that they can be stacked together; that some marks only convert to other marks; and that some marks essentially wrap / prepare a mark for conversion to its final, destination mark, using Udon as the source mark.

Feel free to adjust and propose if anything I've written is, well, _off the mark_.

### DIFF
--- a/using/shell.md
+++ b/using/shell.md
@@ -287,11 +287,25 @@ Which outputs a new `urbit.pill` to `pier/.urb/put/urbit.pill`
 
 ##### `&` - Mark conversion
 
-Convert between marks using `&`:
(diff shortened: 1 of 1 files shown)


belisarius222:
I don't think work is happening on the mark system now or soon. The experimental ford-fiesta branch modifies it a bit, but that will likely remain mothballed for at least another six months, if not longer.

—
~rovnys-ricfer
https://urbit.org

matildepark:
Ah, okay. Thank you, @belisarius222, I've corrected.

PR #534, opened 2019-06-26
Remove reference to comets in Hoon tutorial

matildepark:
Now that comets aren't the de facto introductory ship in the Hoon onboarding, referencing them is inconsistent. In my PR doing the former, I missed a reference to a comet — this PR smashes it back into dust. I love comets, but they have their place.


rmariani:
Good catch

PR #539, opened 2019-06-28
Update contributing.md

msutherl:
Add some guidance and remove outdated "Pentalogue".

### DIFF
--- a/getting-started/contributing.md
+++ b/getting-started/contributing.md
@@ -4,70 +4,22 @@ weight = 5
 template = "doc.html"
 description = "A guide to getting involved with the Urbit project."
 +++
+
 We accept submissions for improving our [code](https://github.com/urbit/urbit/) and [documentation](https://github.com/urbit/docs/) alike.
 
-If you're interested in contributing, but don't know where to start, feel free to [send us an email](mailto:support@urbit.org).
+If you're looking to get involved, there are a few things you can do:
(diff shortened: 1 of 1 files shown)


rmariani:
Curious about the discussion that led us to cut "the Pentalogue"

matildepark:
Second this question — the pentalogue is the de facto code of conduct. I think it still has a place (albeit, perhaps, underneath the content about how to contribute that's been provided here).

As to the Council, I wish them all the best.

msutherl:
@rmariani @matildepark there wasn't a discussion, but I've been meaning to remove it since I discovered it was there. Thing is, I have no idea where it came from, it's kind of weird, and it doesn't serve any purpose right now.

> The goal of these guidelines is to provide clear principles which can be used by moderators for the resolution of disputes in this jurisdiction.

We don't currently moderate the network in any meaningful way, and if we were, I'd rewrite this. For one, it seems to concern behavior "on the Urbit network," which is not a challenge we currently face.

Regardless, I don't think the 'contributing' section is the right place for this. If it were a traditional code of conduct as is found in many a Github repository, I'd move it to a CoC file in the urbit root directory or something like that.

msutherl:
I'm going to force this one through for now and happy to discuss out of band. I agree that having community guidelines is a nice thing.

PR #541, opened 2019-07-01
Fix links to runes in Irregular Forms page

adklempner:
Some of the links point to the page for the second char of the rune, which doesn't list the rune in question


rmariani:
Thanks!

PR #542, opened 2019-07-01
Adding cencol rune to its header.

matildepark:
Also places a period politely.

`### "cencol"` => `### %: "cencol"`


msutherl:
@matildepark @rmariani just a quick note about this – these changes broke some anchor links elsewhere. For those cases, I fixed it thusly:

`### "cencol"` => `### %: "cencol" {#cencol}`

That `{#...}` construction is Zola's way of setting anchor link IDs manually. This might also be unnecessary, but I couldn't figure out what the anchor link ID for `%: "cencol"` is.

PR #545, opened 2019-07-01
Adding runes with col suffixes to /rune/ pages

matildepark:
I kept running into this, so I checked them all, and they were each missing the col suffix rune — eg. `tiscol`, `barcol`, `buccol`, `colcol` et al. I added them.

I was wondering if this meant col suffix runes had rendering issues in the past. In my commit earlier today adding `cencol`, it rendered fine, so I checked the commit history, and it looks the col suffix runes were just a casualty of [this commit](https://github.com/urbit/docs/commit/f1229c17b955540fd2cbc760a26506f7b549406d#diff-8b4bfeb59044a1736c1846009587b102) during the transition from Udon to Markdown. In any case, this PR adds them back.

### DIFF
--- a/reference/hoon-expressions/rune/bar.md
+++ b/reference/hoon-expressions/rune/bar.md
@@ -163,7 +163,7 @@ A trivial core:
 101
 ```
 
-### "barcol"
+### |: "barcol"
 
 Produce a gate with a custom sample.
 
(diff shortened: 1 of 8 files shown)


msutherl:
Good catch – thanks!

PR #548, opened 2019-07-07
Added some dojo prompts for consistency

jfranklin9000:
Changed dojo prompts to what is actually displayed
Changed some examples to match existing filesystem
Fixed some universal address examples
Added missing section headers

I took the dojo prompts from the current era, not 0.8.0/cc-release.
I don't know if the prompts will change in 0.8.0.

`~lodleb-ritrul` could be changed to `~some-urbit`,
and `gen/example/hoon` no longer exists,
but I didn't change those.


rmariani:
a welcome change

PR #552, opened 2019-07-10
Fix broken link to next lesson in 2.2

adklempner:
The link at the bottom of 2.2 points to a missing page. I'm assuming the next link should point to 2.3


rmariani:
You assumed correctly!

PR #554, opened 2019-07-10
Remove dated reasoning for loobeans

Fang-:
This seems to have been carried over from older docs, but probably shouldn't have been.

(Why? It's stale, it's meme-y. And it's annoying.)


matildepark:
A similar rationale from the 2013 Urbit book:

image

PR #558, opened 2019-07-11
Fix links in 2.4

adklempner:
Adds links to standard library pages for Set Logic and Map Logic, along with link to next section, 2.5 Type Polymorphism


rmariani:
The succeeding page is not written yet. But I will make a PR myself and add those links

PR #560, opened 2019-07-11
Update getting-started/installing-urbit for 0.8.0.

rmariani:
Describes how to download static binaries for macOS and 64-bit Linux, and replaces the instructions for other operating systems with a catch-all using Nix and nixpkgs.


rmariani:
(Do not merge this PR until 0.8.0 is out)

jtobin:
@rmariani I added another commit that should also go in at the same time here -- it patches up some of the URLs that point at the soon-to-be-deprecated arvo repository.

(Also I just happened to notice a missing header that should point at the docs repo, so fixed that up too.)

rmariani:
Great, thanks!

PR #561, opened 2019-07-11
Change '/lib' example from 'react' to 'number-to-words' in using/shel…

jfranklin9000:
…l.md

`lib/react.hoon` does not exist

### DIFF
--- a/using/shell.md
+++ b/using/shell.md
@@ -376,9 +376,9 @@ with `/+`. Read / write.
 **Examples:**
 
 ```
-~your-urbit:dojo> /+  react
-~your-urbit:dojo> ::  now we can use arms from lib/react.hoon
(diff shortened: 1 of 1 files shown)


using/shell.md
```diff
@@ -376,9 +376,9 @@ with `/+`. Read / write.
 **Examples:**
 
 ```
-~your-urbit:dojo> /+  react
-~your-urbit:dojo> ::  now we can use arms from lib/react.hoon
-~your-urbit:dojo> (react-vale:react %div)
+~your-urbit:dojo> /+  number-to-words
+~your-urbit:dojo> ::  now we can use arms from lib/number-to-words.hoon
```
joemfb: FWIW, standalone comments like this don't work in :dojo.

PR #565, opened 2019-07-17
Rune naming fixes in reference/hoon-expressions/irregular.md

jfranklin9000:
I'm looking for the discussion with the final decision on character names.  Did ban/led win over gal/gar?  `hoon.hoon` suggests so but we have  some docs that still use gal/gar.

### DIFF
--- a/reference/hoon-expressions/irregular.md
+++ b/reference/hoon-expressions/irregular.md
@@ -26,7 +26,7 @@ I: `+(p)`
 ### .= dottis
 
 [docs](@/docs/reference/hoon-expressions/rune/dot.md) \\=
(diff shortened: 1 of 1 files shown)


ohAitch:
I believe gal/gar/sel/ser/pat etc won, and hoon.hoon just hasn't been
updated yet

jfranklin9000:
Is the table in https://urbit.org/docs/learn/hoon/hoon-tutorial/hoon-syntax/ correct as far as you recall?

ohAitch:
Yep - it should be reordered so mic is in its alphabetical place, but
that(and plpr vowel) ended up as the only thing changed from 2015 iirc

rmariani:
handled in https://github.com/urbit/docs/pull/567

PR #571, opened 2019-07-18
added text editor tutorial to hoon 1.1

syzygyzer:
noobs need to know they need a text editor, so here is a short intro to them and a list of text editors with hoon support


rmariani:
the `#####` header is really small (smaller than normal text), so we don't use it. Can you combine the "Text Editors" and "List of Editors" sections, so that `####` can be the maximum header-depth? You can make the combined description of both more laconic, which I think is a good idea anyway for the purposes of economizing space.

rmariani:
Went ahead and made the changes myself :)

PR #588, opened 2019-07-31
Azimuth typos + removing "Aegean" mention

matildepark:
Was reading the Azimuth article, saw some stuff to amend -- like linking to the Node repos (and changing the name of the second one), changing "Aegean" to "Landscape" (there's no other mentions of Aegean in the docs and Landscape best fits what it describes), and so I am putting in a PR here.

### DIFF
--- a/concepts/azimuth.md
+++ b/concepts/azimuth.md
@@ -6,11 +6,11 @@ description = "An explanation of the Urbit identity layer."
 +++
 Azimuth is a general-purpose public-key infrastructure (PKI) on the Ethereum blockchain, used as a platform for Urbit identities that we call _points_. You need a point to use the Arvo network.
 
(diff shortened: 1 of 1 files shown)


rmariani:
Aegean is a different concept from Landscape AFAIK: https://urbit.org/primer/#what-urbit-is-for

We never really talk about Aegean, so it's probably worth cutting the reference.

matildepark:
Oh, interesting. I see, I see. Isn't "the pattern for **creating software experiences** for individual Urbit communities" Landscape, then, and Aegean the pattern of societies themselves?

Dunno. I can just remove Landscape from this.

matildepark:
Ok, I just rephrased it so if/when Aegean comes up we can add it in seamlessly later. Leaving Aegean in with no other references to it in the Docs is less preferable and Landscape is less exact.

rmariani:
Landscape is one software experience, one that we made.

But looks good, thanks

PR #590, opened 2019-08-01
adds link to moon-spawning instructions, adds PATH advice for install…

rmariani:
…ation.


matildepark:
I think the PATH advice here would break Urbit booting [for the short term](https://github.com/urbit/urbit/issues/1369#issuecomment-516296030). Unsure when that's changing, but apparently a fix is coming...?

using/operating-a-star.md
```diff
@@ -27,7 +29,23 @@ Below are some handy operations for the use and maintenance of your star.
 
 ### DNS Proxying
```
joemfb: The "DNS Proxying" terminology used here in the past conflated the dynamic DNS records and TCP proxying that constituted arvo.network -- it's not really a thing. I'd change this title to "arvo.network DNS records", or something like that.

using/operating-a-star.md
```diff
@@ -27,7 +29,23 @@ Below are some handy operations for the use and maintenance of your star.
...
+
+Star (and planet) operators can request a domain name for their ship under as `arvo.network` (so `~marzod` would be `marzod.arvo.network`). For such a request to be made and fulfilled, they must be hosting their ship someplace with a public IP address, and it's http server must be listening on port 80.
+
+If you're a star operator who wants to set up DNS routing with your sponsor galaxy, you must run:
+
+```
+> :dns|request
+```
+
+You'll be prompted to enter the public ip address of the ship. You can also pass the ip address as an argument, using the @if syntax. For example:
```
joemfb: `@if` needs backticks

rmariani:
> I think the PATH advice here would break Urbit booting [for the short term](https://github.com/urbit/urbit/issues/1369#issuecomment-516296030). Unsure when that's changing, but apparently a fix is coming...?

@belisarius222 @vvisigoth @philipcmonk Please advise on this

rmariani:
also @jtobin

joemfb:
@rmariani, thanks for the DNS fixes.

@matildepark is right, the release binaries of urbit currently don't work when run out of $PATH due to urbit/urbit#1369. This is bad, and we will fix it. In the meantime, this line will be confusing to users, who should follow the current install instructions *exactly*.

rmariani:
Alright, gonna mothball this then. Thanks @joemfb

PR #595, opened 2019-08-02
Fix typo in setup.md

Foltik:
The docs say to run a generator `mygen.hoon` you need to type `+myhoon <argument>` in the Dojo, but if I'm not mistaken it should be `+mygen <argument>`


rmariani:
Good catch!

PR #604, opened 2019-08-07
Add summaries and metadata to the entire standard library list

### LINKED ISSUE #601: Tooltips?
Looking at the [standard library page](https://urbit.org/docs/reference/library/) is a bit overwhelming to me at the moment, but it seems rather integral to the developer experience.

It seems like it would be a good idea to incorporate tooltips, so mousing over the library title would show its name. 

Likewise, this could also be a good thing for codeblocks -- imagine mousing over runes, and having it explain what it is and does.

## Implementation

My initial idea for how this could be implemented would be essentially to use `data-tooltip="Some text here"` attributes on links and pieces of text that should explain other things. In JavaScript, a script would quickly scan the DOM, and if an element has that attribute, it would create an event-listener for it to render an (absolutely positioned) tooltip on mouseover.

This is, however, unscalable as documentation changes, as it repeats information that is in rapid change.

We would want fewer repeated pieces of information -- so what would be *best* is if a piece of text, or a link, is labeled for `data-tooltip=true` and its associated `href` would have a `description` element to scan for. 

Then in JavaScript, it would scan the DOM, and add a *single* event listener that on mouseover would get the page and print the description in a tooltip for all applicable elements (much like Wikipedia currently does on article mouseover). 

This could also work well for any piece of arbitrary text to have a tooltip as long as they are given a source page to pull from.

### Downsides

This essentially makes `/reference/library/_index.md` not suitable for Markdown link syntax -- adding attributes is outside its scope. It could be a good case for shortcodes to make the writing experience a bit cleaner.

It would also add a bit of syntactic baggage to code blocks -- unless we were to codify it *once* in `scripts.js`, so that in JavaScript it *always* processes runes inside code blocks and adds tooltips to them.

What do you think?

matildepark:
Fixes #601. This PR, in conjunction with [urbit.org#163](https://github.com/urbit/urbit.org/pull/163), provides tooltips and metadata for every arm of the standard library.

- Converts all links in _index.md from inline Markdown to HTML
- Adds tooltip class to those links
- Provides new, brief descriptions for every arm of the library as `title` attributes (yes, I went to each page, and consistently and succinctly described the functionality; if I couldn't, I just used the summary the documentation itself used)

## The good

- This PR's text is de facto metadata for the entire standard library functionality, organised by its term. If you use some regex, you can transform this into JSON trivially and make it into anything you like — up to and including IDE extensions. VSCode could mouse over `flop`, `vum:ag` or `lust` and tell you what they mean, for example.

## The bad

- HTML is harder to read when editing — but hey, it's a single page.
- When stdlib changes in future, it must be considered part of the workflow to **add a title attribute to its link on this page**. There are two listings — alphabetical and per section — so find and replace is a necessity.

### DIFF
--- a/reference/library/_index.md
+++ b/reference/library/_index.md
@@ -13,1718 +13,1718 @@ Can be navigated [alphabetically](#alphabetical) or [by section](#by-section).
 
 ### a
 
-<a class="tooltip" href='https://urbit.org/docs/reference/library/4j/#ab' title="Primitive parser engine"><code>++ab</code></a>
-<a class="tooltip" href='https://urbit.org/docs/reference/library/3b/#abs-fl' title="Absolute value"><code>++abs:fl</code></a>
-<a class="tooltip" href='https://urbit.org/docs/reference/library/3a/#abs-si' title="Absolute value (signed integer)"><code>++abs:si</code></a>
-<a class="tooltip" href='https://urbit.org/docs/reference/library/4o/#abel' title="Compiler alias"><code>++abel</code></a>
-<a class="tooltip" href='https://urbit.org/docs/reference/library/4h/#ace' title="Parse space"><code>++ace</code></a>
-<a class="tooltip" href='https://urbit.org/docs/reference/library/1a/#add' title="Add"><code>++add</code></a>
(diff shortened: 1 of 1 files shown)


matildepark:
Oh, right. Fixes #601.

rmariani:
LGTM. Thanks Matilde!

PR #615, opened 2019-08-14
Replaced calls to "caesar" with "shift"/"unshift"

Kevingislason:


rmariani:
Thanks for catching this!

PR #617, opened 2019-08-16
update cass and cuss

cxk280:
The docs incorrectly state that `++cass` and `++cuss` returns `cords`. They return `tapes`. This PR also updates the relevant examples, removing some that actually `nest-fail`, and synchronizes them between the two methods.

### DIFF
--- a/reference/library/4b.md
+++ b/reference/library/4b.md
@@ -185,7 +185,8 @@ XX document
 
 To lowercase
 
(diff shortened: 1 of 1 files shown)


cxk280:
The relevant examples also need to be updated. I'll get that pushed up as well shortly.

rmariani:
Thanks for catching this!

PR #619, opened 2019-08-19
add generators tutorial fixes

cxk280:
This PR removes the redundant use of `sole-request:sole` instead of simply `sole-request` from the `%get` generator example. It also makes some other small clarifications in the text.

### DIFF
--- a/learn/hoon/hoon-tutorial/generators.md
+++ b/learn/hoon/hoon-tutorial/generators.md
@@ -463,7 +463,7 @@ we can jump in with an example:
 =,  generators
 :-  %get
 |=  [* [url=tape ~] ~]
-^-  (sole-request:sole (cask json))
+^-  (sole-request (cask json))
 %+  curl  (scan url auri:de-purl:html)
 |=  hit/httr:eyre
 ?~  r.hit  !!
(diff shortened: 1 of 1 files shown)


learn/hoon/hoon-tutorial/generators.md
```diff
@@ -463,7 +463,7 @@ we can jump in with an example:
 =,  generators
 :-  %get
 |=  [* [url=tape ~] ~]
-^-  (sole-request:sole (cask json))
+^-  (sole-request (cask json))
```
baudtack: This is incorrect.  `sole` is not exposed here and this code fails with `find.sole-request`

learn/hoon/hoon-tutorial/generators.md
```diff
@@ -502,11 +502,11 @@ and the tail is a gate that takes a `tape` of the URL that we are trying
 to access.
 
 ```
-^-  (sole-request:sole (cask json))
+^-  (sole-request (cask json))
 ```
 
 Here we make sure our gate is producing a `sole-request` which contains a `cask`
-of the type `json`.
+of the type `json`. Please note that since we exposed the `sole` namespace with `=,`, we do not have to write `sole-request:sole` here.
```
baudtack: This is not correct. =, was not used to expose sole

PR #620, opened 2019-08-20
Sail and Udon docs: fix links, simulate udon rendering with

jfranklin9000:
markdown (double quotes, block quote), remove incorrect info
(inline italicization does not work in block quotes), use
;br; instead of \<br\> in sail expression example source.

Here is output of the **udon** renderer (`|static`) showing that italics
render in block quotes (ignore the green block quote background):

udon-block-quote

### DIFF
--- a/using/sail-and-udon.md
+++ b/using/sail-and-udon.md
@@ -967,12 +967,12 @@ destination.
 **Example:**
 
 ```
-[I'm an inline-style link](https://www.urbit.org/docs/reference/shell)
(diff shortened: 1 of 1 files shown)


baudtack:
Sorry, confused here. Are those smart quotes that are in the doc?

jfranklin9000:
Normal double quotes in the doc are rendered as smart quotes in udon.  We are simulating udon with markdown in this part of the patch.

Here is output of the **udon** renderer:

smart-quotes

We have to put the smart quotes right in the (markdown) doc to show what the udon parser will do.

PR #626, opened 2019-08-30
kiss -> task

ryjm:
Removes references to `kiss` and replacing them with `task` where appropriate - protects readers from trying to figure out what the hell a `kiss` is (which was replaced with `task` a while back).


rmariani:
Thanks for this fix!

PR #640, opened 2019-09-23
Represent ace as 14 instead of 1 in libraries.md

pkova:
Mainly for hoon school 201 week 1. It's kinda annoying ranking poker hands with aces as 1, even though you of course have to check both ways when it comes to straights.


rmariani:
Thanks Pyry!

PR #644, opened 2019-09-26
fixed syntax update errors in library 2a section

jfranklin9000:
`/` was changed to `]` instead of `=`


rmariani:
Thanks for this

PR #655, opened 2019-10-12
draft of gall async monad doc

baudtack:


matildepark:
This uses old IA and should be moved to `/tutorials/hoon`.

PR #660, opened 2019-10-17
re-added tutorial on booting a ship

syzygyzer:
as far as I can tell we lost this page in the shuffle. I don't actually know if the link to `/using/creating-a-development-ship/` works or not, since its now under the `urbit.org` repo instead of `/docs` and I don't actually know how to test this, so I need someone who knows how to check this to do so.


rmariani:
Good catch, poprox, but this information belongs under the `/using` section of urbit.org.

PR #667, opened 2019-10-21
adding ford lesson

syzygyzer:
This was done the dumb way since the original fork was done with the old file organization.


rmariani:
Can you update the index page to include this in the chapter 2 list?

PR #670, opened 2019-10-22
fix link, add hoon highlighting

syzygyzer:


syzygyzer:
went too aggressive, highlighted dojo prompts, fixing that

PR #685, opened 2019-10-28
fix links

syzygyzer:
didn't realize we don't need the .md in links anymore

### DIFF
--- a/tutorials/hoon/test-sets.md
+++ b/tutorials/hoon/test-sets.md
@@ -8,7 +8,7 @@ template = "doc.html"
 
 In this walkthrough we will show how to use Ford as a testing suite.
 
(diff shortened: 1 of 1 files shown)


matildepark:
To be clear on this:

You only need to mention `.md` when using [Zola's internal link syntax](https://www.getzola.org/documentation/content/linking/#internal-links), prefacing the internal link with `@`. Otherwise, you shouldn't, because the permalink is `/permalink/` not `/permalink.md`.

Using `@` is good, because then it won't build the site unless all the internal links work — this sounds annoying, but it enforces that we don't have broken links all over the place, and it's easy to catch and fix on my part.

We have no firm rule on this — it's up to your preference given the above information.

syzygyzer:
Thanks, didn't know that about `@`, will push another commit changing to that style

PR #702, opened 2019-11-05
Fix tutorial links

syzygyzer:
also fixes some typos and style errors

### DIFF
--- a/tutorials/hoon/bank-account.md
+++ b/tutorials/hoon/bank-account.md
@@ -75,6 +75,6 @@ Each of these arms produces a gate which takes an `@ud` argument. Each of these 
 +>.$(balance (add balance amount))
 ```
 
-`+>` is [wing syntax](@/docs/reference/hoon-expressions/limb/wing/). This particularly wing construction looks for the the tail of the tail (the third element) in `$`, the subject of the gate we are in, which is the entire `new-account` door. We change `balance` to be the result of adding `balance` and `amount` and produce the door as the result. `withdraw` functions the same way only doing subtraction instead of addition.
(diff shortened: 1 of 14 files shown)


tutorials/hoon/bank-account.md
```diff
@@ -75,6 +75,6 @@ Each of these arms produces a gate which takes an `@ud` argument. Each of these
 +>.$(balance (add balance amount))
 ```
 
-`+>` is [wing syntax](/docs/reference/hoon-expressions/limb/limb/). This particularly wing construction looks for the the tail of the tail (the third element) in `$`, the subject of the gate we are in, which is the entire `new-account` door. We change `balance` to be the result of adding `balance` and `amount` and produce the door as the result. `withdraw` functions the same way only doing subtraction instead of addition.
+`+>` is [wing syntax](@/docs/reference/hoon-expressions/limb/wing/). This particularly wing construction looks for the the tail of the tail (the third element) in `$`, the subject of the gate we are in, which is the entire `new-account` door. We change `balance` to be the result of adding `balance` and `amount` and produce the door as the result. `withdraw` functions the same way only doing subtraction instead of addition.
```
rmariani: Is this the right link syntax?

matildepark: No, needs to end with `.md`. You're linking to the file's relative path, and Zola converts it to the permalink.

rmariani: I thought so. @syzygyzer, can you fix this please?

tutorials/hoon/hoon-syntax.md
```diff
@@ -307,21 +307,21 @@ Standard library functions are often built with other standard library functions
 
 ## Reference Materials
 
-The Hoon syntax can be intimidating for the uninitiated, so it's good to remember where you can look up expressions can be found. The [reference section](/docs/reference/) itself is a good place to find the reference materials that you need. These children sections are likely to be useful:
+The Hoon syntax can be intimidating for the uninitiated, so it's good to remember where you can look up expressions can be found. The [reference section](@/docs/reference/) itself is a good place to find the reference materials that you need. These children sections are likely to be useful:
```
rmariani: `@/docs/reference/` needs to be fixed too

PR #710, opened 2019-11-07
rewrites section on nested cores

syzygyzer:
The section on nesting cores was a bit misleading and didn't elaborate much on why you would do it. I've rewritten it to be more explicit about how core nesting works in the standard library and how it affects visibility of arms.

I did have a couple of exercises that I wrote a few days ago that I removed. I couldn't figure out why the first one behaved as it did and neither did @baudtack, so it seemed too technical to be included. The second one was rolled into the rewritten section.

### DIFF
--- a/tutorials/hoon/arms-and-cores.md
+++ b/tutorials/hoon/arms-and-cores.md
@@ -181,7 +181,7 @@ To reinforce your newfound understanding of arms and cores, let's go over the va
 
 ### Address-Based Wings
 
-In the last lesson, you saw how the following expressions return legs based on an address in the subject: `+n`, `.`, `-`, `+`, `+>`, `+<`, `->`, `-<`, `&`, `|` etc.  When these resolve to the part of the subject containing an arm, they **don't** evaluate the arm.  They simply return the indicated noun fragment of the subject, as if it were a leg.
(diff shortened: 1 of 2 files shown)


tutorials/hoon/arms-and-cores.md
```diff
@@ -410,7 +410,7 @@ This almost looks like a function call of sorts.
 
 Let's take a quick look at the battery of one core in the dojo to show that this is true by inputting one into the dojo.
 
-```hoon
+```
```
baudtack: Why did you remove the hoon tag here? does it color dojo stuff weirdly?

syzygyzer: As near as I've been able to tell, 95%+ of the time dojo output appear in the docs it is without the hoon tag. So I've been removing the other 5% when I see them. They are technically different things. I think if we want to also highlight dojo output we should create a separate color scheme for it to emphasize the difference.

syzygyzer:
I thought of a way this could be improved - what is a dojo command I can use that will explicitly show that e.g. add is in the subject of biff but not vice versa?

joemfb:
```
> add:biff
<1.vwd {{a/@ b/@} <41.mac 1.ane $141>}>
> biff:add
/~master-morzod/home/~2019.11.6..19.05.32..6ebd/sys/vane/ford:<[4.828 18].[4.828 59]>
-find.biff
```

tutorials/hoon/arms-and-cores.md
matildepark: In general when I was seeding the code blocks with metadata I skipped all the Dojo prompts; Dojo syntax isn't identical to Hoon syntax and I wanted to avoid potentially wonky results. I don't know. I tried this code block in my VS Code's Hoon syntax highlighting and it looked alright. So maybe I was off.

syzygyzer:
I moved part of doors.md to arms-and-cores.md, about using the `..arm` syntax. This is required to understand the content on nesting cores.

PR #717, opened 2019-11-12
section on nesting runes, lego block diagrams

syzygyzer:
This gives a little more mental scaffolding for the new Hooner. The only thing I am unsure of about this PR is whether I need to do anything else to make sure that the images are centered, since markdown down not support centering.


syzygyzer:
I added in a bit of a "mathy" exercise. I tried to phrase it loosely enough that the student should be able to determine to what extent they can do this exercise. For example, just being able to draw a few examples would be enough for many people, but a more rigorous student may try to write down explicitly the two sets and find the bijection. For that reason I did not include the solution, but I'm not sure if this is a good choice or not.

PR #719, opened 2019-11-14
Egg timer code nit & corrections

Fang-:
This triggered me. Also we were using `card` in place of `move` a bunch? Pretty sure the code wouldn't compile.

I have a couple other style changes I'd make, but they're a bit more aggressive, so just listing them here instead:
- `(pair bone card)` -> `[bone card]` because you never ever care about accessing these and `+pair` isn't exactly modern hoon.
- `+>.$` -> `+*  this  .` etc.
- `:_  ~  move` ->  `[move]~`, double-`:_`s are pretty hard to follow imo, and `[item]~` lists are pretty common in these contexts. (Alternatively, `[move ~]`.)

### DIFF
--- a/tutorials/hoon/egg-timer.md
+++ b/tutorials/hoon/egg-timer.md
@@ -21,7 +21,7 @@ The Arvo operating system is divided up into modules called vanes. Gall is the v
   [ost %wait /egg-timer (add now t)]
 ++  wake
   |=  [=wire error=(unit tang)]
-  ^+  [*(list card) +>.$]
+  ^+  [*(list move) +>.$]
   ~&  "Timer went off!"
   [~ +>.$]
 --
@@ -59,7 +59,7 @@ The `door` we've made has two arms `poke-noun` and `wake`. Gall is capable of di
 ```hoon
 ++  poke-noun
   |=  t=@dr
-  ^+  [*(list card) +>.$]
+  ^+  [*(list move) +>.$]
   :_  +>.$  :_  ~
   [ost %wait /egg-timer (add now t)]
 ```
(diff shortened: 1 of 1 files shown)


tutorials/hoon/egg-timer.md
```diff
@@ -96,12 +96,12 @@ The final part of this `move` is:
 
 `now` is the current time of type `@da`, and `t` was declared as `@dr`. Because they are both atoms, we can add `now` and `t` these two to get an atom that is `t` units of time into the future from `now`. That produced atom can be interpreted as a `@da`.
 
-That's all for our `poke-noun` arm. But what about when the timer goes off? Behn will create a `gift`, a similar construct to how we created an `card`, only this time it will end up being dispatched back to us via Gall in the `++wake` arm. Any app that wants to use a timer trigger needs to have an arm called `++wake`.
+That's all for our `poke-noun` arm. But what about when the timer goes off? Behn will create a `gift`, a similar construct to how we created a `card`, only this time it will end up being dispatched back to us via Gall in the `++wake` arm. Any app that wants to use a timer trigger needs to have an arm called `++wake`.
```
ohAitch: move?

Fang-: I left this uncorrected because a `move` is just a `card` on a """path""" (`bone`), ie it's the effect itself, which is also true for `gift`.

tutorials/hoon/egg-timer.md
```diff
@@ -31,7 +31,7 @@ The first thing to notice is that we are creating a `core` (`|%`) and a `door` (
 
 ```hoon
 |%
-+$  card   (pair bone move)
++$  card  (pair bone move)
 +$  move  [%wait path @da]
```
ohAitch: Okay right this is just backwards and I think the source of my confusion

Fang-: Oh god I touched this and didn't even notice. Hah, thanks for the heads up!

ohAitch: At least that answers the question of which parts of the code would and wouldn't compile!

ohAitch:
Uh to reiterate, this branch currently contains the text "An `card` is a `pair` of `bone` and `move`" in contradiction with the corrected code, and did not seem ready for merge.

PR #723, opened 2019-11-15
Add mount instructions to setup page

Fang-:
It's talking about committing things to home desk, but doesn't actually tell you to have the desk mounted first.


syzygyzer:
What is the difference between `|mount /=home=` and `|mount %`? Previously setup instructions said to use `|mount %`.

Fang-:
`%` resolves to your current directory. By default, that's `/=home=` (aka latest revision of home desk on our ship), but the user can change it. As such, `/=home=` fully and explicitly is more reliable, and makes it more clear what's happening here.

PR #724, opened 2019-11-17
Examples fixes for buc runes

jfranklin9000:
The buccen **Discussion** states:
> Make sure the first item in your $% terminates, or the default will be an infinite loop!

The example for the default (`$:foo`) was wrong so I'm wondering if the text should say "last" instead of "first".

What does it mean to "terminate"?  A non-recursive structure?  If someone can give me an example of a non-terminating item I'll try it in the first and last spots and update the text, if needed.

### DIFF
--- a/reference/hoon-expressions/rune/buc.md
+++ b/reference/hoon-expressions/rune/buc.md
@@ -104,26 +104,30 @@ Regular form: **2-running**.
 
 A `$%` is a tagged union, a common data model.
 
-Make sure the first item in your `$%` terminates, or the default will
-be an infinite loop!
+Make sure the last item in your `$%` terminates, or the default will
+be an infinite loop!  Alteratively, you can use `$~` to define a custom
+type default value.
 
 ##### Examples
 
 ```
 ~zod:dojo> =foo $%([%foo p=@ud q=@ud] [%baz p=@ud])
 
+~zod:dojo> (foo [%foo 4 2])
+[%foo p=4 q=2]
+
 ~zod:dojo> (foo [%baz 37])
 [%baz p=37]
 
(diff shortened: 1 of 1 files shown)


Fang-:
> What does it mean to "terminate"? A non-recursive structure?

Correct. I think a non-terminating example would be something like:

```hoon
+$  step
  $%  [%split a=step b=step]
      ...
  ==
```

Not necessarily in scope for this PR, but might be worth referring to `$~` in this discussion, since it can be used to set a custom default, giving you an alternative to "reorder the `$%`".

jfranklin9000:
Testing with `gen/naked/hoon`:
```
|%
+$  buccen
  $%  [%fir bc=buccen]  ::  works
      [%mid @]
      [%las @]
  ==
--
|=  a=@
  [a *buccen]
```
```
> +naked 7
[7 [%las 0]]
```
```
|%
+$  buccen
  $%  [%fir @]
      [%mid bc=buccen]  ::  works
      [%las @]
  ==
--
|=  a=@
  [a *buccen]
```
```
> +naked 7
[7 [%las 0]]
```
```
|%
+$  buccen
  $%  [%fir @]
      [%mid @]
      [%las bc=buccen]  ::  inf-loop
  ==
--
|=  a=@
  [a *buccen]
```
```
> +naked 7
bail: meme
bailing out
pier: work error: stream closed
pier: work exit: status 0, signal 6
```
Conclusion: the text needs to be changed from "first" to "last" (done).

joemfb:
Last is correct (as of the most recent round of language changes): https://github.com/urbit/urbit/blob/master/pkg/arvo/sys/hoon.hoon#L8003

Issue #727, opened 2019-11-20
conditionals tutorial nitpicking


belisarius222:
The example function in the [conditionals tutorial](https://urbit.org/docs/tutorials/hoon/conditionals/) is a little strange:
```
:-  %say
|=  [* [n=@ud ~] ~]
:-  %noun
?:  ?&  =(0 (mod n 2))
        ?&  (gte n 1)
            ?!  (gth n 100)
        ==
    ==
    %.y
%.n
```

I'd write that as:
```
:-  %say
|=  [* [n=@ud ~] ~]
:-  %noun
^-  ?
?&  =(0 (mod n 2))
    (gte n 1)
    (lte n 1000)
==
```

Differences:
- `?&` is n-ary, not binary, so nesting two of them is not idiomatic
- `lte` is more succinct than negated `gth`
- outer `?:` is superfluous
- I added a cast, since that's good practice for almost every gate

To make use of `?!` and nested conditional expressions, you could have a couple extra examples that build off this one.

Write a function that produces `%.n` if `n` is an odd number between 1 and 1000, and `%.y` otherwise (this is the negated version of the previous function, so see the added `?!`).
```
:-  %say
|=  [* [n=@ud ~] ~]
:-  %noun
^-  ?
?!
?&  =(0 (mod n 2))
    (gte n 1)
    (lte n 1000)
==
```

If you want to wrap it in a `?:`, you could ask for a function that produces `%valid` if n is an odd number between 1 and 1000, and `%invalid` otherwise.  We clearly don't cast this output to boolean.

```
:-  %say
|=  [* [n=@ud ~] ~]
:-  %noun
?:  ?&  =(0 (mod n 2))
        (gte n 1)
        (lte n 1000)
    ==
  %valid
%invalid
```

That might not be a bad time to mention that you could make a boolean variable `.valid` if you prefer that style.
```
:-  %say
|=  [* [n=@ud ~] ~]
:-  %noun
=/  valid=?
  ?&  =(0 (mod n 2))
      (gte n 1)
      (lte n 1000)
  ==
?:(valid %valid %invalid)
```

PR #728, opened 2019-11-20
Conditionals

syzygyzer:
Addresses #727 . @belisarius222 's points are good and I agree that it is better to present idiomatic code. It's also good to present more than one `wut` rune, moved the old code (with minor changes) to the "Other Runes" section, showing how to use `?:` and `?!` but also saying that doing so in this manner would be poor practice.

### DIFF
--- a/tutorials/hoon/conditionals.md
+++ b/tutorials/hoon/conditionals.md
@@ -12,9 +12,10 @@ In this lesson, we will write a generator that takes an integer and checks if it
 |=  [* [n=@ud ~] ~]
 :-  %noun
 ^-  ?
-?&  =(0 (mod n 2))
+?&
+    =(0 (mod n 2))
     (gte n 1)
-    (lte n 1000)
+    (lte n 100)
 ==
 ```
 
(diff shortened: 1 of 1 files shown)


syzygyzer:
Oh yeah I'm still unsure whether to refer to the `?` type as booleans, loobeans, or `flag`s.

belisarius222:
Flag is the clear winner

—
~rovnys-ricfer
https://urbit.org

tutorials/hoon/conditionals.md
```diff
@@ -11,13 +11,12 @@ In this lesson, we will write a generator that takes an integer and checks if it
...
-?:  ?&  =(0 (mod n 2))
-        ?&  (gte n 1)
-            ?!  (gth n 100)
-        ==
-    ==
-    %.y
-%.n
+^-  ?
+?&
+    =(0 (mod n 2))
```
syzygyzer: I put `=(0 (mod n 2))` on a separate line from `?&` for the sake of the following explanation, but this is bad practice (I think) so I'm unsure about this choice and whether its worth making the explanation slightly more confusing by using `?&  =(0 (mod n 2))` instead

belisarius222:
Putting the conditional on the next line would be nonstandard practice. How does it affect the explanation, though?

—
~rovnys-ricfer
https://urbit.org

syzygyzer:
I worry that some new Hooners might think that the first conditional is somehow privileged because it is on the same line as `?&` and this is a very early lesson so there's lots of room for misunderstanding.

On the other hand, it seems inconsistent to revise this code because it is not idiomatic and replace it with other code that is still not idiomatic.

Probably the right thing to do is to put it on the same line and then just emphasize in the explanation that the ordering of the children does not matter and there is nothing special about being on the same line as `?&`. This would also be a good spot to emphasize putting heavier lines at the bottom.

PR #731, opened 2019-11-22
Simplify language with E-Prime

thunderrabbit:
Simplify this sentence by removing "what ... is"

https://en.wikipedia.org/wiki/E-Prime


thunderrabbit:
The online diff does not highlight the change I made:

OLD: ~What~ the return value ~is~ depends solely on the argument(s), and nothing else.

NEW: The return value depends solely on the argument(s), and nothing else.

PR #739, opened 2019-11-26
updates arvo/gall material with agent.txt

syzygyzer:
This address #738. This is pretty much a copy and paste of @philipcmonk's agent.txt, I just added some hoon tags to some code blocks and made it clear where other words were names of arms.

As I mention in #738, I think this is high enough quality to just publish it as the new arvo/gall page at the same time Gall is updated with the breach, just so we can have current documentation (and given that the last major revision on our arvo/gall page was in 2016). I'll revisit it in the next few weeks for more serious editing.

### DIFF
--- a/tutorials/arvo/gall.md
+++ b/tutorials/arvo/gall.md
@@ -64,7 +64,7 @@ definable in a regular recursive data type).
 An agent is defined as a core with a set of arms to handle various
 events.  These handlers usually produce a list of effects and the next
 state of the agent.  The interface definition can be found in
-sys/zuse.hoon, which at the time of writing is:
+`sys/zuse.hoon`, which at the time of writing is:
 
 ```hoon
 ++  agent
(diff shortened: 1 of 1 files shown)


syzygyzer:
One issue with this that I forgot to mention is whether we use `+foo` or `++foo` to refer to arms. I think the style we use most frequently in the current documentation is `++`, but the original `agents.txt` used `+` and I see this style used in e.g. urbit-dev as well. I don't really have a preference either way, and perhaps there is a solid argument for one or the other that I'm unaware of. But it's something we need to decide one way or the other and throw it in a style guide.

philipcmonk:
++foo was the old convention when there wasn't +$, +* or the other arm types.  +foo is newer and applies to any of those.  Technically I wouldn't consider it "wrong" to use ++foo if it is actually a ++, but it's uncommon now.  I do find myself saying "slus foo" still, though.

ohAitch:
The solid argument against +foo is +foo means gen/foo; the argument against ++foo is yeah it means you don't have to remember whether it's ++ or +* (which I think is now deprecated?) or +$

PR #741, opened 2019-11-29
Update the-subject-and-its-legs.md

risruc-habteb:
Correct answer for 1.6a, which did not match the exercise statement.


risruc-habteb:
Here's to hoping I've done this properly. My first go.
Starting small!

PR #753, opened 2019-12-10
add gall reference docs

liam-fitzgerald:
Added gall reference documentation.

Documents each of the gall arms with an example app. I struggled to write meaningful examples for the reference documentation without a wider context, so I have written a small omnibus app. It uses almost all the features of gall, so the reader can understand the examples with context they can understand. Somewhat plagiarised from agent.txt, but has been reworked to fit more of a reference style document. 

cc @rmariani

### DIFF
--- a/reference/gall.md
+++ b/reference/gall.md
@@ -1,0 +1,765 @@
+# Gall
+
+## Reference Agent
+
+This app stores a number for each user. You can increment your number, increment
+someone else's number, start tracking someone else's number by poking. It
+exposes a scry namespace to inspect everyone's number.
+
+### `app/example-gall.hoon`
+
+```hoon
+/-  *example-gall
+/+  default-agent, verb
+|%
++$  card  card:agent:gall
++$  note
+  $%
+    [%arvo =note-arvo]
+    [%agent [=ship name=term] =task:agent:gall]
+  ==
++$  state  [local=@ud remote=(map @p @ud)]
+--
+=|  state
+=*  sta  -
+^-  agent:gall
+|_  =bowl:gall
++*  this  .
+    def   ~(. (default-agent this %|) bowl)
+::
+++  on-init
+  ^-  (quip card _this)
+  `this(local.sta 1)
+::
+++  on-save
+  !>(sta)
+::
+++  on-load
+  |=  old=vase
+  ^-  (quip card _this)
+  `this(sta !<(state old))
+::
+++  on-poke
+  |=  [=mark =vase]
+  ^-  (quip card _this)
+  ?>  =(mark %example-gall-action)
+  =/  act=example-gall-action  !<(example-gall-action vase)
+  ?-  act
+      ::
+    [%increment ~]
+  :-  [%give %fact `/local %atom !>(+(local.sta))]~
+  this(local.sta +(local.sta))
+      ::
+    [%increment-remote who=@p]
+  :_  this
+  :~  :*
+    %pass
+    /inc/(scot %p who.act)
+    %agent
+    [who.act %example-gall]
+    %poke  %example-gall-action  !>([%increment ~])
+  ==  ==
+      ::
+    [%view who=@p]
+  :_  this
+  [%pass /view/(scot %p who.act) %agent [who.act %example-gall] %watch /local]~
+      ::
+    [%stop-view who=@p]
+  :_  this(remote.sta (~(del by remote.sta) who.act))
+  [%pass /view/(scot %p who.act) %agent [who.act %example-gall] %leave ~]~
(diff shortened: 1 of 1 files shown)


reference/gall.md
```diff
@@ -0,0 +1,765 @@
...
++$  card  card:agent:gall
++$  note
+  $%
+    [%arvo =note-arvo]
+    [%agent [=ship name=term] =task:agent:gall]
+  ==
++$  state  [local=@ud remote=(map @p @ud)]
+--
+=|  state
+=*  sta  -
```
Fang-: Might be better practice to do versioned state here. Also makes it more obvious how upgrading can happen in the `+on-load` docs below. Then:

```hoon
+$  state-0  [%0 local=@ud remote=(map @p @ud)]
--
=|  state-0
=*  state  -
```

That way you can avoid the three-letter-word. Also, below, you're free to do just `local` instead of `local.sta`, since the `=|` here already sticks it right in your subject, and the `=*` is just useful for if you need to refer to the state in its entirety. (Though "don't call out `.state` explicitly" is more of a personal preference than a guideline. Wonder if anyone hard-disagrees though.)

reference/gall.md
```diff
@@ -0,0 +1,765 @@
...
+  ?>  =(mark %example-gall-action)
+  =/  act=example-gall-action  !<(example-gall-action vase)
+  ?-  act
+      ::
+    [%increment ~]
+  :-  [%give %fact `/local %atom !>(+(local.sta))]~
+  this(local.sta +(local.sta))
+      ::
+    [%increment-remote who=@p]
+  :_  this
```
Fang-: Indentation here (and for other `?-` and `?+`) should look like:

```hoon
  ?-  act
      [%increment ~]
    :-  [%give %fact `/local %atom !>(+(local.sta))]~
    this(local.sta +(local.sta))
  ::
      [%increment-remote who=@p]
    :_  this
    etc
```

reference/gall.md
```diff
@@ -0,0 +1,765 @@
...
+  |=  =path
+  ^-  (unit (unit cage))
+  ?+  path  [~ ~]
+      ::
+    [%x %local ~]
+  ``[%atom !>(local.sta)]
+      ::
+    [%x %remote who=@ta ~]
+  ~&  remote.sta
+  =/  res  (~(got by remote.sta) (slav %p -.+.+.path))
```
Fang-: Path part access generally happens as `i.t.t.path`. It isn't much better, but at least it's faces. You might also want to `=*  location  i.t.t.path` on the line before, for clarity.

reference/gall.md
```diff
@@ -0,0 +1,765 @@
...
+  ~&  res
+  ``[%atom !>(res)]
+      ::
+    [%y %remote ~]
+  =/  dir=(map @ta ~)
+    (molt (turn ~(tap by remote.sta) |=([who=@p *] [(scot %p who) ~])))
```
Fang-: I imagine this tripping people up. Tall form might be better.

reference/gall.md
```diff
@@ -0,0 +1,765 @@
...
+(unit (unit cage))
+```
+
+If this arm produces `[~ ~ data]`, then `data` is the value at the the
+given path.  If it produces `[~ ~]`, then there is no data at the given
+path and never will be.  If it produces `~`, then we don't know yet whether
+there is or will be data at the given path. The head of the path is known as the
+`care`. Requests with a care of `%x` should return a vase that matches the mark
+at the end of the path. Requests with a care of `%y` should return a vase with
+an `arch` inside.
```
Fang-: Perhaps also worth noting is that, while the mark that's being asked for isn't included in the `path` that you're given, your output `cage` needs to match it (or be convertible to it?) in order for it to not crash on the `.^` end.

reference/gall.md
```diff
@@ -0,0 +1,765 @@
...
+  then the poke failed, and a printable explanation (eg a stack trace)
+  is given in the `tang`.
+
+- `%watch-ack`: acknowledgment (positive or negative) of a subscription.
+  If negative, the subscription is already ended (technically, it never
+  started).
+
+- `%fact`: update from the publisher.
+
+- `%kick`: notification that the subscription has ended.
```
Fang-: Worth noting that this happens either because of the target app killing the subscription, or ames doing it due to pressure. Most of the time, you want to try resubscribing. If you were kicked, and it's no longer valid to access, you'll get a negative `%watch-ack` and end your flow there.

rmariani:
So happy to see this, Liam. This will be a welcome addition to our forthcoming suite of Arvo reference docs

PR #760, opened 2019-12-13
add key:by and val:by to map reference

liam-fitzgerald:
adds key:by and val:by to map reference documentation.

Have used `++` despite it being inconsistent with the rest of the documentation, because `+-` is no longer used.

### DIFF
--- a/reference/library/2i.md
+++ b/reference/library/2i.md
@@ -610,6 +610,46 @@ A map.
 ```
 
 ---
+
+### `++  key:by`
+
+Set of keys
+
+Produces a the set of all keys in map `a` that is prepended to set `b`,
+which is empty by default.
+
+#### Accepts
+
+`a` is a map.
+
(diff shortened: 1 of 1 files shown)


reference/library/2i.md
```diff
@@ -610,6 +610,46 @@ A map.
 ```
 
 ---
+
+### `++  key:by`
+
+Set of keys
+
+Produces a the set of all keys in map `a` that is prepended to set `b`,
```
joemfb: "Produces a the"

reference/library/2i.md
```diff
@@ -1056,4 +1096,43 @@ An atom.
...
 
----
\ No newline at end of file
+---
+
+### `++  val:by`
+
+List of values
+
+Produces a the list of all values in map `a` that is prepended to list `b`,
```
joemfb: And here too

PR #763, opened 2019-12-14
Modernize reel, roll, sort examples

Fang-:
These were using some weird construction to make a wrapper gate, probably for long-gone compatibility reasons. This just passes in the 

Also does a typecast instead of `limo`. Plenty of other examples seem to do so, and this is more common in recently-written hoon.


ohAitch:
Should they maybe use ++ly?

PR #768, opened 2019-12-18
More correct chat glossary entry

Fang-:
This was all kinds of messed up. I don't even know why we have glossary entries for hall and talk. They say they're dead, so why bother describing them?

### DIFF
--- a/glossary/chat.md
+++ b/glossary/chat.md
@@ -1,24 +1,24 @@
 +++
-title = "chat-cli"
+title = "chat"
 
 template = "doc.html"
 [extra]
 category = "arvo"
 +++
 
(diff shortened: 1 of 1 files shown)


syzygyzer:
I think I know why we have entries for Hall and Talk:

I wrote the entries Hall and Talk during my first month at Tlon as a part time contractor. At the time I was just performing whatever tasks were directly handed to me without questioning them too much, and I remember being told that Hall/Talk either are or soon would be obsolete and so assumed that I was being asked to document them anyways for posterity, rather than question whether we should really have them. So I think that is how that got through - a combination of an out-out-date Asana task along and an inexperienced employee.

I've been here long enough now though to have the opinion that those pages are out of place though and should be removed.

PR #769, opened 2019-12-18
Corrects the description of Jael

syzygyzer:
I checked a few potential places and didn't see that we explain exactly what life number and breach number are anywhere in the glossary (or even the docs for that matter). This doesn't seem like the right place to hide those definitions, but should probably be mentioned in the docs somewhere and linked to from here. I'll make an issue for it.


syzygyzer:
#770

PR #779, opened 2020-01-09
added new hoon terms to the glossary

baudtack:


matildepark:
You wouldn’t mind adding these to glossary.js in the static folder on urbit.org, would you?

rmariani:
I think that the "analogy to conventional programming concepts" part should come first, at least for gate, since the purpose of this is to demystify our jargon via hyperlinks in the tutorial. Then we can say "more specifically..." or "more accurately..."

baudtack:
> I think that the analogy to conventional programming concepts part should come first

Strong disagree. The actual definition should come first as long as it's short enough.

rmariani:
The purpose of this initiative is to demystify our strange jargon in the tutorial by hyperlinking it to a quick reference. Restating the more "academic" definition won't help the people we're trying to help

baudtack:
``` 
Term - definition

explanation/analogy
```

is a clearer format imho. If the definition is short enough it will still come up fine in a tool tip but the

rmariani:
Alright, that's fine. As long as the reader gets to the analogy very quickly

PR #785, opened 2020-01-27
Linked all hoon related glossary entries in tutorials

baudtack:


rmariani:
I feel like pluralization should be included in the hypertext to glossary entries. That's pretty standard everywhere I've seen

baudtack:
Having worked on this today, I'm concerned about linking every occurrence of these terms creates a lot of noise that makes editing difficult.

Fang-:
As mentioned to @baudtack out of band, this does indeed seem rather noisy. A wikipedia-style approach of linking just the first occurrence on the page seems a lot more sensible.

rmariani:
We discussed this an we were on the fence, but we want to take advantage of tooltips when they encounter a term again. Spaced repetition. People being mystified by our jargon is a recurring problem in Hoon School, and a lot of these are large documents. I could be convinced either way, but I will leave it up to David

Fang-:
> we want to take advantage of tooltips when they encounter a term again. Spaced repetition.

Very good! Always loved those specialist technical forums that auto-augmented abbreviations and such with explanation tooltips. But unless we implement some convoluted on-hover page summary fetching, links aren't tooltips, and their usability is much worse as a result. (Getting out of the flow of the page to read the thing, etc.)

PR #788, opened 2020-01-29
add reference documentation for gut:by

liam-fitzgerald:


reference/library/2i.md
```diff
@@ -503,6 +503,49 @@ A noun.
     ! exit
 ```
 
+---
+### `+-gut:by`
+
+Grab value with default
+
+Produce the value located at key `b` withing map `a`. Use default value `c` if
```
rmariani: typo here. Should be `within`

Issue #749, opened 2019-12-05
!< rune


matildepark:
> A new rune, !<, has been introduced. It takes a mold and a vase, dynamically checks that the type in the vase matches the mold, and produces the value typed with that type.

Should be added to runes.js on urbit.org repo, too.

PR #792, opened 2020-01-31
added !< entry

syzygyzer:
Addresses #749

I noticed while editing this page that some of the marks like `%zpgr` don't
match what is in `hoon.hoon`. This sort of thing keeps coming up and needs a
overall strategy, and is nested inside of some larger docs issues. I'll make an
issue about this. In this case, I just went with what is listed in `hoon.hoon`
instead of what it "ought" to be.

----

#

### DIFF
--- a/reference/hoon-expressions/rune/zap.md
+++ b/reference/hoon-expressions/rune/zap.md
@@ -60,14 +60,16 @@ a typed value from a `vase`.
 ##### Examples
 
 ```
-> !<  @p  !>  ~zod
-~zod
-
 > !<  @  !>  ~zod
 0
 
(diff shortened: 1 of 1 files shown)


syzygyzer:
Oh yeah, all 3 examples utilize !> to make a vase. It does well to emphasize how !< is a kind of partial inverse, but are there any other essentially different ways to make vases to make the examples more colorful?

belisarius222:
You can call +slop or +slam on other vases to produce a new vase. You can also run code against a vase with +slap, and you can grab the vase of a subtree within a vase with +slot.

—
~rovnys-ricfer
https://urbit.org

PR #798, opened 2020-02-06
replaces out of date ames short technical overview

syzygyzer:
It was previously way out of date. I kept the introduction and scrapped the
rest, and then put in the comment at the top of ames.hoon in its stead with some
minor stylistic changes. This is up to date according to @belisarius222

Last time we did something like this for Gall we ended up sitting on the PR for
weeks. I think we should be more aggressive in updating when this sort of
opportunity arises.

----

#


belisarius222:
In addition to Clay and Gall, Jael also uses Ames (for now, at least).

belisarius222:
It's still true that apps use Gall and Clay, but Jael also uses Ames, independently.

belisarius222:
Also, this Ames commentary looks out of date.  I think you have an old master branch, @syzygyzer.

PR #804, opened 2020-02-12
move trace: scrollable, expanded table formatting

matildepark:


matildepark:
I'm still not really into how it looks and feels — it's not immediately apparent it's scrollable and the actual readability is weird. It's a new sort of documentation component — and I wonder if it would be better as a diagram or graphic?

syzygyzer:
I agree. Since I was building a diagram for this anyways, I am going to see what I can do about making things detailed enough to scrap the table.

syzygyzer:
The diagram is shaping up well enough that the table will be superfluous, so I'm closing this.

PR #805, opened 2020-02-13
Aqua doc

baudtack:
First run at an aqua doc mostly based on @philipcmonk 's original documentation. Feels like this is good enough for a first pass until we document all the available aqua functions.


rmariani:
Looks good. Gonna merge, we can augment it later

PR #807, opened 2020-02-18
information about life and rift number

syzygyzer:
Addresses #770

Currently we do not say anywhere what life and rift number are. I have added a
description in tutorials/guide-to-breaches.md defining these concepts, and added
a new glossary page on breaches that has a bidirectional link with the glossary
page on Ames that defines these terms as well.

I am a little foggy as to where this terminology really comes from - Ames,
Jael, or Azimuth. I emphasized the connection with Ames in these edits, but
could easily be convinced that I should talk more about Jael or Azimuth.

Thus I set `category = "arvo"` but that could be changed.

----

#


rmariani:
A most welcome addition

PR #809, opened 2020-02-20
bucgal and bucgar runes

syzygyzer:
Partially addresses #687

I am not sure that I use the word type/mold/structure correctly here so I need a
good eye on that.

I'm also not sure whether `$>` and `$<` can be used outside of getting types
from a `$%`. This is how it is used in all the code I looked at, but I didn't
comb the entire codebase for every instance of `$>`.

----

#

### DIFF
--- a/reference/hoon-expressions/rune/buc.md
+++ b/reference/hoon-expressions/rune/buc.md
@@ -175,18 +175,27 @@ requirement given by `q`.
 ##### Examples
 
 ```
-> =foo $%([%bar p=@ud q=@ud] [%baz p=@ud])
-> =m $<(%bar foo)
-> (m [%bar 2 4])
+~zod:dojo> =foo $%([%bar p=@ud q=@ud] [%baz p=@ud])
+
+~zod:dojo> =m $<(%bar foo)
+
+~zod:dojo> (m [%bar 2 4])
 ford: %ride failed to execute:
-> (m [%baz 2])
+
+~zod:dojo> (m [%baz 2])
 [%baz p=2]
+
+~zod:dojo> ;;($<(%foo [@tas *]) [%foo 1])
+ford: %ride failed to execute:
+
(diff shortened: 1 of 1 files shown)


joemfb:
There's no requirement that the original structure be a `$%`. But that's the main use-case as restrictions are applied to the head of a cell.

Some examples:

```
> ;;([@tas *] [%foo 1])
[%foo 1]
> ;;([@tas *] [%bar 1])
[%bar 1]
```

```
> ;;($>(%foo [@tas *]) [%foo 1])
[%foo 1]
> ;;($>(%foo [@tas *]) [%bar 1])
ford: %ride failed to execute:
```

```
> ;;($<(%foo [@tas *]) [%foo 1])
ford: %ride failed to execute:
> ;;($<(%foo [@tas *]) [%bar 1])
[%bar 1]
```

reference/hoon-expressions/rune/buc.md
```diff
@@ -157,6 +157,59 @@ Irregular (structure mode): `[a b c]` is `$:(a b c)`.
 [p=0 q=%$]
 ```
 
+
+### $< "bucgal"
+
+`[%bsld p=spec q=spec]`: Filters a pre-existing mold to obtain a mold 
+that excludes a particular structure.
```
philipcmonk: "pre-existing" suggests some kind of restriction on the mold, maybe better to say "Filters a mold to obtain a new mold..."

reference/hoon-expressions/rune/buc.md
```diff
@@ -157,6 +157,59 @@ Irregular (structure mode): `[a b c]` is `$:(a b c)`.
...
+
+##### Syntax
+
+Regular:  **2-fixed**.
+
+##### Discussion
+
+This can be used to obtain type(s) from a list of types `p` that do not satisfy a
+requirement given by `q`.
```
philipcmonk: I think `p` and `q` are backward.

PR #810, opened 2020-02-21
page on common vane tasks

syzygyzer:
In preparation for documenting the public interface of each vane, I have made a
page which discusses each of the tasks in `+vane-task` in `zuse`, which are
tasks that are common to most vanes.

Things I am unsure about:

 - Should I add more links? I haven't written much for the reference section yet
   and haven't given any thought as to whether it should be quite as link-festooned as the tutorial section.
 - This sort of feels like it could be tutorial material. The documents I'm
   working on that reference this one are solidly reference material, but I
   wonder if this page lacks that "just the facts" feel that reference material
   usually has, and that it lacks concrete examples. On the other hand, this
   might just be inevitable for material about a generic interface that is
   ignoring implementation details by necessity. The examples feel like they
   need to be in each specific vane API page.

----

#


joemfb:
It seems like this could go one of two ways, either folding this information into vane-specific documents, or filling out a document that explains the nature and obligations of any vane such. The latter would be very useful, but basically innovative. The former seems much easier now.

As it stands, I'm not sure it's particularly helpful to document the shared nature of these tasks. It's more than just a minor implementation detail, but it's not obvious exactly how much more. My $0.02...

syzygyzer:
Following a discussion we just had in the docs meeting, what I'm going to do with this now is think a bit more about where this could be integrated in the tutorials section. I'm working on vane-specific API documentation now and I just end up repeating a lot of the information here, which IMO should generally be avoided in the reference section.

syzygyzer:
Whhops, did things in the wrong order. I'm gonna use this branch to better segregate what I'm doing with the vane APIs (my original intention) and move some edited version of the content originally here somewhere to the tutorial, but should have closed this PR first.

PR #817, opened 2020-03-12
Behn public API

syzygyzer:
This PR adds a new page, `reference/vane-apis/behn.md`, that lists each of the
`task`s that Behn performs, what they do, and what `gift`s they may return in
response.

Some possible issues:
i) I include the source code for each arm. Since we currently do not have a way
to pull source automatically into the docs, this creates an issue where it could be
out of date quickly. I think I will probably remove it, but would like feedback
on that choice first. Furthermore, sometimes what each arm does is split into
several places (and this is especially true for other vanes), so showing the
source for just the arm can be deceptive.
ii) I am not yet an expert on reading Hoon, so I cannot be sure that I've
covered all possible `gift` responses correctly. I'm pretty sure I have for
Behn, but not as much for e.g. Ames.
iii) For returns, I only talk about what `gift`s may be returned, which is not
the same thing as what the arms return. This is not what is done in
the Gall API docs, which instead focuses on what the arms return. A more
complete documentation would include both.
iv) I don't include any examples. After some time messing around with a simple
goad-like app to pass vanes `card`s to get very artificial examples, I've come
to think that it would be better for engineers to provide the examples, since
they have a much better feel for how these things are actually used in practice
than I do. Coming up with examples myself I think would eat up a lot more
manhours than just having engineers write the examples themselves, since what
I'd end up doing is just talking to engineers to get the examples anyways. Best
to cut out the middleman.

----

#

### DIFF
--- a/reference/vane-apis/behn.md
+++ b/reference/vane-apis/behn.md
@@ -60,16 +60,9 @@ If the set of timers is nonempty when Behn is `%pass`ed a `%crud` `task`, Behn
 
 `%drip` allows one to delay `gift`s until a timer set for `now` fires.
 
-For example, say an app is subscribed to updates from Clay. If Clay `%give`s
-updates to the app directly and the app crashes, this may cause Clay to crash as
-well. If instead Clay `%pass`es Behn a `%drip` `task` wrapping the update
-`gift`, Behn will set a timer for `now` that, when fired, will cause the update
-`gift` to be given. If it causes a crash then it will have been in response to
-the `%drip` move, thereby isolating Clay from the crash. Thus `%drip` acts as a sort of buffer against cascading
-sequences of crashes.
-
 `%drip` only handles `gift`s, and can only schedule `gift`s for as soon as
 possible after the prescribed timer fires.
+
 
 #### Accepts
 
@@ -82,10 +75,22 @@ mov=vase
 #### Returns
 
 In response to a `%drip` `task`, Behn will `%give` a `%meta` `gift` containing
-the `gift` originally `%give`n to Behn when `%drip` was first called.
+the `gift` originally `%give`n to Behn when `%drip` was first called. That makes
+Behn its own client for `%drip`.
 
-Is that really correct? It looks like the gift returned is actually a response to
-a pass it passed itself, not the original event?
+#### Example
+
+
+Client -- %drip --> Behn -- %wait --> Behn -- %wake --> Behn -- %give --> Target
+
+
+Say an app is subscribed to updates from Clay. If Clay `%give`s
+updates to the app directly and the app crashes, this may cause Clay to crash as
+well. If instead Clay `%pass`es Behn a `%drip` `task` wrapping the update
+`gift`, Behn will set a timer for `now` that, when fired, will cause the update
+`gift` to be given. If it causes a crash then it will have been in response to
+the `%drip` move, thereby isolating Clay from the crash. Thus `%drip` acts as a sort of buffer against cascading
+sequences of crashes.
 
 
 
@@ -142,8 +147,7 @@ This `task` has no arguments.
 
 #### Returns
 
-This `task` returns `[moves state]`.
-
+This task returns nothing.
 
 ### `%vega`
 
@@ -156,7 +160,7 @@ This `task` has no arguments.
 
 #### Returns
 
-This `task` returns `[moves state]`.
+This `task` returns nothing.
 
 
 
--- a/reference/vane-apis/behn.md
+++ b/reference/vane-apis/behn.md
@@ -66,13 +66,13 @@ the stack trace.
 
 `%drip` allows one to delay a `gift` until a timer set for `now` fires.
 
-A Client gives Behn a `%drip` task wrapping a `gift` to be `give`n to a Target.
+A Client `%slip`s Behn a `%drip` task wrapping a `gift` to be `give`n to a Target.
 This launches a sequence of `move`s as written here:
 
 ```
-Client -- %drip --> Behn -- %wait --> Behn -- %wake --> Behn -- %meta --> Target
+Client -- %slip %drip --> Behn -- %pass %wait --> Behn -- %give %wake --> Behn -- %give %meta --> Target
 ```
-Here the `%meta` move is a wrapper for the `%gift` inside of the initial `%drip` wrapper.
+Here the `%meta` `move` is a wrapper for the `%gift` inside of the initial `%drip` wrapper.
 
 `%drip` only handles `gift`s, and can only schedule `gift`s for as soon as
 possible after the prescribed timer fires.
@@ -88,7 +88,8 @@ mov=vase
 
 #### Returns
 
-In response to a `%drip` `task`, Behn will `%give` a `%meta` `gift` containing
+In response to a `%drip` `task`, Behn will `%pass` a `%wait` to itself, which
+then triggers a `%wake` `gift` to itself, which then causes Behn to `%give` a `%meta` `gift` containing
 the `gift` originally `%give`n to Behn when `%drip` was first called. That makes
 Behn its own client for `%drip`.
 
--- a/reference/vane-apis/behn.md
+++ b/reference/vane-apis/behn.md
@@ -1,0 +1,366 @@
++++
+title = "Behn"
+weight = 2
+template = "doc.html"
++++
+
+# Behn
+
+In this document we describe the public interface for Behn. Namely, we describe
+each `task` that Behn can be `pass`ed, and which `gift`(s) Behn can `give` in return.
+
+
+## Tasks
+
+### `%born`
+
+Each time you start your Urbit, the Arvo kernel calls the `%born` task for Behn. When
+called, Behn gets the current time from Unix and updates its list of timers
+accordingly.
+
+#### Accepts
+
+```hoon
+[~]
+```
+
+#### Returns
+
+`%born` does not return a `gift`.
+
+#### Source
+```hoon
+::  +born: urbit restarted; refresh :next-wake and store wakeup timer duct
+  ::
+  ++  born  set-unix-wake(next-wake.state ~, unix-duct.state duct)
+```
+
+
+### `%crud`
+
+`%crud` is called whenever an error involving Behn occurs, which for now means
+that Behn has no timers in its state.
+
+When called, it checks to see if the set of timers is empty. If so, it prints
+and error report containing `error`, otherwise it passes the error to `+wake`.
+
+Behn should be the first vane to be activated in a `%wake`. Occasionally this does not
+happen, but we do not yet handle this error.
+
+#### Accepts
+
+```hoon
+[tag=@tas error=tang]
+```
+Here, `tag` is a `@tas` denoting the type of error, and `error` is an error message.
+
+#### Returns
+
+If the set of timers is nonempty when Behn is `%pass`ed a `%crud` `task`, Behn
+`%give`s a `%wake` `card` to itself.
+
+```hoon
+[%wake %behn-crud-no-timer tag error]
+```
+ 
+
+#### Source
+```hoon
+++  crud
+    |=  [tag=@tas error=tang]
+    ^+  [moves state]
+    ::  behn must get activated before other vanes in a %wake
+    ::
+    ::    TODO: uncomment this case after switching %crud tags
+    ::
+    ::    We don't know how to handle other errors, so relay them to %dill
+    ::    to be printed and don't treat them as timer failures.
+    ::
+    ::  ?.  =(%wake tag)
+    ::    ~&  %behn-crud-not-first-activation^tag
+    ::    [[duct %slip %d %flog %crud tag error]~ state]
+    ::
+    ?:  =(~ timers.state)  ~|  %behn-crud-no-timer^tag^error  !!
+    ::
+    (wake `error)
+ ```
+ 
+
+### `%drip`
+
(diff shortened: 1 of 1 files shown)


syzygyzer:
Oh yeah, lastly, the reason this is all a single commit is because Magit started
giving me some weird error when I was trying to make a PR with the original
branch `poprox/behn-public-interface`. I don't think its a big deal with
documentation to squash the git history of the initial page into a single
commit, but that's why there's only one commit here.

reference/vane-apis/behn.md
```diff
@@ -0,0 +1,366 @@
...
+::  +born: urbit restarted; refresh :next-wake and store wakeup timer duct
+  ::
+  ++  born  set-unix-wake(next-wake.state ~, unix-duct.state duct)
+```
+
+
+### `%crud`
+
+`%crud` is called whenever an error involving Behn occurs, which for now means
+that Behn has no timers in its state.
```
philipcmonk: The most common error is actually that whoever Behn woke up crashed.  In this case, Behn gets the `%crud`, which it passes along to the person Behn tried to wake up as a `%wake` with a non-null `error`.

reference/vane-apis/behn.md
```diff
@@ -0,0 +1,366 @@
...
+
+When called, it checks to see if the set of timers is empty. If so, it prints
+and error report containing `error`, otherwise it passes the error to `+wake`.
+
+Behn should be the first vane to be activated in a `%wake`. Occasionally this does not
+happen, but we do not yet handle this error.
```
philipcmonk: When does this not happen?  Pretty sure it always happens if I understand what you're saying.

reference/vane-apis/behn.md
```diff
@@ -0,0 +1,366 @@
...
+ 
+
+### `%drip`
+
+`%drip` is utilized to delay `%give`ing a `gift`.
+
+`%drip` allows one to delay `gift`s until a given condition is met. For example,
+if Clay crashes and you do not wish for an app to then crash when it tries to
+access Clay before it is restarted, `%drip` can delay that access until Clay has
+been restored.
```
philipcmonk: There is no condition.  All it does is delay giving the gift until a later event (by setting a timer for `now` and giving the gift once that timer fires).  For example, this makes it so that if an app is subscribed to Clay updates, if the app crashes while processing an update, it doesn't crash Clay (since the update came as a `%drip`, so it came in a separate event).

reference/vane-apis/behn.md
```diff
@@ -0,0 +1,366 @@
...
+
+#### Accepts
+
+```hoon
+@da
+```
+ 
+#### Returns
+
+This `task` returns nothing.
```
philipcmonk: I'd say this returns a `%wake`, just not immediately.

reference/vane-apis/behn.md
```diff
@@ -0,0 +1,366 @@
...
+
+```hoon
+[tag=@tas error=tang]
+```
+Here, `tag` is a `@tas` denoting the type of error, and `error` is an error message.
+
+#### Returns
+
+If the set of timers is nonempty when Behn is `%pass`ed a `%crud` `task`, Behn
+`%give`s a `%wake` `card` to itself.
```
philipcmonk: Not to itself, to the next client in the set of timers.

reference/vane-apis/behn.md
```diff
@@ -0,0 +1,366 @@
...
+
+`%drip` takes in a `%give` `move` in a `vase`.
+
+#### Returns
+
+In response to a `%drip` `task`, Behn will `%give` a `%meta` `gift` containing
+the `gift` originally `%give`n to Behn when `%drip` was first called.
+
+Is that really correct? It looks like the gift returned is actually a response to
+a pass it passed itself, not the original event?
```
philipcmonk: It is correct.  Behn becomes its own client for `%drip`.  Basically,

```
Client -- %drip --> Behn -- %wait --> Behn -- %wake --> Behn -- %give --> Target
```

Actually, the "Client" and "Target" are the same from Behn's perspective.  The client should call this with `%slip` instead of `%pass`, which lets the client impersonate the target.

reference/vane-apis/behn.md
```diff
@@ -0,0 +1,366 @@
...
+This `task` is sent by the interpreter in order to free up memory.
+ This `task` is empty for Behn, since it is not a good idea to forget your timers.
+
+#### Accepts
+
+This `task` has no arguments. 
+
+#### Returns
+
+This `task` returns `[moves state]`.
```
philipcmonk: `[moves state]` without adding anything to `moves` is just a no-op.  It doesn't return anything.

reference/vane-apis/behn.md
```diff
@@ -0,0 +1,366 @@
...
+This `task` informs the vane that the kernel has been upgraded. Behn does not do
+anything in response to this.
+
+#### Accepts
+
+This `task` has no arguments.
+
+#### Returns
+
+This `task` returns `[moves state]`.
```
philipcmonk: Same as above; doesn't return anything.

reference/vane-apis/behn.md
syzygyzer: I need a little more clarity. Does this sound right? Behn attempts to wake up Ames. Ames has crashed, so it passes Behn a `%crud` task, and Behn then responds by giving Ames a `%wake` gift with an error message?

I considered a few other possibilities of what you meant but this seemed like the simplest one (though maybe incorrect).

reference/vane-apis/behn.md
syzygyzer: I wrote this based on this comment in  `+crud`:

```
::  behn must get activated before other vanes in a %wake
    ::
    ::    TODO: uncomment this case after switching %crud tags
    ::
    ::    We don't know how to handle other errors, so relay them to %dill
    ::    to be printed and don't treat them as timer failures.
    ::
    ::  ?.  =(%wake tag)
    ::    ~&  %behn-crud-not-first-activation^tag
    ::    [[duct %slip %d %flog %crud tag error]~ state]
    ::
```

Though since I wrote this, it has been updated, and now does check for this kind of error:

```
::  behn must get activated before other vanes in a %wake
    ::
    ?.  =(%wake tag)
      ~&  %behn-crud-not-wake^tag
      [[duct %slip %d %flog %crud tag error]~ state]
    ::
```

So it looks like what I wrote is checked for now, unless I'm misunderstanding something here.

reference/vane-apis/behn.md
syzygyzer: I'm not sure I fully understand, how does this look? https://github.com/urbit/docs/pull/817/commits/df742bdb92325ed640e1be3d0633d59c68513003

reference/vane-apis/behn.md
syzygyzer: Thanks, I will include this information. Where does giving a `%meta` gift to itself show up in this sequence?

reference/vane-apis/behn.md
philipcmonk: Behn attempts to wake up Ames, but Ames crashes.  This whole event (not just move) is thrown away -- nobody can react to a crash directly.  However, vere happens to notice that the event didn't complete, so it injects a *new* event that says "you were going to try to process an event, but it failed, and here's a stack trace in case you want to handle that (eg print it out)".

reference/vane-apis/behn.md
philipcmonk: Looks like it does check for this error, yeah.  It shouldn't ever happen.

reference/vane-apis/behn.md
philipcmonk: Looks good

reference/vane-apis/behn.md
philipcmonk: The last `%give --> Target` is a `%meta` give, since it doesn't know what type of gift the client wants to ultimately send to the target.

reference/vane-apis/behn.md
syzygyzer: How does this look? https://github.com/urbit/docs/pull/817/commits/f268d6d8752e037364587e59656478f8177c309d

reference/vane-apis/behn.md
syzygyzer: I split this up between a few commits so I'll just paste the revised drip entry here:
 

### `%drip`

`%drip` allows one to delay a `gift` until a timer set for `now` fires.

A Client gives Behn a `%drip` task wrapping a `gift` to be `give`n to a Target.
This launches a sequence of `move`s as written here:

```
Client -- %drip --> Behn -- %wait --> Behn -- %wake --> Behn -- %meta --> Target
```
Here the `%meta` move is a wrapper for the `%gift` inside of the initial `%drip` wrapper.

`%drip` only handles `gift`s, and can only schedule `gift`s for as soon as
possible after the prescribed timer fires.


#### Accepts

```hoon
mov=vase
```

`%drip` takes in a `%give` `move` in a `vase`.

#### Returns

In response to a `%drip` `task`, Behn will `%give` a `%meta` `gift` containing
the `gift` originally `%give`n to Behn when `%drip` was first called. That makes
Behn its own client for `%drip`.

#### Example


Say an app (the Target) is subscribed to updates from Clay (the Client). If Clay `%give`s
updates to the app directly and the app crashes, this may cause Clay to crash as
well. If instead Clay `%pass`es Behn a `%drip` `task` wrapping the update
`gift`, Behn will set a timer for `now` that, when fired, will cause the update
`gift` to be given. If it causes a crash then it will have been in response to
the `%drip` move, thereby isolating Clay from the crash. Thus `%drip` acts as a sort of buffer against cascading
sequences of crashes.

reference/vane-apis/behn.md
philipcmonk: Sure, this works.

reference/vane-apis/behn.md
philipcmonk: - The initial client's call to Behn is not actually a `%give`, it's more like a `%pass`.  Technically it's a `%slip`, which "`%pass` but impersonate this duct", which is a hack.  I would use the word "pass", but maybe not "`%pass`"

- The `%meta` that goes to the target vane is a `[%give %meta ...]`, so I would still write it as a `%give` in the little diagram.

syzygyzer: Thanks, I'll poke around and try to figure out what %slip is, since we should definitely mention that it is not an ordinary %pass.

I wrote %meta instead of %give since the other labels in the sequence say which card it is instead of the type of card (type probably isn't the right word) and I thought it should be consistent. What I should actually do is make a small diagram in the style of the move trace diagram that has all of the relevant information.

syzygyzer: I'm still confused about which part is the `%slip`. Does the client `%slip` a `%drip` to Behn wrapping a `%gift`, or does it `%pass` Behn a `%drip` wrapping a `%slip`?

philipcmonk: Oh you're right, `%meta` is correct in the diagram.

`%slip` is a sibling of `%pass` and `%give` in +wind.  Technically, `%pass` says "push this wire onto the duct and transfer control to this other vane", `%give` says "pop a wire off the duct and transfer control back to whoever pushed the wire on", and `%slip` says "transfer control to this other vane but don't push or pop a wire".  This means a `%give` in response to a `%slip` will go directly back to the caller of the vane that sent the `%slip`.  This is super hacky and shouldn't really exist.

ohAitch:
So it's just a tail call?

philipcmonk:
Yes, except that the response comes back from the other vane.  So, you get the situation where you pass a note to one vane and get a gift back from another.

reference/vane-apis/behn.md
syzygyzer: Since this doc is the wrong place to discuss slip, I decided to just say that `%drip` uses `%slip` and made an issue for me to document `%slip` elsewhere. I started on a diagram yesterday, but didn't come up with a satisfactory way to represent `%slip`s yet so I'm just gonna stick with the text diagram for now.

How's this look?


### `%drip`

`%drip` allows one to delay a `gift` until a timer set for `now` fires.

A Client `%slip`s Behn a `%drip` task wrapping a `gift` to be `give`n to a Target.
This launches a sequence of `move`s as written here:

```
Client -- %slip %drip --> Behn -- %pass %wait --> Behn -- %give %wake --> Behn -- %give %meta --> Target
```
Here the `%meta` `move` is a wrapper for the `%gift` inside of the initial `%drip` wrapper.

`%drip` only handles `gift`s, and can only schedule `gift`s for as soon as
possible after the prescribed timer fires.


#### Accepts

```hoon
mov=vase
```

`%drip` takes in a `%give` `move` in a `vase`.

#### Returns

In response to a `%drip` `task`, Behn will `%pass` a `%wait` to itself, which
then triggers a `%wake` `gift` to itself, which then causes Behn to `%give` a `%meta` `gift` containing
the `gift` originally `%give`n to Behn when `%drip` was first called. That makes
Behn its own client for `%drip`.

#### Example


Say an app (the Target) is subscribed to updates from Clay (the Client). If Clay `%give`s
updates to the app directly and the app crashes, this may cause Clay to crash as
well. If instead Clay `%pass`es Behn a `%drip` `task` wrapping the update
`gift`, Behn will set a timer for `now` that, when fired, will cause the update
`gift` to be given. If it causes a crash then it will have been in response to
the `%drip` move, thereby isolating Clay from the crash. Thus `%drip` acts as a sort of buffer against cascading
sequences of crashes.

philipcmonk: Perfect!

PR #819, opened 2020-03-12
Ames public API

syzygyzer:
This PR adds a new file, `references/vane-apis/ames.md`. This is very much a
work in progress, but I'm at the point where I need people to read over it to
help me out.

Unlike for the Behn page, I do specify some of what the arms return rather than
just the possible `gift`s. Not sure if this is worthy or not, but I'd like to be
consistent.

As with the Behn PR #817, I do not include any examples, and for the same
reason. I think engineers could product much better examples much faster than I
would by talking to engineers and writing them myself.

************

I've had some amount of trouble trying to figure out what `gift`s a given `task`
may return. Here are some specific questions:

I'm not sure on the returns for `%hear` and `%hole`. There are four arms in `ames.hoon`
named `+on-hear`, and I wrote about `on-hear:event-core`. Since it handles raw
packet receipt, I assumed that it would return an ack, but I couldn't find where
this might be happening. I also assumed that since it handles raw packets, it
could be considered to return almost any `gift`. Clarification desired.

I can't say for sure whether `%heed` returns a `gift` or not.

Does `%plea` return a `gift`?

************

I have not included the source of the arms here. There's way too much that
happens outside of the main arm for each task so its not much better than noise
to include it IMO. This is in contrast to the current state of #817, which
currently includes the source of each arm, but as I state there I think I will
remove it.

***********

Style issue: does the `+` go in front of an arm name when we're also specfiying
what core it is in? By that I mean, should I be writing `+on-hear:event-core` or
just `on-hear:event-core`? The former seems better, but I can't recall seeing
anywhere where we write like that.

----

#

### DIFF
--- a/reference/vane-apis/ames.md
+++ b/reference/vane-apis/ames.md
@@ -44,7 +44,7 @@ message (typically an IP address).
 
 `%hear` can trigger a number of possible returns. It can trigger the release of
 zero or more additional packets via `%send` `gift`s. It may also trigger a `%boon`
-or `%memo` `gift` to a local vane in the case of a completed message. 
+or `%plea` `gift` (collectively referred to as a `%memo` within Ames) to a local vane in the case of a completed message. 
 
 
 ### %heed
@@ -93,7 +93,7 @@ message (typically an IP address).
 #### Returns
 
 Like `%hear`, `%hole` can trigger additional packets in the form of `%send`
-`gift`s, and may also yield a `%boon` or `%memo` `gift` to a local vane.
+`gift`s, and may also yield a `%boon` or `%plea` `gift` to a local vane.
 
 ### %jilt
 
--- a/reference/vane-apis/ames.md
+++ b/reference/vane-apis/ames.md
@@ -109,11 +109,7 @@ The `ship` we no longer wish to track.
 
 #### Returns
 
-`+on-jilt` returns `event-core` with the `duct` associated to `ship` removed from `heeds`, assuming it is
-there. Otherwise it returns `event-core` unchanged.
-
-Does `%jilt` return any gifts?
-
+This `task` returns no `gift`s.
 
 ### %plea
 
--- a/reference/vane-apis/ames.md
+++ b/reference/vane-apis/ames.md
@@ -43,17 +43,6 @@ message (typically an IP address).
 
 `%hear` can trigger any number of possible returns...?
 
-#### Source
-
-```hoon
-  ::  +on-hear: handle raw packet receipt
-  ::
-  ++  on-hear
-    |=  [=lane =blob]
-    ^+  event-core
-    (on-hear-packet lane (decode-packet blob) ok=%.y)
-```
-
 
 ### %heed
 
@@ -84,25 +73,6 @@ If the `ship` is indeed being unresponsive, as measured by backed up `%boon`s,
 Ames will `give` a `%clog` `gift` to the requesting vane containing the
 unresponsive peer's urbit address.
 
-#### Source
-
-```hoon
-  ::  +on-heed: handle request to track .ship's responsiveness
-  ::
-  ++  on-heed
-    |=  =ship
-    ^+  event-core
-    =/  ship-state  (~(get by peers.ames-state) ship)
-    ?.  ?=([~ %known *] ship-state)
-      %+  enqueue-alien-todo  ship
-      |=  todos=alien-agenda
-      todos(heeds (~(put in heeds.todos) duct))
-    ::
-    =/  =peer-state  +.u.ship-state
-    =/  =channel     [[our ship] now channel-state -.peer-state]
-    abet:on-heed:(make-peer-core peer-state channel)
-```
-
 
 ### %hole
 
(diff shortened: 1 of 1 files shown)


reference/vane-apis/ames.md
```diff
@@ -0,0 +1,303 @@
...
+[=lane =blob]
+```
+
+`%hear` takes in a `blob`, which is essentially a large atom (around 1kB or less)
+that is the raw data of the message, and a `lane`, which is the origin of the
+message (typically an IP address).
+
+#### Returns
+
+`%hear` can trigger any number of possible returns...?
```
philipcmonk: Yep, depending on many factors (esp. congestion control), it can cause 0 or more additional packets to be released.  It may also trigger a `%memo` or `%boon` to a vane (if it completes a message).

reference/vane-apis/ames.md
```diff
@@ -0,0 +1,303 @@
...
+=ship
+```
+
+The ship to be tracked.
+
+#### Returns
+
+The `+on-heed` arm returns `event-core` with `heeds` modified to include a
+`duct` that points to the input `ship`. (this isn't a `gift` so maybe I should
+not include this?)
```
philipcmonk: Yeah, this is how it stores the subscription internally, but it's not a part of the public api, so I would skip it.

reference/vane-apis/ames.md
```diff
@@ -0,0 +1,303 @@
...
+```hoon
+[=lane =blob]
+```
+`%hole` takes in a `blob`, which is essentially a large atom (around 1kB or less)
+that is the raw data of the message, and a `lane`, which is the origin of the
+message (typically an IP address).
+
+#### Returns
+
+Same confusion as with `%hear`.
```
philipcmonk: Same answer as `%hear`.

reference/vane-apis/ames.md
```diff
@@ -0,0 +1,303 @@
...
+```
+
+The `ship` we no longer wish to track.
+
+#### Returns
+
+`+on-jilt` returns `event-core` with the `duct` associated to `ship` removed from `heeds`, assuming it is
+there. Otherwise it returns `event-core` unchanged.
+
+Does `%jilt` return any gifts?
```
philipcmonk: No, it doesn't

reference/vane-apis/ames.md
```diff
@@ -0,0 +1,303 @@
...
+
+
+### %plea
+
+`%plea` is the `task` used to send messages over Ames. It extends the
+`%pass`/`%give` semantics across the network. As such, it is the most
+fundamental `task` in Ames and the primary reason for its existence.
+
+Ames `pass`es a `%plea` `note` to another vane when it receives a message on a
```
philipcmonk:
```suggestion
Ames also `pass`es a `%plea` `note` to another vane when it receives a message on a
```

One of the confusing things about ames is that `%plea` is both a task and a note, so best to mark the difference.  (Incidentally, I think it would be even more confusing if they had different names)

reference/vane-apis/ames.md
```diff
@@ -0,0 +1,303 @@
...
+"forward flow" from a peer, originally passed from one of the peer's vanes to
+peer's Ames.
+
+Ames `pass`es a `%plea` to itself to trigger a heartbeat message to be sent to
+our ship's sponsor
```
philipcmonk: The heartbeat is now done in app/ping.hoon instead of internally, so this can be omitted.

reference/vane-apis/ames.md
```diff
@@ -0,0 +1,303 @@
...
+Here, `vane` is the destination vane on the remote ship, `path` is the internal
+route on the receiving ship, and `payload` is the semantic message content.
+
+#### Returns
+
+`event-core` is returned, modified to include the received `%plea` (this is not
+a `gift`).
+
+Not sure a `gift` is returned? But if there would be an ack `gift` anywhere, I
+feel like it would be here.
```
philipcmonk: No returns to the caller.  This may trigger more messages to be given to unix, but that depends on congestion control (how many packets are in flight, etc).

reference/vane-apis/ames.md
```diff
@@ -0,0 +1,303 @@
...
+
+#### Returns
+
+```hoon
+    =~  (emit duct %pass /turf %j %turf ~)
+        (emit duct %pass /private-keys %j %private-keys ~)
+```
+
+`%init` returns `event-core` with two new `move`s to be performed that subscribe
+to `%turf` and `%private-keys` in Jael.
```
philipcmonk:
```suggestion
`%init` sends two moves that subscribe to `%turf` and `%private-keys` in Jael.
```

Putting moves in event-core is internal move management, but eventually they just turn into normal moves.

reference/vane-apis/ames.md
```diff
@@ -0,0 +1,303 @@
...
+
+#### Returns
+
+```hoon
+    =.  ships.bug.ames-state  (sy ships)
+    event-core
+```
+
+This `task` returns the `event-core` modified so that debug output is shown only
+for `ships`.
```
philipcmonk: As in other cases, this the state change, which is interesting, but not part of the public interface.  Unless you intend to describe the state changes?

reference/vane-apis/ames.md
```diff
@@ -0,0 +1,303 @@
...
+`%vega` is called whenever the kernel is updated. Ames currently does not do
+anything in response to this.
+
+#### Accepts
+
+`%vega` takes no arguments.
+
+#### Returns
+
+`%vega` returns `event-core`.
```
philipcmonk: This means it's a no-op.

PR #823, opened 2020-03-16
tutorials/hoon/1.1–1.2: text, typographic, and formatting corrections

fakezod:
This commit fixes some text/typo, typographic, capitalisation, formatting, and indentation related issues in the `Hoon` tutorial.


fakezod:
@baudtack Thanks. I’ll apply your standard of using two spaces after a period. Any reason for using `'` and `""` instead of `’` and `“”`? I can change that, if you want, but it would be interesting, why.

baudtack:
>  I’ll apply your standard of using two spaces after a period. 

Apparently, I'm just wrong about this. No idea why I thought that.

> Any reason for using `'` and `""` instead of `’` and `“”` in copy (non-code) text? I can change that, if you want, but it would be interesting, why.

Yeah the reason is that this actually is code. It's markdown code and has to be parsed. In the future we'll also want to run it through other kinds of transformers to produce non html formats and the smart quotes can be a pain in the butt to deal with even if they are typographically nicer.

fakezod:
> Apparently, I'm just wrong about this. No idea why I thought that.

@baudtack Alright, so … no double spaces, only single spaces. Yes?

> Yeah the reason is that this actually is code. It's markdown code and has to be parsed. In the future we'll also want to run it through other kinds of transformers to produce non html formats and the smart quotes can be a pain in the butt to deal with even if they are typographically nicer.

And I’ll remove each occurence of smart quotes. If I got both comments right, then I’ll make the changes, commit, and pull.

baudtack:
Yes. single spaces are fine. If you find any smart quotes, feel free to eradicate them. Also the ellipsis character you mentioned should be changed as you described.

fakezod:
@baudtack Alright, made the changes as desired. Had to add three newly corrected files due to some branch/master issues. Will, starting tomorrow, try to make one branch per edited file and commit that as a pull request. Will then probably be more pull requests but easier/faster to review one pull request per file. Sorry for this one big pull request.

fakezod:
@baudtack Did you have another look at this?

PR #826, opened 2020-03-19
rune: describe =* tistar as macro, not alias

Fang-:
Given the "evaluate when used" nature of =*, it's more accurately described as a
macro than an alias.

This is an attempted first pass at the copy here. Obviously rough around the edges, but kinda highlights already how =* is more clearly described as a macro. "It's calculated every time you use the macro" seems very obvious.

(Just a proposal, taking all challengers.)

### DIFF
--- a/reference/hoon-expressions/rune/tis.md
+++ b/reference/hoon-expressions/rune/tis.md
@@ -423,11 +423,11 @@ Regular: **running**.
 
 ### =* "tistar"
 
-`[%tstr p=term q=hoon r=hoon]`: define an alias.
+`[%tstr p=term q=hoon r=hoon]`: define a macro.
 
 ##### Produces
 
-`r`, compiled with a subject in which `p` is aliased to `q`.
+`r`, compiled with a subject in which `p` is a macro for `q`.
 
 ##### Syntax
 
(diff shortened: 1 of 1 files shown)


baudtack:
I agree alias isn't great here, but I don't see macro as a substantive improvement.

joemfb:
I'm opposed to this; `=*` is about deferred evaluation, macros are usually about source/code transformation.

liam-fitzgerald:
A macro/alias is different to a face (from the users perspective) in that a) it doesn't interrupt the label search and b) it doesn't modify the subject. The word alias doesn't convey a) at all. Indeed, a) is the primary use case for tistar at the moment. Defining it as a macro that resolves to a wing (or whatever else), helps build intuition that it can be used wherever a wing (or whatever else) is used. 

@joemfb in eager languages usually the only means of deferring evaluation are a thunk/zero arg callback or a macro. Whilst it doesn't technically introduce a macro, I'd argue that the pedagogical benefits of calling it a macro outweigh the drawbacks of slightly lying about what it is.

baudtack:
We should change this. There are  many names in the system that should  be changed. But they should only be changed to something that is a clear and substantial improvement. Urbit has had a bad history of just changing the names of things with next to zero benefit. I'm very much for calling this something other than alias and extremely opposed to that thing being macro.

belisarius222:
How about "deferred expression"?  A tistar assigns a name to an expression that will be evaluated in each place the name is dereferenced.

—
~rovnys-ricfer
https://urbit.org

baudtack:
Not... thrilled with it but it would be an acceptable change.

belisarius222:
Keep in mind, the real name for this is a "tistar".  Anything else is just a description.

—
~rovnys-ricfer
https://urbit.org

joemfb:
I'd like to record my continued objection to this change.

> The difference between macroing and pinning ...

mfw

Fang-:
Yeah the merge here seems a tad premature wrt above discussion. At the very least the main paragraph deserves a light rewrite.

I like the "deferred expression" suggestion.

PR #834, opened 2020-03-30
gall-reference: change unit to list in %fact card

liam-fitzgerald:
Changes `(unit path)` to  `(list path)` in the gall docs to match the updated gall cards


liam-fitzgerald:
cc: @rmariani @baudtack

liam-fitzgerald:
bump

PR #836, opened 2020-04-01
Dill public api

syzygyzer:
This is a rough draft of documentation of the Dill public API. It is a work in progress and there are several areas that need work:

I don't understand how `%flow` works. It's a lark expression `+>` that is just the core it
is defined in and thus contains the card and I'm not sure what that gets sent
to. It has the form `[p=@tas q=(list gill:gall)]`

Similarly, `%harm` only appears as `+>` in `+call` in `dill.hoon` and the
associated card is empty. I don't see how this can do anything.

`$hook`, `$talk` appears in task:able:dill in `zuse.hoon` and has the comment "this term
hung up" but hook appears nowhere in `dill.hoon`. is `hook` deprecated?

I'm pretty sure what `%talk` does is print a `tape` to the screen, but following
along what is going on in `+from` is difficult.

The comments say that `%veer` is for installing vanes but I can't see how that
is happening in `dill.hoon` as its another `dump kyz` call, and I also don't
understand why that functionality would be in Dill (related to the fact that
Dill is technically the first vane?)

`%verb` is a task with an empty card that is commented as verbose mode. I assume
this must toggle verbose mode but I don't see how - its another `dump kyz` task.

### DIFF
--- a/reference/vane-apis/dill.md
+++ b/reference/vane-apis/dill.md
@@ -306,7 +306,7 @@ This `task` updates the kernel.
 [p=@t q=@t]
 ```
 
-`p` is related to hoon and `q` is related to Arvo but I don't know how.
+`p` is the contents of the new `hoon.hoon` and `q` is the contents of the new `arvo.hoon`.
 
 #### Returns
 
@@ -340,6 +340,11 @@ This `task` is used to install `zuse` and vanes. It is handled by `+veer` in
 ```hoon
 [p=@ta q=path r=@t]
 ```
+
+`p` is the vane letter, `q` is the formal `path` to the
+vane, and `r` is the source code for the vane. The vane letter is `%` followed
+by the first letter of the vane in lowercase, such as `%d` for Dill. `p=%$` in
+the case of `zuse`.
 
 #### Returns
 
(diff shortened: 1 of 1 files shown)


syzygyzer:
All my questions above have been answered by @philipcmonk, I will update the rest of this tomorrow.

syzygyzer:
This is almost done.

I'm not sure what the input `[p=@t q=@t]`for `%lyra` is. The main code for this
seems to be in `/arvo/lib/hood/helm.hoon` and it looks like `p` and `q`
are paths (or at least they have fas in them) related to Hoon and Arvo respectively, but they're not `path`s so I'm
confused about what these arguments are supposed to be.

I don't understand what the inputs `[p=@ta q=path r=@t]` are for `%veer` either.
This is used for installing `zuse` and vanes.

philipcmonk:
For veer, those inputs are [vane-letter formal-path-to-vane source-code-for-vane] where vane-letter is the first letter of the vane or %$ for zuse.

For lyra, it's [source-code-of-hoon source-code-of-arvo].  The .^(@ %cx ...) lines are scries that load the content of the file at that path.

PR #837, opened 2020-04-02
Iris public api

syzygyzer:
This PR introduces `/docs/reference/vane-apis/iris.md`, which contains information on the `task`s that Iris can be passed.

I only have one lingering issue:

I'm a bit confused about how `%cancel-request`s are done. The corresponding
`task` has no arguments, so I assume that which fetch to be canceled must
correspond to the duct that the task came on and have written the entry for it based on that assumption, but I haven't understood the code well enough to know for sure that that is what is
happening. The rest of Iris seems to work on an ID number system, one for each http connection, and there are several comments that suggest that ducts should be used to keep track of connections

### DIFF
--- a/reference/vane-apis/iris.md
+++ b/reference/vane-apis/iris.md
@@ -13,58 +13,167 @@ each `task` that Iris can be `%pass`ed, and which `gift`(s) Ames can `%give` in 
 
 ### `%born`
 
+Each time you start your Urbit the Arvo kernel calls the `%born` `task` for
+Iris. This causes Iris to cancel any outstanding http requests and resets all
+connection states.
+
 #### Accepts
 
+```hoon
+[ ~ ]
+```
+
 #### Returns
+
+In response to a `%born` `task` Iris `give`s a `%http-response %cancel` `gift`
+to each outstanding connection.
 
 
 ### `%cancel-request`
 
+This `task` cancels a previous fetch. Iris knows which request is meant based on the
+`duct` that the `task` comes on.
+
+ We note that `%cancel-request` is also a `gift` that Iris can `%give`. 
+
 #### Accepts
+
+```hoon
+[ ~ ]
+```
 
 #### Returns
 
+```hoon
+[%cancel-request id=@ud]
+```
+Iris `%give`s a `%cancel-request` `gift` in response to this `task`. `id` is
+obtained via a `(map duct @ud)`, with the `duct` corresponding to one along
+which the `%cancel-request` `task` came with. 
 
 
 ### `%crud`
 
+`%crud` is called whenever an error involving Iris occurs. It produces a crash
+report in response.
 
 #### Accepts
 
+```hoon
+[p=@tas q=(list tank)]
+```
+
+`p` is the type of error, `q` is the error message.
+
 #### Returns
+
+Iris does not `%give` a `gift` in response to a `%crud` `task`, but it does
+`%slip` Dill a `%flog` `task` instructing it to print the error.
 
 
 ### `%receive`
 
+The `%receive` `task` is used to receive a response to an `http-request` that
+was made.
+
 #### Accepts
 
+```hoon
+[id=@ud =http-event:http]
+```
+
+`id` is the identification number assigned by Iris to the http connection under
+consideration. `http-event` is a packet that contains a header and data.
+
 #### Returns
+
+The response depends on what kind of packet `http-event` is: `%start`,
+`%continue`, or `%cancel` and whether the `complete` flag is `%.y` or `%.n`.
+
+If the packet says to `%start` or `%continue` and `complete=%.n`, Iris will `%give` a
+`%http-response %progress` `gift`.
+
+If the packet says to `%start` or `%continue` and `complete=%.y`, Iris will
+`%give` a `%http-response %finished` `gift`.
+
+If the packet says to `%cancel`, Iris will `%give` a `%http-response %cancel`
+`gift` in return. The `complete` flag is unused here.
 
 
 ### `%request`
 
+`%request` is used to fetch a remote resource. `%request` is also a `gift` Iris
+can `%give`.
+
 #### Accepts
 
+```hoon
+[=request:http =outbound-config]
+```
+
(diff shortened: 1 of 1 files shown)


syzygyzer:
I spoke with @philipcmonk about the `%cancel-request` issue and it sounds like it is fine as it is. So this is ready to be reviewed.

reference/vane-apis/iris.md
```diff
@@ -0,0 +1,179 @@
...
+```
+
+#### Returns
+
+```hoon
+[%cancel-request id=@ud]
+```
+Iris `%give`s a `%cancel-request` `gift` in response to this `task`. `id` is
+obtained via a `(map duct @ud)`, with the `duct` corresponding to one along
+which the `%cancel-request` `task` came with.
```
philipcmonk, line 52: It'd be worth noting the `%cancel-request` it gives is to Unix.  It's canceling the request in the runtime.

reference/vane-apis/iris.md
```diff
@@ -0,0 +1,179 @@
...
+#### Returns
+
+Iris does not `%give` a `gift` in response to a `%crud` `task`, but it does
+`%slip` Dill a `%flog` `task` instructing it to print the error.
+
+
+### `%receive`
+
+The `%receive` `task` is used to receive a response to an `http-request` that
+was made.
```
philipcmonk, line 77: Mention "from Unix"

reference/vane-apis/iris.md
```diff
@@ -0,0 +1,179 @@
...
+ A `$outbound-config` contains the number of redirects and retries that Iris
+ will attempt. By default this is 5 redirects (the recommended limit for the
+ http standard) and 3 retries.
+
+#### Returns
+
+```hoon
+[id=@ud request=request:http]
+```
+Iris will `%give` a `%request` `gift` in response to a `%request` `task`. This
```
philipcmonk, line 130: "to Unix"

reference/vane-apis/iris.md
syzygyzer: I think I see why this is but I want to make sure - the `%born` `task` sets `outbound-duct.state.ax` to `duct`, and `%born` originates from Unix so `outbound-duct` always points to Unix. Then `outbound-duct` is the `duct` on which `+cancel` returns, thus Iris gives `%cancel-request` to Unix.

philipcmonk: Yes, precisely

syzygyzer: Is this return really a response to the `%cancel-request` `task` then since it does not return on the same `duct` that it was passed with? Or is it considered a response to `%born` somehow?

There is this comment in `iris.hoon` that makes it sound like that this is really a return of `%born`, but its not an ideal state of affairs and should really be a pass t unix.

```
:
    ::  the original eyre keeps track of the duct on %born and then sends a
    ::  %give on that duct. this seems like a weird inversion of
    ::  responsibility, where we should instead be doing a pass to unix. the
    ::  reason we need to manually build ids is because we aren't using the
    ::  built in duct system.
    ::
    ::  email discussions make it sound like fixing that might be hard, so
    ::  maybe i should just live with the way it is now?
    ::
```

philipcmonk: I wouldn't call it a "response" per-se, but it's triggered by the `%cancel-request` task.  Much of what Iris does is just translating from other vanes to unix, so most things you ask Iris to do just turn into a card destined for Unix.

syzygyzer: Here's some changes to reflect this. I added a note at the top about "returns to unix" to avoid repeating the same information for `%request` and `%cancel-request`, but I could also see the merit of putting that information under one of the `task`s and then referring to it from the other one.

https://github.com/urbit/docs/pull/837/commits/f23b595b63a162e815eb67d1e1e0555c05c315b2

reference/vane-apis/iris.md
```diff
@@ -7,7 +7,18 @@ template = "doc.html"
...
+return.
+
+## Returns to Unix
+
+Much of what Iris does is translating `task`s from other vanes into `card`s
+meant for Unix via the `%request` and `%cancel-request` `task`s. What this means in practice is that Iris may be `%pass`ed a
+`task` via some `duct`, which in response Iris then `%give`s a `gift` to Unix
+via another `duct`. This `gift` will actually be a response to the `%born`
+`task` that Iris was `%pass`ed when the Urbit was resumed, which came along the
+duct to Unix. This is not an ideal situation but also one not easily remedied.
```
philipcmonk, line 20: I'd lose the last sentence.  I think it's basically righteous for vanes to be translators between urbit-space and unix.

reference/vane-apis/iris.md
```diff
@@ -0,0 +1,193 @@
...
+ will attempt. By default this is 5 redirects (the recommended limit for the
+ http standard) and 3 retries.
+
+#### Returns
+
+```hoon
+[id=@ud request=request:http]
+```
+Iris will `%give` a `%request` `gift` to Unix in response to a `%request`
+`task`. See [Returns from Unix](#returns-from-unix). This `gift`
```
philipcmonk, line 144: *to unix

reference/vane-apis/iris.md
syzygyzer: OK I removed it. I think I must still be misunderstanding something here though. I didn't think having one task trigger a give response to another task was proper form. Is this actually the norm and I've built up the wrong idea in my head?

philipcmonk: Sometimes it's the right thing, yeah.  Subscriptions commonly do this for example: A subscribes to B on duct /foo, then C does something to B on duct /bar that updates the subscription, so B gives an update on /foo to A (and maybe some kind of acknowledgment on /bar to C).

Ducts are conceptual stack traces not bound by mechanical constraints like the literal nock stack trace you see when something crashes.  So conceptually, when the runtime sends a `%born` event, it's saying "I'm running an HTTP client, so lmk if you want to send any requests".  This "causes" Iris to use that whenever someone asks it to perform an HTTP request.

PR #845, opened 2020-04-14
Jael public API

drbeefsupreme:
This PR adds `/docs/reference/vane-apis/jael.md`. It describes `task`s that can
be `%pass`ed to Jael and what `gift`s may be returned in response.

Edit 2020.12.03 - finally looped back to this! My only outstanding question is that the comment above the `%moon` task says its for changing private keys, but the code looks like it changes public keys. I ignored the comment, but I also don't know the details on how moon keys are stored, so perhaps it really does store private keys but uses the `+public-keys:feel` arm.

I also finally came to the conclusion that including `%pass`es that a `task` can trigger might be considered an implementation detail and thus doesn't necessarily belong in an API doc. I'm not really sure on this. I have chosen the worst path, the middle one, and so have a few tasks that say what `%slip`s and `%pass`es they might trigger and some that do not. I would appreciate another perspective on this.

### DIFF
--- a/reference/vane-apis/jael.md
+++ b/reference/vane-apis/jael.md
@@ -9,11 +9,40 @@ template = "doc.html"
 In this document we describe the public interface for Jael.  Namely, we describe
 each `task` that Jael can be `%pass`ed, and which `gift`(s) Jael can `%give` in return.
 
+
 ## Tasks
 
 ### `%dawn`
 
+This `task` is called only a single time per ship during the vane initialization
+phase immediately following the beginning of the [adult
+stage](@/docs/tutorials/arvo/arvo.md#structural-interface-core) to load Azimuth information such as private keys and the sponsor into
+Jael. This `task` is `%pass`ed to Jael by Dill, as Dill is the first vane to be loaded for
+technical reasons, though we consider Jael to be the true "first" vane.
+
 #### Accepts
+
+`%dawn` accepts a `dawn-event`:
+
+```hoon
+    +$  dawn-event
+      $:  =seed
+          spon=(list [=ship point:azimuth-types])
+          czar=(map ship [=rift =life =pass])
+          turf=(list turf)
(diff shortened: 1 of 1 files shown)


syzygyzer:
This is ready to be reviewed except for the `%listen` `task`, which somehow escape my list on things to ask about today.

reference/vane-apis/jael.md
```diff
@@ -0,0 +1,395 @@
...
+with the Azimuth smart contract.
+
+`node` is the URL of the Ethereum node used to register the ship's keys with Azimuth.
+
+#### Returns
+
+Jael `%give`s an `%init` `gift` to Unix. This occurs after the Dill `%slip`
+init.
+
+#### Slips
```
baudtack, line 76: Is what a %slip is explained somewhere else? not saying this is the right place for it but I have no idea what a %slip is. I infer it's something akin to a move?

drbeefsupreme: Yeah it's explained at `tutorials/arvo/arvo.md`. You can find it searching for `%slip` in the search bar, so I think it is fine not linking it here.

reference/vane-apis/jael.md
```diff
@@ -0,0 +1,395 @@
...
+Jael `%give`s a `%done` `gift` in response to a `%plea` `task`. 
+
+#### Passes
+
+Jael `%pass`es itself the wrapped `task`.
+
+
+### `%private-keys`
+
+This `task` creates a subscription to the private keys kept by Jael.
```
baudtack, line 240: What are the private keys that jael holds? are these your networking keys?

baudtack (approved):
Looks pretty good! Learned more about jael reading this than I knew before so that's nice. Approved pending technical review by a jael expert.

philipcmonk (approved):
Sorry for not getting around to this... It looks good!

PR #846, opened 2020-04-15
Removes stray ser from gall tutorial

jfranklin9000:
I also noticed in the two  `++on-poke` arms `state` is `~&`'d slightly differently: `~&  state=state` vs `~&  >  state=state`.  The second one apparently is not  a syntax error.  It prints as `>   state=42` (gar with three aces) in _green_.  Wild.  Questions is, do we want both versions to be the same for consistency?


jfranklin9000:
I updated this PR and screwed up the git-fu.  Will close and re-PR.

PR #848, opened 2020-04-16
remove references to factories

syzygyzer:
This is a small PR that removes two references to "factories", which seems to be
an old name for molds. I asked in the infrastructure channel if this terminology
was still used and a couple people commented that it is not.

----

#


syzygyzer:
Found a sentence that talks about molds like they aren't used anymore, so I changed the wording to be accurate.

PR #853, opened 2020-04-21
Make documentation of ++spin more accurate.

krendil:
Fix order of arguments b and c.
Fix return type.

Addresses issue #851


jfranklin9000:
This still hasn't been deployed to urbit.org.

https://urbit.org/docs/reference/library/2b/#spin

matildepark:
12 days eh? Whoops, I'll push.

PR #861, opened 2020-04-24
update lustar description

syzygyzer:
This addresses #860

There's a few outstanding issues before this should be merged.

1. I'm not sure the syntax is right. I couldn't find where in `hoon.hoon` that lus arms are defined. It's not in the same spot as most of the other runes and grepping for `%lstr` or `$lstr` in the repo doesn't hit anything. So saying `+*  a=term  b=hoon` is just an informed guess. I also call the rune syntax "Regular: **variadic**" but we don't use the word variadic anywhere else in the docs. This might be fine since `+*` is the only variadic rune that doesn't use a terminating rune, so its fine if it has a unique descriptor, but since variadic has its own definition outside of Hoon maybe I should just leave that spot blank, as is the case for other variadic runes such as `|%`, which just says "Regular:" followed by a block of code.

2. I wrote "Under the hood, `+*` gets compiled to `=*`". This is what @Fang- told me but it would be best to say what it actually reduces to if possible. Again I figure I could work this out from reading the code that defines `+*` but I couldn't find it.

3. Related to 2, I called the names that `+*` gives "aliases". I have no opinion on what this should be called since I don't have a strong enough background to really say either way. The other options seem to be calling them macros or tistars. The last discussion on this matter I'm aware of is #826 and since that was merged prematurely, we could continue that discussion here...

4. As has been the case for previous rune PRs, I'd appreciate contributions of examples. I put in the most common uses I've seen, but I'm sure there's more that could be added.


reference/hoon-expressions/rune/lus.md
```diff
@@ -135,42 +135,74 @@ Any Hoon expression, `q`, may be used to define the arm computation.
...
++*  a=term  b=hoon
+    c=term  d=hoon
+    ...
+    e=term  f=hoon
 ```
 
-`p` is the arm name, `q` is the name of the argument for the constructor function enclosed in square brackets, and `r` is a structure expression that defines the constructor function.
+`a`, `c`, `e` are arm names and `b`, `d`, `f` are any Hoon expression. Note that unlike all other
+runes with a variable number of arguments, the list of arguments of `+*` does
+not end with a terminating rune.
```
Fang-, line 153: > all other runes
> terminating rune

I still feel uncomfortable with calling `++` and `--` and such _runes_. `--` is a terminator (for a rune), `++` is, uh, technically superfluous syntax, but good at clarifying where the arm names/sections in a core are.

reference/hoon-expressions/rune/lus.md
```diff
@@ -135,42 +135,74 @@ Any Hoon expression, `q`, may be used to define the arm computation.
...
-Arms produced by `+*` are essentially type constructors.  They are used to construct new types from ones you already have.  Consider `list`, which takes some type -- e.g., `@`, `tape`, etc. -- and returns a new type, e.g., `(list @)`, `(list tape)`, etc.
+The primary use of `+*` is to assign aliases to doors (see Examples below).
+However, they may also be used to define "virtual arms" which behave just like
+ordinary arms except they allow nesting within cores that requires a specific
+number of arms, such as Gall apps. Put another way, no matter how many virtual
+arms are defined in a `+*` statement, the interpreter will see it as only being
+a single arm for the purposes of nesting.
 
-The Hoon subexpression, `r`, must be a structure expression.  That is, it must be either a basic structure expression (`*`, `~`, `^`, `?`, and `@`), or a complex expression made with the `$` family of runes (including irregular variants).
+Under the hood, `+*` gets compiled to `=*`.
```
Fang-, line 167: See [its destructuring](https://github.com/urbit/urbit/blob/3196179fbdc7afe4f5d4c8143b85ee7c5b9c7d41/pkg/arvo/sys/hoon.hoon#L8680-L8688).

If I'm getting this right (@joemfb please confirm) that means that this:

```hoon
|_  z=@ud
+*  n  1
++  x  (add z n)
++  y  (sub z n)
--
```

becomes roughly:

```hoon
=|  z=@ud
|%
++  x
  =*  n  1
  (add z n)
++  y
  =*  n  1
  (sub z n)
--
```

reference/hoon-expressions/rune/lus.md
```diff
@@ -135,42 +135,74 @@ Any Hoon expression, `q`, may be used to define the arm computation.
...
+
+`+*` arms must always come at the beginning of the battery, before any other
+type of lus arm.
 
 ##### Discussion
 
-Arms produced by `+*` are essentially type constructors.  They are used to construct new types from ones you already have.  Consider `list`, which takes some type -- e.g., `@`, `tape`, etc. -- and returns a new type, e.g., `(list @)`, `(list tape)`, etc.
+The primary use of `+*` is to assign aliases to doors (see Examples below).
+However, they may also be used to define "virtual arms" which behave just like
+ordinary arms except they allow nesting within cores that requires a specific
```
Fang-, line 162: > behave just like ordinary arms

Wondering if this is 100% true.

reference/hoon-expressions/rune/lus.md
```diff
@@ -135,42 +135,74 @@ Any Hoon expression, `q`, may be used to define the arm computation.
...
 ```
-> =c |%
-       +*  triple  [a]  [a a a]
-       +*  wrap-flag  [a]  [? a]
-     --
+This is the idomatic way to assign the alias `this` to the door.
 
-> `(triple.c @)`[12 14 16]
-[12 14 16]
+Gall apps have a fixed number of arms, but sometimes you'd like to have more.
```
Fang-, line 178:
```suggestion
Gall app cores have a fixed number of arms, but sometimes you'd like to have more.
```

But also, the general solution there is to

```hoon
=<
  |_  =bowl:gall
  ++  on-init  etc
  --
::
|_  =bowl:gall
++  more-arms  etc
--
```

or similar, often with a `+*` pointing to that second core for easy of access. We don't (and shouldn't) go around slapping a thousand `+*` into agent cores.

reference/hoon-expressions/rune/lus.md
```diff
@@ -135,42 +135,74 @@ Any Hoon expression, `q`, may be used to define the arm computation.
...
-> `(wrap-flag.c @)`[& 22]
-[%.y 22]
-```
\ No newline at end of file
+```hoon
+|_  foo
++*  this  .
++*  samp  +<
++*  cont  +>
+```
```
Fang-: Huh, I thought this didn't work?

reference/hoon-expressions/rune/lus.md
```diff
@@ -135,42 +135,74 @@ Any Hoon expression, `q`, may be used to define the arm computation.
...
+
+You are not restricted to lark expressions. Functions may be defined as well.
+Another common snipper seen in Gall apps is the following.
```
Fang-, line 202: We should avoid giving lark as examples, then. The `+*` is important here, we don't want people scratching their heads over its contents while trying to reason about the `+*` itself.

reference/hoon-expressions/rune/lus.md
joemfb: Yeah, the core arms (`++`, `+*`, `+$`) are not runes, but separators (like the comma in `%=(a b 1, c 2)`). And the terminators (`--`, `==`) are *definitely* not runes.

reference/hoon-expressions/rune/lus.md
joemfb: That's right.

reference/hoon-expressions/rune/lus.md
joemfb: Yeah, you only get (up to) one `+*`. Needless to say, that's pretty controversial.

reference/hoon-expressions/rune/lus.md
syzygyzer: Wow, this is news to me. I am pretty sure this is not made explicit in Hoon school (but I'll check later). And we definitely have these things in the runes portion of the reference section and other spots too. I'm going to make a new issue for this.

reference/hoon-expressions/rune/lus.md
syzygyzer: I suppose I misunderstood your answer when I asked this in our call @Fang- . I should definitely make a habit of testing all code before putting it into a PR at the very least.

reference/hoon-expressions/rune/lus.md
syzygyzer: Yes, good point!

reference/hoon-expressions/rune/lus.md
baudtack: We used to refer to them as "ascii digraphs" but at some point we just gave up. I agree they are not runes, but calling them ascii digraphs is cumbersome and we also had to document _what_ they did in the cases of the lus ones in particular so it got filed under runes as that is where you're likely to look for that information.

reference/hoon-expressions/rune/lus.md
```diff
@@ -135,42 +135,82 @@ Any Hoon expression, `q`, may be used to define the arm computation.
...
+runes with a variable number of arguments, the list of arguments of `+*` does
+not end with a terminating rune.
+
+`+*` arms must always come at the beginning of the battery, before any other
+type of lus arm.
 
 ##### Discussion
 
-Arms produced by `+*` are essentially type constructors.  They are used to construct new types from ones you already have.  Consider `list`, which takes some type -- e.g., `@`, `tape`, etc. -- and returns a new type, e.g., `(list @)`, `(list tape)`, etc.
+The primary use of `+*` is to assign aliases to doors (see Examples below).
```
baudtack, line 160: Unless I am misunderstanding, this use of this rune should be discouraged as it' deprecated in favor of using |$

lukechampine:
Thanks for this. `+*` confused the hell out of me when I was trying to understand Gall apps.

IMO `+*` is too weird, with its unterminated jogging syntax and its position sensitivity, to the point where people are confused as to whether it's even a "real rune." @joemfb explained it as:

> `+*` creates `=*` aliases in every arm of the core; it's really just a macro that transforms the other arms.

Which helped me a lot. But it also made me wonder whether something like this would be possible?

```
:: instead of:
::
|_  =bowl:gall
+*  this  .
    main  ~(. +> bowl)
    def   ~(. (default-agent this %|) bowl)
::
:: write:
::
+*  this  .
    main  ~(. +> bowl)
    def   ~(. (default-agent this %|) bowl)
  ==
|_  =bowl:gall
```

that is, instead of `+*` being defined inside the door, it's a proper rune that takes a door as one of its arguments; "Transform the supplied door, adding this list of aliases to each of its arms." This dispenses with the weird position requirement and makes it clear that `+*` is operating on the door as a whole, rather than being its own arm or something.

Sorry, I know this isn't really the place to be suggesting new Hoon syntax, but the other issue was closed as "wontfix," so... ¯\\\_(ツ)\_/¯

reference/hoon-expressions/rune/lus.md
syzygyzer: I've been trying to come up with a short and concise example of doing something other than assigning aliases with lustar, but I haven't quite been able to wrap my head around the examples I've come across in the code base.

Everywhere I've looked, you have something like

```
=<
  |_  =bowl:gall
  +*  do   ~(. +> bowl)
  ++  stuff  foo:do
::etc
--
|_  =bowl:gall
  ++  foo  %bar
::etc
--
```

I'm pretty sure I understand what is going on here - `do` is used to call arms in the second door with the sample of the first, but I don't understand the syntax with `%~`. The documentation says the syntax is `(arm door args)`, but `.` is the subject, which is the door, which isn't an arm...?

Are doors somehow considered to be arms, is the subject `.` is referring to not actually the parent door, is the doc on `%~` wrong, or what?

ohAitch:
As per [a recent-ish urbit-dev post](https://groups.google.com/a/urbit.org/forum/#!topic/dev/_vSJWsxdBhQ), imo all uses of `+*` should be replaced with `++`, and the core chapter system extended slightly to let this not break static gall. (I am unaware of any non-static-gall motivated uses of the rune.)

But as things stand, the lack of terminator is particularly confusing and inconsistent, certainly.

reference/hoon-expressions/rune/lus.md
syzygyzer: I tried playing around with them a bit and get the impression that I am wrong, but I don't understand what's going on well enough to say what exactly is different.

reference/hoon-expressions/rune/lus.md
syzygyzer: Alright, I changed it to talk about "`++` arms" instead, and changed terminating runes->terminators.

reference/hoon-expressions/rune/lus.md
lukechampine: IIUC `.` is in fact an arm; an arm is a subset of the subject, which can include the entire subject. So `~(. +> a)` is slamming `+>` (the door) with `a` to produce a core, and then evaluating `.` in the resulting expression, which is the core itself.

loganallenc:
+1 I hate `+*` and think it needs to be removed.

reference/hoon-expressions/rune/lus.md
syzygyzer: I think deferring to how `+*` turns into `=*`s properly explains what is actually meant by virtual arm. https://github.com/urbit/docs/pull/861/commits/ee469e5acb24eb098227a8fbf68d33d390fd45a4

syzygyzer:
Anybody have any lingering concerns on this PR? I see question on lustar semi-regularly so I'd like to see this merged soon.

joemfb:
@syzygyzer, I'm not sure the distinction between "alias" and "virtual arm" is helpful, they are in fact the same thing, and work the same way. Otherwise this looks right to me.

@lukechampine, it might be possible, but it would be an even bigger departure from how we currently macro-expand runes. To be useful, the aliases need to be "inside" the core (ie, have the core in their lexical subject). The easy way to do that is to add them to every arm, which is what I did. But it'd probably be better to represent explicitly in the `%core` type. (And maybe make `+*` a unary syntax, and allow them anywhere in any battery, but this is getting far afield ...)

ohAitch:
Or remove +* from the language, and fix the part where gall requires a
fixed set of arms :)

syzygyzer:
@joemfb How's this? https://github.com/urbit/docs/pull/861/commits/6bfa63e9c4f5b486b3dfe922067acfb2deafe636

I still mention "virtual arm" once, since I've seen that term used enough times in this context that I don't want to completely leave it out.

PR #877, opened 2020-05-12
barbuc rune documentation

syzygyzer:
This PR adds a new entry to `bar.md` describing the barbuc `|$` rune. It also
fixes some minor typos and adds a clarification to the `|*` entry on when to use
it instead of `|$`.

I have some outstanding questions that need to be answered before it can be
merged.

1) I'm not sure that the `$` arm in a `|$` core ought to be
utilized. I looked for a place in the Urbit code where this happens but came up
empty-handed.

2) Usage of `|$` over `|*` looks to be about style to me, and I wrote the entry
from that perspective, but I'd like some confirmation that it really is just a
restricted form of `|*` used for making structures.

3) I don't know what `|$` expands to. I included the code from `hoon.hoon` that I
think this can be inferred from, but I found it inscrutable and thought it may
be a better use of my time to just ask what the expansion is.

See related issue https://github.com/urbit/docs/issues/813
----

#


reference/hoon-expressions/rune/bar.md
```diff
@@ -8,12 +8,77 @@ Core expressions produce cores. A core is a cell of `[battery payload]`.
...
+I'm having some trouble deciphering this, would appreciate help. From `hoon.hoon`:
+```hoon
+ {$brbs *}  =-  ?~  -  !!
+                       [%brtr [%bscl -] [%ktcl body.gen]]
+                   %+  turn  `(list term)`sample.gen
+                   |=  =term
+                   ^-  spec
+                   =/  tar  [%base %noun]
+                   [%bsts term [%bssg tar [%bshp tar tar]]]
+```
```
Fang-, line 52: Expansion is as follows:

```hoon
|$  [a b]
hoon
```

becomes

```hoon
|*  [a=$-(* *) b=$-(* *)]
^:
hoon
```

(I leave out the `$~` around the `$-`, because I don't think they actually change anything? Not sure why they're there... cc @joemfb)

reference/hoon-expressions/rune/bar.md
```diff
@@ -8,12 +8,77 @@ Core expressions produce cores. A core is a cell of `[battery payload]`.
...
+Five core runes (`|=`, `|.`, `|-`, `|*`, and `|$`) produce a
 core with a single arm, named `$`. We can recompute this arm
-with changes, which useful for recursion among other things:
+with changes, which is useful for recursion among other things.
 
 ## Runes
 
+### |$ "barbuc"
+
+Produces a wet gate that creates a structure parameterized by molds.
```
Fang-, line 19: >structure

Might be better to more explicitly call the result a type/mold? Same applies below.

reference/hoon-expressions/rune/bar.md
joemfb: The difference between `$-(* *)` and `$~(* $-(* *))` is material -- I invite you try it out!

reference/hoon-expressions/rune/bar.md
joemfb: `|$` produces a mold builder, which is a wet gate with one or more molds as its sample, producing a mold.

joemfb:
> I'm not sure that the $ arm in a |$ core ought to be utilized. 

*Ought* is a mighty big word in this context ;) I don't know about that, but I had to use `$` to recur correctly in the definition of `+wite` (see `arvo.hoon`).

reference/hoon-expressions/rune/bar.md
syzygyzer: Thanks, I had a feeling that I wasn't saying what it produced quite right. I've changed the wording everywhere to say it produces a mold builder.

https://github.com/urbit/docs/pull/877/commits/1d189072daccbf7ade5851682608a4e11163ba25

reference/hoon-expressions/rune/bar.md
Fang-: Oh, I had tried everything, except this:

```hoon
> (($-(* *) 5) 6)
0
> (($~(* $-(* *)) 5) 6)
6
```

Amended my expansion above. Can't say I exactly understand this though...

syzygyzer: Thanks https://github.com/urbit/docs/pull/877/commits/6bcae1514962d02d246cb198e67fcdc90d94d153

I need to look up how the most basic runes are defined in this fashion so I can figure out what the hell is going on here :)

reference/hoon-expressions/rune/bar.md
```diff
@@ -8,12 +8,78 @@ Core expressions produce cores. A core is a cell of `[battery payload]`.
...
+```
+
+AST:
+```hoon
+[%brbs sample=(lest term) body=spec]
+```
+
+##### Semantics
+
+`|$` is used to produce a wet gate that makes a mold builder. The mold builder
```
Fang-, line 36:
```suggestion
`|$` is used to produce a wet gate that is a mold builder. The mold builder
```

...right?

syzygyzer: Yeah, lol. I was trying to describe the rune as a mold builder builder, but mangled the wording to make it sound like the gate is making a mold builder instead.

ohAitch:
"|$ declares mold builder wet gates"?

syzygyzer:
> "|$ declares mold builder wet gates"?

I'm guessing this was in some earlier commit but its not there anymore.

syzygyzer:
Or are you suggesting to use that as the description @ohAitch? It initially looked like a sentence fragment to me so I thought you were pointing out a grammar error.

ohAitch:
I am suggesting to use that as a fragment of a description :P

syzygyzer:
@ohAitch Yeah I think the old wording was too clunky, I implemented your suggestion.

syzygyzer:
Actually, mold builders are always gates right? "Wet mold builder" sounds better to me than "wet gate mold builder".

reference/hoon-expressions/rune/bar.md
```diff
@@ -8,12 +8,78 @@ Core expressions produce cores. A core is a cell of `[battery payload]`.
 The `battery` is code, a battery of Nock formulas.  The `payload` is the data
 needed to run those formulas correctly.
 
-Four core runes (`|=`, `|.`, `|-`, and `|*`) produce a
+Five core runes (`|=`, `|.`, `|-`, `|*`, and `|$`) produce a
 core with a single arm, named `$`. We can recompute this arm
-with changes, which useful for recursion among other things:
+with changes, which is useful for recursion among other things.
```
joemfb, line 13: > We can recompute this arm

This is true of any arm, but this framing implies that might not be the case.

ohAitch:
Mold builders are always wet gates even, since dry gates have a fixed type

reference/hoon-expressions/rune/bar.md
```diff
@@ -8,12 +8,78 @@ Core expressions produce cores. A core is a cell of `[battery payload]`.
...
 
 ## Runes
 
+### |$ "barbuc"
+
+Declares a mold builder wet gate with one or more molds as its
+sample.
```
joemfb, line 20: A "mold builder" is simply a wet gate from one-or-more molds to a mold. A mold is a (maybe) partial, (definitely) idempotent, (usually) identity  function from noun to some more specific type.

reference/hoon-expressions/rune/bar.md
```diff
@@ -8,12 +8,78 @@ Core expressions produce cores. A core is a cell of `[battery payload]`.
...
+[%brbs sample=(lest term) body=spec]
+```
+
+##### Semantics
+
+`|$` is used to declare a wet gate that is a mold builder. The mold builder
+is polymorphic in its input types.  `a` is a `lest` of `term` used as
+identifiers for the input types. `b` is a structure built from elements of `a`.
+The output of `|$` is a mold builder that produces a `spec` obtained by
+substituting the input types parameterized by `a` into `b`.
```
joemfb, line 40: A `$spec` is the usual compile-time representation of a mold. A mold-builder produces a mold, not a spec.

reference/hoon-expressions/rune/bar.md
syzygyzer: https://github.com/urbit/docs/pull/877/commits/0a751e0c212968d4fbd362a9d6713e133fb84195

reference/hoon-expressions/rune/bar.md
syzygyzer: How's this? https://github.com/urbit/docs/pull/877/commits/1a0e1a4616736a323a80eddc823a049b9cfe0cf0

reference/hoon-expressions/rune/bar.md
syzygyzer: Thank you! I have been confused about this since forever. Also rolled this into https://github.com/urbit/docs/pull/877/commits/1a0e1a4616736a323a80eddc823a049b9cfe0cf0 which is the same commit for the above convo

syzygyzer:
This PR was a bigger mess than I thought it was when I put it up, but its been really helpful for my understanding of Hoon's type system I knew was very scrambled but I wasn't aware how scrambled, so thank you for all the suggestions and comments and corrections @joemfb @Fang- @ohAitch

PR #884, opened 2020-05-27
bucbar rune entry

drbeefsupreme:
This PR adds an entry to `/docs/reference/hoon-expressions/rune/buc.md` that
explains the `$|` rune.

I made some minor changes to the matter at the top, such as removing any
reference to "factories" and mentioning `$|` in the context of validation.

There is one outstanding missing section, which is what `$|` expands to. I would
appreciate help on that.

----

#


reference/hoon-expressions/rune/buc.md
```diff
@@ -52,6 +51,84 @@ Irregular: `*` makes `%noun`, `^` makes `%cell`, `?` makes
...
+        ::  assert
+        ::
+        :+  %wtbn
+          ::  run the verifier
+          ::
+          [%cnhp [%tsbn $/3 q.mod] $/2]
+        ::  produce verified product
+        ::
+        $/2
+```
```
Fang-, line 80: Oof ok. The `$/2` etc expand to `[%$ 2]` etc, and are tree addressing. So, this becomes, using the more-familiar lark syntax:

```hoon
=+  relative(mod p.mod)
?>  (=>(+ q.mod) -)
-
```

In a slightly more readable style, with some shorthand/fake names:

```hoon
=/  =hoon  processed-spec
?>  (=>(+.hoon validator) -.hoon)
-.hoon
```

So, it's evaluating our validator code against the head of the (processed) spec, then calling the resulting function with the tail of the (processed) spec, and asserting that the result is true (as in, good/sane data).

I _think_ the head vs tail of the generated hoon is about getting the bunt vs the type-function itself, but not 100% clear on that. `+relative` just produces a hoon, so idk how this is safe/sane to do.

Once again pinging @joemfb for the details and sanity check. Most of the above I just figured out on the spot. (lmk if you want me to ping someone else going forward)

reference/hoon-expressions/rune/buc.md
```diff
@@ -52,6 +51,84 @@ Irregular: `*` makes `%noun`, `^` makes `%cell`, `?` makes
...
+
+
+##### Syntax
+
+Regular: **2-fixed**
+
+##### Discussion
+
+`$|` takes two arguments, a mold and a Hoon expression that is typically a gate
+that is used to validate the structure produced by the mold to ensure that the noun
```
Fang-, line 90: As per the above, I think it _needs_ to be a gate?

reference/hoon-expressions/rune/buc.md
```diff
@@ -52,6 +51,84 @@ Irregular: `*` makes `%noun`, `^` makes `%cell`, `?` makes
...
+has a certain shape. It crashes if the input fails the validation test.
+
+For example, the elements of a `set` are treated as being
+unordered, but the values will necessarily possess an order by where they are in
+the memory. Thus if every `set` is stored using the same order scheme then faster algorithms involving `set`s may be
+written. This is not the same thing as casting - it is forcing a
```
Fang-, line 96: Maybe worth mentioning explicitly, if you just insert things in random locations in the set tree, they might be missed or mistreated by the algorithms that are in place.

reference/hoon-expressions/rune/buc.md
```diff
@@ -22,10 +21,10 @@ x)))`, it's a normalizer and can be used as a mold.
...
 definition perfectly.  This is not actually a problem.)
 
-Validation, though very
-important, is a rare use case.  Except for direct raw input,
-it's generally a faux pas to rectify nouns at runtime -- or even
-in userspace.
+Validation (done with [`$|`](#bucbar), though very important, is a rare use
+case. Except for direct raw input, it's generally a faux pas to rectify nouns at runtime -- or even
+in userspace. Nonetheless they are sometimes utilized for structures that will
+be faster to use if they satisfy some validating gate.
```
joemfb, line 27: I think these normative claims are too strong. And the word "rectify" isn't quite right. `$|` is used to add validation predicates that the type system cannot itself enforce. It does not affect bunting, casting, or type inference. And it doesn't change the product of a mold-gate. It simply applies additional validation when clamming with a mold.

reference/hoon-expressions/rune/buc.md
```diff
@@ -7,9 +7,8 @@ aliases = ["docs/reference/hoon-expressions/rune/buc/"]
 The `$` family of runes is used for defining custom types.  Strictly speaking,
 these runes are used to produce 'structures'.  A structure is a compile-time
 value that at runtime can be converted to either an example value (sometimes
-called a 'bunt' value) for its corresponding type, or to a 'factory' (sometimes
-called a 'mold').  An example value is used as a placeholder for sample values,
-among other things.  A factory/mold is used as a data validator.
+called a 'bunt' value) for its corresponding type, or to a 'mold'.  An example value is used as a placeholder for sample values,
+among other things.  A mold is used as a data validator.
```
joemfb, line 11: I think this section is a little confusing. A `$spec` is a compile-time representation of a mold. A mold can be "bunted", producing a default value, or used to "clam" some arbitrary noun, validating the shape of that noun (and thereby proving its type to the compiler) or crashing.

reference/hoon-expressions/rune/buc.md
joemfb: I don't think we want to include the code-generation from $spec to $hoon in these entries, some of it is fairly complicated and subtle. I'm honestly not sure if the "expands to" sections are worth it for molds.

It seems fairly straightforward to say that `$|(a-mold a-validation-gate)` constructs a mold that will apply `a-validation-gate` to the product of `a-mold` when clammed, but less obvious to write out the code that it will use to do so.

That being said, this code is generating a mold-gate from a `$|`. Let's say that the mold is `a` and the validation-gate is `b` (ie, `$|(a b)`. The resulting gate will

- pin the product of clamming with `a` (call this `foo`)
- call `b` (the validation gate) on foo
- assert that the product of `(b foo)` is `&`
- and produce `foo`

So `$|(a b)` is something like

```
|=  x=*
=/  foo  ;;(a x)
?>  (b foo)
b
```

reference/hoon-expressions/rune/buc.md
joemfb: Yes, it needs to be a gate producing a flag/loobean.

reference/hoon-expressions/rune/buc.md
syzygyzer: Yeah I didn't know what was meant by "rectify" here either so I just left most of it as it was.

Should I just nix everything that's there and replace it with a paraphrase of what you wrote here? I'm not sure which specific sentences you think are too strong.

reference/hoon-expressions/rune/buc.md
syzygyzer: I am still confused. We have at least five interrelated concepts here: types, structures, `$spec`s, molds, and bunt values. Can you evaluate the correctness of each of the following statements? I'm trying to combine what you wrote with what was already there, so there are some contradictions and this is going to be confusing since I am confused.

1. Types exist on an abstract mathematical level. They are simply sets (in the mathematical sense, not the Hoon sense) of nouns.
2. Structures are built from buc runes at compile time. They give the computer the definition of a type in terms of other basic types built into Hoon or other structures.
3. Alternatively, structures are also just types (and so only exist in a mathematical sense). The distinction arises since Hoon only has a few basic types to start with, and structures are types built from these building blocks.
4. A `$spec` is actually what is meant by #2, which means its either the same thing as a structure, or it is how structures are represented in the computer.
5. Molds are built by the compiler to determine which nouns fit a `$spec` (or structure?) and thus are a member of the type it corresponds to. The runtime can use a mold to either clam a noun or bunt it to produce an example value.
6. When a mold is used to clam a noun, it proves its type to the _runtime_ (you wrote compiler, but I don't understand why a compiler would be feeding nouns to a mold at compile-time).
7. The runtime can be asked to bunt a mold, which produces an example value.

To summarize where on the ladder each of these ideas live:
Mathematical: types, maybe structures?
Compile-time: `$spec`s, molds, maybe structures?
Runtime: clamming and bunting.

reference/hoon-expressions/rune/buc.md
baudtack: I think doing what you suggest makes sense here @syzygyzer

baudtack:
@syzygyzer have @joemfb 's comments been resolve here?

reference/hoon-expressions/rune/buc.md
syzygyzer: Any time to answer these questison @joemfb ?

reference/hoon-expressions/rune/buc.md
baudtack: @syzygyzer do you have the time to make this change this week?

syzygyzer:
No. @joemfb would you be able to meet sometime next week and we can dig into the dirty details on types so I can clear up my confusion once and for all?

joemfb:
@syzygyzer, sorry I never answered your questions. i'll probably have to put you off a couple more weeks, then I'll be free to discuss.

reference/hoon-expressions/rune/buc.md
syzygyzer: Ignore this and check my new commit

syzygyzer:
I've split the introductory material at the top into #997 and revised the bucbar entry using my newfound understand of the relationship between spec, type, and molds.

Please review this when you get a chance @joemfb

drbeefsupreme:
I'm merging this with @jalehman 's permission. We're cleaning up old PRs that have had ample time to be reviewed.

PR #885, opened 2020-05-28
bucpam rune entry

syzygyzer:
This PR adds an entry for bucpam `$&` to
`docs/reference/hoon-expressions/rune/buc.md`. See
https://github.com/urbit/docs/issues/687 for some discussin of this rune.

I am not sure about the restrictions I mention here, namely that the first
argument must be a tagged union type `combined` and the second argument must be a gate that
accepts values of `combined` and returns values of one particular type in `combined`.

The `Defaults to` behavior is something I determined from a couple of tests but
I cannot be certain that that is always the behavior. This is a representative
test:

```
> =old0 ?([%0 @])
> =old1 ?([%1 @])
> =old2 ?([%2 ^])
> =combined $%(old0 old1 old2)
> =adapting $&(combined |=(a=combined ?-(-.a %0 [%2 1 +.a], %1 [%2 2 +.a], %2 a)))
> *adapting
[%2 0 0]
> =combined $%(old1 old2 old0)
> =adapting $&(combined |=(a=combined ?-(-.a %0 [%2 1 +.a], %1 [%2 2 +.a], %2 a)))
> *adapting
[%2 1 0]
> =combined $%(old2 old0 old1)
> =adapting $&(combined |=(a=combined ?-(-.a %0 [%2 1 +.a], %1 [%2 2 +.a], %2 a)))
> *adapting
[%2 2 0]

```

Given my previous level of confusion over all the nomenclature around type
theory in Hoon, reviewers ought to pay close attention to if I'm saying things
like "normalize", "mold", "type" etc. correctly.

----

#


Fang-:
This might be my first time seeing this rune? It doesn't seem to be used anywhere in arvo it seems. Might be a cute way to write state adapters, but maybe not?

Best I got for you is "I believe it", except that I don't see anything that enforces the kind of type you're doing this on, union or not. I'll defer to whoever actually knew about the existence of this already.

syzygyzer:
@Fang- You are the person who brought this rune to my attention!

You're right that it's currently not used anywhere though. I had it half written up before I bothered to grep for it and would have just left it off it I weren't already mostly done.

PR #887, opened 2020-06-05
Hoon school lesson on =>

syzygyzer:
When I first wrote this part last year I didn't have a complete understanding of
what `=>` was doing, especially that `=> a b` where `a` and `b` are cores puts
`a` in the context of `b`. I've rephrased and expanded the section in `tutorials/hoon/arms-and-cores.md` about the
usage of this rune, added another example, and rephrased the associated
exercise.

One thing to note when reviewing this is that up until this point in Hoon
school, the payload of gates has been said to be `[sample context]` but the
general notion of cores (introduced in this section)
have only been said to have a `payload` and it had not yet been stated that a
payload is always a cell called `[sample context]`. In fact because of this, I am not actually
sure this is always how `+3:core` is named. Thus I think this needs at
least one more edit to clarify this point, since I start off by recalling that
the payload of a gate is `[sample context]` but then just talk about cores and
their contexts in general without ever outright saying that the payload of a
core is always `[sample context]`.

----

#

### DIFF
--- a/tutorials/hoon/arms-and-cores.md
+++ b/tutorials/hoon/arms-and-cores.md
@@ -488,7 +488,7 @@ This core has one arm `dec` which implements decrement. If we look at the head o
  
 Again, being able to read Nock is not essential to understanding Hoon.
 
-### Core Nesting
+### Cores and Contexts
 
 Let's take a quick look at how cores can be combined with `=>` to build up
 larger structures.  `=>  p=hoon  q=hoon` yields the product of `q` with the product of `p` taken as the subject.
--- a/tutorials/hoon/arms-and-cores.md
+++ b/tutorials/hoon/arms-and-cores.md
@@ -497,8 +497,7 @@ We can use this to set the context of cores. For example:
 ~zod:dojo> =>([1 2] |=(@ 15))
 <1.fed {@ @ud @ud}>
 ```
-Here we have created a gate that takes in an `@` and returns `15` with `[1 2]`
-as its context.
+Here we have created a gate with `[1 2]` as its context that takes in an `@` and returns `15`.
 
 `=>` (and its reversed version `=<`) are used extensively to put cores into the
 context of other cores.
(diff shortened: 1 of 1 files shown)


tutorials/hoon/arms-and-cores.md
```diff
@@ -99,7 +99,7 @@ First let's use the dojo to bind the face `a` to the value `12`, and the face `b
 
 Now use the following multi-line expression to bind `c` to a core.  As before, feel free to cut and paste the following expression into the dojo (starting at `=c`):
 
-```hoon
```
Fang-, line 102: -1 for removing these. It's good metadata to have around.

tutorials/hoon/arms-and-cores.md
```diff
@@ -490,8 +490,23 @@ Again, being able to read Nock is not essential to understanding Hoon.
 
 ### Core Nesting
```
Fang-, line 491: This title is kind of weird. Is it really nesting? It's certainly not hoon type nesting, for sure. idk if that'll lead to confusion down the line or not.

tutorials/hoon/arms-and-cores.md
syzygyzer: They're dojo prompts though. We don't use that tag for dojo elsewhere in the docs.

syzygyzer: I definitely think we should have another tag for dojo, but that's a separate issue.

tutorials/hoon/arms-and-cores.md
syzygyzer: You're right, that was one reason I decided to rewrite this section, but I forgot to change the title.

tutorials/hoon/arms-and-cores.md
matildepark: When I was tagging all the code blocks during the urbit.org write I sometimes slipped `hoon` in if the code block was predominantly Hoon (not Dojo specific syntax).

PR #889, opened 2020-06-09
Add links to library docs 1a-2a

alecbrick:
I've felt that the standard library docs are lacking in links between doc pages, so I've added some.

Let me know if we should standardize on whether we should be including `++` when linking within the standard library, or whether or not these links should be in `this font` or not.

### DIFF
--- a/reference/library/1a.md
+++ b/reference/library/1a.md
@@ -12,13 +12,13 @@ Produces the sum of `a` and `b`.
 
 #### Accepts
 
-`a` is an [`atom`]('/docs/glossary/atom').
+`a` is an [`atom`](/docs/glossary/atom).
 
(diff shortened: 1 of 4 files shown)


rmariani:
This is helpful! But some of these links do not work, such as the ones that link to nonexistent places like `docs/glossary/library/1c.md`

alecbrick:
Links are fixed, I think!

rmariani:
Thanks. Some of your links have file extensions (`.md`), which should not be present in links

matildepark:
They absolutely do need to be in relative links (`@/` etc) or the build will fail.

rmariani:
Oh right I got that backwards. Ignore!

baudtack:
@alecbrick please address the comments here and reopen the pr when that is done.

PR #890, opened 2020-06-09
Add micfas example.

sigilante:
While I'm at it, colfas :/ is used in the example but undocumented.  Want me to add that one as well?  (Unless that's not a colfas rune but an irregular form I've not seen before.)


syzygyzer:
Thanks @sigilante, `:/` is actually not a rune so I believe this was a typo. You can find a listing of all the runes in `hoon.hoon` under `+$  hoon` around line 6670

joemfb:
I just saw this comment -- `:/` works, and looks like irregular syntax for `;/`. Is sail documented?

syzygyzer:
@joemfb I don't think Sail is documented outside of a page on the Glossary and the mic page https://urbit.org/docs/reference/hoon-expressions/rune/mic/

I don't know anything about it personally.

I'll make a quick PR mentioning that this is irregular syntax in a moment. Where can I find the Hoon that defines irregular syntax?

joemfb:
`+vast` is the hoon parser (both meanings work).

PR #892, opened 2020-06-12
basic doc of urlockpy as an example of airlock library

baudtack:


reference/vane-apis/airlock.md
```diff
@@ -0,0 +1,131 @@
+# Airlock
+
+Airlock is the term to describe the set of apis for interacting with a running Urbit ship from outside of the Urbit network.
```
rmariani, line 3: APIs should be caps

reference/vane-apis/airlock.md
```diff
@@ -0,0 +1,131 @@
...
+`subscribe(ship, app, path)`
+
+    `ship` - ship on which the gall application lives
+
+    `app` - gall application to subscribe to
+
+    `path` - path to subscribe on
+
+
+Follows is a simple script to send a message and subscribe to a chat channel. The specifics require an understanding of the `chat-store` and `chat-hook` interfaces.
```
rmariani, line 57: "Follows" should be "below"

loganallenc:
Worth noting: Airlock was a codename, I don’t think it works as a “brand name” for APIs that interact with Urbit

galenwp:
I always kind of liked the name 'airlock'. Do you have another suggestion? Or a reason you don't like it?

loganallenc:
I don’t think there’s any reason to brand “APIs that interact with Urbit”... More terminology is confusing. Just say it’s an API that talks to Urbit.

baudtack:
Strong disagree. I like Airlock. It's particularly evocative. Does the api layer for urbit _need_ a name? probably not. But we have a nice one and feels like a shame to not use it.

PR #913, opened 2020-07-17
global replace of odor with aura

syzygyzer:
This old terminology was all over the place in the standard library docs, and in
one spot on the Eyre doc. I just did a find-and-replace and didn't check
anything closely, hopefully we weren't using odor as some colorful language somewhere.

----

#

### DIFF
--- a/reference/library/4b.md
+++ b/reference/library/4b.md
@@ -375,9 +375,9 @@ A `++tape`.
 ---
 ### `++sand`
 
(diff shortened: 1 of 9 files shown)


matildepark:
I always thought these were separate things. But idk

syzygyzer:
I asked in infra and were told they were the same
--

—

~tidrel-tabpub

https://urbit.org

PR #917, opened 2020-07-20
purging or warning labels on arvo kernel and ford

syzygyzer:
Per the current docs cleanup effort, this PR does the following:

 - Warns that `arvo/arvo.md` is out of date. I wrote this doc - I don't know the full extent of
   this yet, and its mostly still accurate, but `move`s were changed to
   potentially including debugging information and the way in which the kernel
   updates itself was altered since I wrote it. I've put a warning at the top.
   This is prioritized to be updated this week.
 - I scrapped `arvo/ford.md` but put a link to the Ford Fusion blog post. This
   is not what I said I'd to in the docs meeting (which was to just delete it),
   but I think linking to the blog post is a better move than totally killing
   it.
 - I removed the Ford and unit testing Hoon school lessons completely. I talked
   to @belisarius222  about how to fix these last week, and am prioritizing doing it this
   week, but since he is on vacation for the next few weeks and he ought to
   review the PR it makes sense to remove it for now.
   
 The reason for the seemingly arbitrary inclusion of both Ford and Arvo docs in
 this PR is simply that this is what was assigned to me and it should only take
 a few minutes to review.

----

#


tutorials/arvo/arvo.md
```diff
@@ -4,6 +4,10 @@ weight = 1
 template = "doc.html"
 aliases = ["/docs/learn/arvo/arvo/"]
 +++
+
+Parts of this document are out of date, please use information here with
+caution. It is being updated as of 2020.07.20.
```
rmariani, line 9: Giving a date for when it is being updated (a future occurrence) is a little confusing. Maybe add a phrase saying that the message will be removed when it's updated?

Also using 

> This

formatting might make it stand out more.

syzygyzer: Fair, how's this sound

https://github.com/urbit/docs/pull/917/commits/613bc58ec6cb671abe776571026662dec2ffde4a

PR #920, opened 2020-07-24
issue #902

baudtack:
updated cheat sheet

### DIFF
--- a/reference/cheat-sheet.md
+++ b/reference/cheat-sheet.md
@@ -5,7 +5,7 @@ template = "doc.html"
 aliases = ["docs/reference/cheat-sheet/"]
 +++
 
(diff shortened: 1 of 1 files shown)


syzygyzer:
I would include `$|` as well with the description "structure that satisfies a validator". It's not listed in the runes page yet but the #884 is pretty much done, I just need to talk to someone to clear up some of my confusion about the type system.

Also the argument `model` is used in a number of places. Is this an old term for `spec`? `spec` is also an argument for a few runes here. If it actually means `spec` then I would change those over as well.

Similar concerns for `toga`, `taco`, and `foot`.

baudtack:
It can probably use a second round of updates but the current state is more updated than previously.

PR #921, opened 2020-07-28
issue #681

baudtack:


rmariani:
This should really be an entry here: https://github.com/urbit/support/wiki

This document is being ported over soon.

PR #927, opened 2020-08-06
issue #806

baudtack:
updated mold descriptions

### DIFF
--- a/tutorials/hoon/molds.md
+++ b/tutorials/hoon/molds.md
@@ -8,7 +8,51 @@ aliases = ["/docs/learn/hoon/hoon-tutorial/molds/"]
 A mold is an idempotent function that coerces a [noun](/docs/glossary/noun/) to
 be of a specific type or crashes.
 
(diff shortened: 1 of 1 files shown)


tutorials/hoon/molds.md
```diff
@@ -8,7 +8,51 @@ aliases = ["/docs/learn/hoon/hoon-tutorial/molds/"]
...
+;;(height [5 11])
+::  produces [feet=5 inches=11]
+```
+```hoon
+;;([feet=@ud inches=@ud] [5 11])
+::  produces [feet=5 inches=11]
+```
+
+
+Without using `;;`, you would have to start the mold definition with a comma in order for it to be parsed as a mold, not just as a pair of gates with faces on them:
```
syzygyzer, line 42: I think we should say com instead of comma.

Also I have no idea what the com below is doing. It doesn't appear on the irregular syntax page or the page on `;` runes, and I'm pretty sure its not mentioned elsewhere in Hoon school. So we should either explain it if it can be done briefly without being too much of a distraction, or perhaps elide the discussion on com entirely if its too far out of scope and something that should be avoided anyways. I'm not sure which without knowing how com is used in general.

PR #936, opened 2020-08-24
parsing documentation - mark's review

syzygyzer:
The first parsing doc PR was merged prematurely, so this is a new one with edits suggested by @Fang- (sorry I wasn't sure of a clean way to do this and maintain your other review).


tutorials/arvo/parsing.md
```diff
@@ -24,37 +24,38 @@ Parsing a string is rarely done all at once. Instead, it is usually done
...
 returns the integer `1`. However, taking into account what was said above about
 parsers returning the unparsed portion of the string as well, we should
 represent this return as a tuple. So we should expect something like this:
 ```
-> parse(integer, "1")
+> parse("1", integer)
 (1, "")
-> parse(integer, "123")
+> parse("123", integer)
 (1, "23")
```
Fang-, line 51: I suggested changing the pseudo-code to look more like hoon (ie `(parse "1" integer)`, `[1 ""]` etc). You gave the comment a :+1:. Is the idea here that you don't want to get it confused for something you might try running?

tutorials/arvo/parsing.md
```diff
@@ -164,29 +163,31 @@ of the input `nail`.
...
-parser has not advanced since the parse failed. `q.edg` is null, indicating that
-the parse has failed.
+parser has not advanced since parsing failed. `q.edg` is null, indicating that
+parsing has failed.
 
 Later we will use [+star](#star) to string together a sequence of `+just`s in
-order to parse multiple characters at once.
+order to parse multiple characters at once. However, it should be noted that
+`+jest`, which is covered next, ought to be used in favor of `+just`, which is
+mostly included here for pedagogical reasons.
```
Fang-, line 184: As per discussion on the comment in the previous PR, not sure if the disclaimer here is warranted.

tutorials/arvo/parsing.md
```diff
@@ -236,7 +237,7 @@ and returns a `rule`.
 
 `+cold` is a `rule` builder that takes in a constant noun we'll call `cus` and a
 `rule` we'll call `sef` and returns a `rule` identical to the `sef` except it
```
Fang-, line 239:
```suggestion
`+cold` is a `rule` builder that takes in a constant noun we'll call `cus` and a
`rule` we'll call `sef`. It returns a `rule` identical to the `sef` except it
```

Maybe this applies elsewhere also.

tutorials/arvo/parsing.md
```diff
@@ -268,13 +269,15 @@ themselves. Thus we have several functions for exactly this purpose.
 
 These functions take in either a `tape` or an `@` (typically representing a
 `cord`) with a `rule` and attempts to parse the input with the `rule`. If the
```
Fang-, line 271:
```suggestion
These functions take in either a `tape` or a `cord`,
alongside a `rule`, and attempt to parse the input with the `rule`. If the
```

I know they technically take a `@`, but doing parsing on atoms that aren't `@t` (or "deeper", `@tas` etc) is unlikely to be sane.

tutorials/arvo/parsing.md
```diff
@@ -475,12 +481,17 @@ as the bunt value of the type that the `rule` produces, as well as a gate that
...
 in the `rule` in a manner acceptable to the compiler. Thus the preferred manner
 to write the above `rule` is as follows:
+
 ```hoon
-|-(;~(plug prn ;~(pose (knee *tape |.(^$)) (easy ~))))
+++  pars
+  |-
+  ;~  plug  prn
+    ;~  pose  knee  *tape
+      |.  ^$  (easy ~)
```
Fang-, line 490:
```suggestion
++  pars
  |-
  ;~  plug  prn
    ;~  pose
      (knee *tape |.(^$))
      (easy ~)
    ==
  ==
```

btw, I think it's more common to wrap `+knee` around the parser on the outside (ie, directly under the `|-` here, and as we also do below). This has the helpful effect of specifying a parsing result type early on. I don't think it technically matters much though.

tutorials/arvo/parsing.md
```diff
@@ -505,31 +516,28 @@ First lets look at the code we're going to use, and then dive into explaining
 it. If you'd like to follow along, save the following as `expr-parse.hoon` in
 your `gen/` folder.
 ```hoon
-:: This generator parses arithmetic expressions
+:: expr-parse: parse arithmetic expressions
```
Fang-, line 519:
```suggestion
::  expr-parse: parse arithmetic expressions
```

tutorials/arvo/parsing.md
```diff
@@ -545,7 +553,7 @@ for expressions.
 
 ```hoon
 ++  factor
-  %+  knee  *int
+  %+  knee  @ud
   |.  ~+  ;~  pose
```
Fang-, line 557:
```suggestion
  |.  ~+
  ;~  pose
```

to match the above.

tutorials/arvo/parsing.md
```diff
@@ -505,31 +516,28 @@ First lets look at the code we're going to use, and then dive into explaining
...
-++  int  @ud
-++  print
-  |=  [message=tape]
-  (scan message expr)
+|=  xprs=tape
+|^  (scan xprs tape)
 ++  factor
-  %+  knee  *int
-  |.  ~+  ;~  pose
+  %+  knee  @ud
```
Fang-, line 524:
```suggestion
  %+  knee  *@ud
```

This goes for all of these instances.

tutorials/arvo/parsing.md
```diff
@@ -562,26 +570,24 @@ with `+pose`, which says to try each rule in succession until one of them works.
 
 Since expressions ultimately reduce to factors, we are actually building a
 recursive rule. Thus we need to make use of `+knee`. The first argument for
-`+knee` is `*int`, since our final answer should be a `@ud`, to which we have
-assigned the macro `int`.
+`+knee` is `@ud`, since our final answer should be a `@ud`.
```
Fang-, line 573:
```suggestion
`+knee` is `*@ud`, since our final answer should be a `@ud`.
```

Fang- (changes requested):
Changes mostly look good. Some small errors, some suggestions.

Also, `+next` might be a useful primitive to include. It just consumes and produces a character, regardless of what it is. As such, `(star next)` parses the entire input, and returns it as a `tape`.

tutorials/arvo/parsing.md
syzygyzer: I thought you were just saying that I should mention that it is pseudocode. I think it definitely makes sense to change the pseudocode to be more Hoon-like though.

tutorials/arvo/parsing.md
syzygyzer: Yeah that makes sense. I added it in supposing that you had knowledge that `jest` was actually deprecated regardless of what the comment said.

tutorials/arvo/parsing.md
```diff
@@ -24,38 +24,43 @@ Parsing a string is rarely done all at once. Instead, it is usually done
...
 ```
 What if we wish to parse the rest of the string? We would need to apply the
-`parse(integer, -)` function again:
+`parse` function again:
 ```
-> parse(integer, parse(integer, "123"))
+> parse(parse("123", integer), integer)
 (12, "3")
-> parse(integer, parse(integer, parse(integer, "123")))
+> parse(parse(parse("123", integer), integer), integer)
```
Fang-, line 58: These two don't match the rest.

Fang- (approved):
One style thing, but good once that's resolved!

PR #939, opened 2020-08-27
hoon section reorganization

syzygyzer:
This addresses #938

All Hoon school material has been moved to /docs/tutorials/hoon/hoon-school

parsing.md and cli-tutorial.md have been moved from /docs/tutorials/arvo to
/docs/tutorials/hoon

hoon-errors.md has been moved from /docs/tutorials to /docs/tutorials/hoon

Links to all of these documents have been amended. Hopefully this was done
successfully. The site builds locally, at least, but I had trouble getting emacs to do
this last night and ended up finding another way to do it today, still using
emacs.

The site does not build without also fixing the links in urbit.org. I will be
uploading a PR there shortly https://github.com/urbit/urbit.org/pull/467. I'm not sure exactly what needs to be done here to
get this working live. My expectation is merging this PR and then the urbit.org
one will pass the checks on the links and that will be all there is to it, but
I'm tagging @matildepark as a heads up since I don't know the details of this
process.

----

#


syzygyzer:
Whoops, not sure what I did on my end that caused that miss. In any case, I like the idea of the workbook being separated out from the Hoon school folder. Thanks!

PR #943, opened 2020-09-01
Ames packet structure

syzygyzer:
Just a short section on how Ames packets are structured as serial data. I will
be adding more to the Ames section over the next few weeks and this is
just a start.

I'll be focusing on information useful to the security auditors, which will
probably end up being close to having a story of the "journey of a packet" here,
which I think is what the Ames page used to have before that was made out of date.

----

#

### DIFF
--- a/tutorials/arvo/ames.md
+++ b/tutorials/arvo/ames.md
@@ -132,8 +132,7 @@ The 32-bit header is given by the following data:
  - 1 bit: whether the packet is encrypted or not,
  - 4 bits: unused.
  
- Every packet between Azimuth addresses are encrypted. The only unencrypted
- packets are self-signed attestation packets from 128-bit comets.
+ Every packet sent between `@p`'s is encrypted except for self-signed attestation packets from 128-bit comets.
  
 ### Body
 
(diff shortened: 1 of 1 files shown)


tutorials/arvo/ames.md
```diff
@@ -111,3 +111,42 @@ ducts as unique keys.  Stop tracking a peer by sending Ames a
...
+
+ - 3 bits: Ames `protocol-version`,
+ - 20 bits: a checksum as a truncated insecure hash of the body, done with
+   `+mug`,
+ - 2 bits: the bit width of the sender address encoded as a 2-bit enum,
+ - 2 bits: the bit width of the receiver address encoded as a 2-bit enum,
+ - 1 bit: whether the packet is encrypted or not,
+ - 4 bits: unused.
+ 
+ Every packet between Azimuth addresses are encrypted. The only unencrypted
```
Fang-, line 135: > Azimuth addresses

Aren't we saying "Urbit IDs" now? Especially in these kinds of contexts.

(Also you say "address" in this section, but "`@p`" below.

syzygyzer: I changed it to @p https://github.com/urbit/docs/pull/943/commits/48c4209a12e4f4c57bc97871e7e028b919c56fec. I'm curious though, are comets considered to be Urbit IDs since they aren't on-chain?

syzygyzer: Actually, the pedant in me is bothered since @p is a data type, not the ship itself, so now I changed it to ships :P

belisarius222 (commented):
LGTM

PR #945, opened 2020-09-11
%plea triggers %send which are in response to %born

syzygyzer:
This PR amends the Ames API doc to explain how `%plea` tasks ultimately end up
triggering Unix to send packets, which is via `%send` gifts in response to the
`%born` task.

----

#


syzygyzer:
ping @belisarius222 - according to my notes you actually told me in a conversation several weeks ago that %send is actually not always a gift in response to %born but it was besides the point of whatever we were talking about so I don't have the details

baudtack (approved):

PR #954, opened 2020-10-05
glossary: updated OTA with useful commands

tylershuster:
Given that people often ask questions along these lines, I thought this would be a good place to put them.

### DIFF
--- a/glossary/ota-updates.md
+++ b/glossary/ota-updates.md
@@ -7,3 +7,9 @@ category = "arvo"
 +++
 
 [Ships](../ship) have the ability to upgrade themselves **over the air**. They receive updates from a sponsor [star](../star) or [galaxy](../galaxy).
+
+By default, you will automatically receive updates (OTAs) from your sponsor. To check your OTA source, run `|ota` in the [dojo](../dojo).
+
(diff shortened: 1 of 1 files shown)


syzygyzer:
While this information is definitely valuable, I'm not totally sure that the glossary is the appropriate place for troubleshooting and specific dojo commands. From my recollection, that was not conceived of as a purpose for the glossary. So I worry that putting such information in there muddies the purpose of the glossary as well as misdirects people from where they should be looking for troubleshooting information. The more distributed troubleshooting information is the more difficult it is to keep it all up to date, and we already have a problem with troubleshooting information being distributed (though it has been moving in the right direction). This is probably a losing battle on the large scale, but at least urbit.org can be managed.

That is just my hot take though and I think this requires more careful thought. E.g. I can imagine a future where the glossary is as far as an ordinary user needs to go in terms of learning Urbit, in which case it might make sense to include some troubleshooting information there.

tylershuster:
I agree with you.

I just don't know where this *should* go. I picked this location because someone on the network linked to it while trying to figure out this information.

The larger conversation seems to be that urbit.org needs to have a clear distinction between "I just want to talk with my friends" people and "I'd like to hack core" people. This is somewhere in the middle.

syzygyzer:
I suppose I should have updated my "hot take" once I had thought about it a bit more. Better late than never!

I think this information should go here: https://urbit.org/using/operations/using-your-ship/ I would suggest under the Landscape heading, and perhaps linking to the OTA glossary entry.

PR #960, opened 2020-10-30
removes section on not using ctrl-z

syzygyzer:
From comments in UC - there is nothing wrong with using Ctrl-Z and so the
corresponding section in `ship-troubleshooting.md` has been removed.

----

#


Fang- (approved):
I said I'd :+1: the PR, so I'm :+1:ing the PR.

As discussed, doing these things really shouldn't lead to serious trouble, so deletion approved!

syzygyzer:
I'm :+1:ing you :+1:

PR #963, opened 2020-11-06
add threads.txt to arvo/gall.md

syzygyzer:
Per @philipcmonk 's request #962 , I've added threads.txt to
the Gall page under `arvo/gall.md`. My only edits were adding in a couple links
and tagging code blocks with `hoon`.

----

#


philipcmonk (approved):
Thanks!

PR #966, opened 2020-11-09
micfas irregular syntax

drbeefsupreme:
Following up on @joemfb commenting that `:/` is irregular syntax for `;/`: https://github.com/urbit/docs/pull/890

----

#

### DIFF
--- a/reference/hoon-expressions/irregular.md
+++ b/reference/hoon-expressions/irregular.md
@@ -43,6 +43,14 @@ Miscellaneous useful macros.
 R: `;:(p q)`
 I: `:(p q)`
 
+### ;/ micfas
+
+[docs](@/docs/reference/hoon-expressions/rune/mic.md#micfas) \\/
+`[%mcnt p=hoon]`: tape as XML element.
+
+R: `;/  p`
+I: `:/p`
+
 ## : col (cells)
 
 The cell runes.
--- a/reference/hoon-expressions/rune/mic.md
+++ b/reference/hoon-expressions/rune/mic.md
@@ -172,6 +172,12 @@ ford: %ride failed to execute:
 ~[%$ ~[%$ 'p']]
 ```
 
+##### Syntax
+
+Regular: **1-fixed**
+
+Irregular: `;/  a` is `:/a`
+
 ##### Examples
 
 ```


joemfb:
The `:/` part looks correct, but `;/` is more general than "tape as XML element". I've never used Sail, and am not sure how to characterize it.

@ixv, what say you?

ixv:
Not sure how its more general, isn't it more specific? Its just the inner text of an xml tag right?

joemfb:
You're right, when used with a tape, this constructs "XML cdata", (ie, a text node):

```
> `mars`:/"foo"
[t=[n=%$ a=[i=[n=%$ v="foo"] t=~]] c=~]
```

But the hoon syntax is more general than that:

```
> :/1
[[%~. [%~. 1] ~] ~]
```

syzygyzer:
I really don't have a clue of what is going on in either of those snippets @joemfb 

```
> `mars`:/"foo"
[t=[n=%$ a=[i=[n=%$ v="foo"] t=~]] c=~]
```

I see from `zuse` that
```
  +$  mars  [a=mark b=mark]
```
Yet I do not see this structure reflected in the output. I find this especially confusing since Hoon is structurally typed.

I'm even more confused by this:
```
> :/1
[[%~. [%~. 1] ~] ~]
```
Is `%~.` a `@tas`? Or is it `%~` being called with the door `.`? Or something else?

I also don't know anything about XML other than "its a markup language", which might be contributing to my confusion, but my lack of understanding seems deeper than just that.

I suppose it would help if I knew what `%$` was, but if its a rune then its undocumented, and if its a `@tas` then I'm not sure what its for.

ohAitch:
It's the empty @tas, cf ++$

baudtack:
@joemfb Can you look at @drbeefsupreme 's questions here?

joemfb:
The `mars` you saw in %zuse (now in %lull) is unrelated, the mars in question is in hoon.hoon:

```
+$  mars  [t=[n=%$ a=[i=[n=%$ v=tape] t=~]] c=~]        ::  XML cdata
```

As for the atom literals -- you can prepend `%` to any atom to make it a constant (`%~zod`, `%1`, &c), this is in addition to the `%foo` constant `@tas` syntax. To see the aura, just remove the `%` and use `?` in dojo:

```
> ? %~.
  %~.
%~.
> ? ~.
  @ta
~.
> `@`~.
0
```

So `~.` is the empty `@ta`, just like `%$` is the empty `@tas` and `''` is the empty `@t`.

---

I've never written much sail, so documenting it would be an exercise in reverse engineering the compiler. That's fine, but  we're also not really using it for anything at the moment. At best, it seems like a low priority. It might be best to remove or stub out the documentation now, and reassess in the context of the development of the language itself.

drbeefsupreme:
Since Sail is not being used, I'm going to close this and open a new PR to remove this rune from the mic page.

PR #973, opened 2020-11-16
landscape arch and graphstore docs

baudtack:
Initial documentation for graph store
Initial documentation for basic landscape architecture to orient userspace developers

Intent is to provide outside developers a clear path for interacting with the standard suite of urbit userspace apps that are backed by graph store

### DIFF
--- a/tutorials/concepts/arch.md
+++ b/tutorials/concepts/arch.md
@@ -1,0 +1,21 @@
++++
+title = "Landscape Architecture"
+weight = 4
+template = "doc.html"
++++
+
+# Landscape Architecture
+
+Landscape is designed using a particular userspace architecture model known as store/hook/thread. This model divides applications into multiple parts, two gall applications and threads.
+
+## Stores
+
+Stores are the first of the two gall models to form this architecture. A store is intended to maintain all the state required for an application or set of applications. Stores are intended to be only accessible by the ship it's running on. While this is not enforced, it is the convention. The primary example of a store is `graph-store` which can be found at `app/graph-store`
+
+## Hooks
+
+Hooks are the second gall application used in this architecture and are not always present. Hooks are used for inter-ship communication. Anything that needs to speak to a store on another ship does so through a hook. This allows a separation of access control concerns and a division between those things which are able to be done to a remove ship and those which may be performed by the host. `graph-store` in particular has two hooks `graph-pull-hook` for outgoing requests, and `graph-push-hook` for incoming requests.
+
+## Thread
+
+Threads are used when you want to have a series of transactions to a store where the order of those transactions matters. No guarantees are made about what order things process in for individual requests to a gall application and threads are intended to address this. Threads do not maintain their own state while inactive and are instead run only upon request. A good example is `ted/graph/create`
--- a/tutorials/concepts/arch.md
+++ b/tutorials/concepts/arch.md
@@ -6,11 +6,11 @@ template = "doc.html"
 
 # Landscape Architecture
 
-Landscape is designed using a particular userspace architecture model known as store/hook/thread. This model divides applications into multiple parts, two gall applications and threads.
+Landscape is designed using a particular userspace architecture mode consisting of stores, hooks, and threads.
 
 ## Stores
 
-Stores are the first of the two gall models to form this architecture. A store is intended to maintain all the state required for an application or set of applications. Stores are intended to be only accessible by the ship it's running on. A store may not send pokes, peeks, or watches. A store should be treated and thought of as a database table. While this is not enforced, it is the convention. The primary example of a store is `graph-store` which can be found at `app/graph-store`
+Stores are the first of the two gall models to form this architecture. A store is intended to maintain all the state required for an application or set of applications. Stores are intended to be only accessible by the ship it's running on. A store may not send pokes, peeks, or watches. A store should be treated and thought of as a database. While this is not enforced, it is the convention. The primary example of a store is `graph-store` which can be found at `app/graph-store`
 
 ## Hooks
 
@@ -18,4 +18,4 @@ Hooks are the second gall application used in this architecture and are not alwa
 
 ## Thread
 
-Threads are used when you want to have a series of transactions to a store where the order of those transactions matters. No guarantees are made about what order things process in for individual requests to a gall application and threads are intended to address this. Threads do not maintain their own state while inactive and are instead run only upon request. A good example is `ted/graph/create`
+Threads are used when you want to have a series of transactions on a store where the order of those transactions matters. No guarantees are made about the order in which individual requests to a gall applications are processed; threads were introduced to address this. Threads do not maintain their own state while inactive and are instead run only upon request. A good example is `ted/graph/create`
--- a/tutorials/concepts/arch.md
+++ b/tutorials/concepts/arch.md
@@ -10,11 +10,11 @@ Landscape is designed using a particular userspace architecture model known as s
 
 ## Stores
 
-Stores are the first of the two gall models to form this architecture. A store is intended to maintain all the state required for an application or set of applications. Stores are intended to be only accessible by the ship it's running on. While this is not enforced, it is the convention. The primary example of a store is `graph-store` which can be found at `app/graph-store`
+Stores are the first of the two gall models to form this architecture. A store is intended to maintain all the state required for an application or set of applications. Stores are intended to be only accessible by the ship it's running on. A store may not send pokes, peeks, or watches. A store should be treated and thought of as a database table. While this is not enforced, it is the convention. The primary example of a store is `graph-store` which can be found at `app/graph-store`
 
 ## Hooks
 
-Hooks are the second gall application used in this architecture and are not always present. Hooks are used for inter-ship communication. Anything that needs to speak to a store on another ship does so through a hook. This allows a separation of access control concerns and a division between those things which are able to be done to a remove ship and those which may be performed by the host. `graph-store` in particular has two hooks `graph-pull-hook` for outgoing requests, and `graph-push-hook` for incoming requests.
+Hooks are the second gall application used in this architecture and are not always present. Hooks are used for inter-app communication. This allows a separation of access control concerns and a division between those things which are able to be done by a store itself and those done by another gall application. `graph-store` in particular has two hooks `graph-pull-hook` for outgoing requests, and `graph-push-hook` for incoming requests.
 
 ## Thread
 
--- a/tutorials/hoon/graph.md
+++ b/tutorials/hoon/graph.md
@@ -6,7 +6,7 @@ template = "doc.html"
 
 # What is Graph Store?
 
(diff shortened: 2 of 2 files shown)


tutorials/concepts/arch.md
```diff
@@ -0,0 +1,21 @@
...
+template = "doc.html"
++++
+
+# Landscape Architecture
+
+Landscape is designed using a particular userspace architecture model known as store/hook/thread. This model divides applications into multiple parts, two gall applications and threads.
+
+## Stores
+
+Stores are the first of the two gall models to form this architecture. A store is intended to maintain all the state required for an application or set of applications. Stores are intended to be only accessible by the ship it's running on. While this is not enforced, it is the convention. The primary example of a store is `graph-store` which can be found at `app/graph-store`
```
tacryt-socryp, line 13: Add:

“A store may not send pokes, peeks, or watches. A store should be treated and thought of as a database table.”

tutorials/concepts/arch.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Hooks
+
+Hooks are the second gall application used in this architecture and are not always present. Hooks are used for inter-ship communication. Anything that needs to speak to a store on another ship does so through a hook. This allows a separation of access control concerns and a division between those things which are able to be done to a remove ship and those which may be performed by the host. `graph-store` in particular has two hooks `graph-pull-hook` for outgoing requests, and `graph-push-hook` for incoming requests.
```
liam-fitzgerald, line 17: Should be "Hooks are used for inter-app communication", see `hark-graph-hook` for an example of this.

tutorials/concepts/arch.md
```diff
@@ -0,0 +1,21 @@
++++
+title = "Landscape Architecture"
+weight = 4
+template = "doc.html"
++++
+
+# Landscape Architecture
+
+Landscape is designed using a particular userspace architecture model known as store/hook/thread. This model divides applications into multiple parts, two gall applications and threads.
```
syzygyzer, line 9: > Landscape is designed using a particular userspace architecture model known as store/hook/thread.

The name of the model is "store/hook/thread"? Or is that what the model consists of?

> This model divides applications into multiple parts, two gall applications and threads.

This sentence doesn't parse in my head. Are the multiple parts the gall applications and threads? Or are there multiple parts, two gall applications, and threads, or is "gall applications and threads" a single thing or should there be an Oxford comma?

tutorials/concepts/arch.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Stores
+
+Stores are the first of the two gall models to form this architecture. A store is intended to maintain all the state required for an application or set of applications. Stores are intended to be only accessible by the ship it's running on. A store may not send pokes, peeks, or watches. A store should be treated and thought of as a database table. While this is not enforced, it is the convention. The primary example of a store is `graph-store` which can be found at `app/graph-store`
```
syzygyzer, line 13: > thought of as a database table.

A single table doesn't seem right. E.g. arbitrary graph database cannot be represented with a single table. This should probably just say database or database schema.

tutorials/concepts/arch.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Hooks
+
+Hooks are the second gall application used in this architecture and are not always present. Hooks are used for inter-app communication. This allows a separation of access control concerns and a division between those things which are able to be done by a store itself and those done by another gall application. `graph-store` in particular has two hooks `graph-pull-hook` for outgoing requests, and `graph-push-hook` for incoming requests.
+
+## Thread
+
+Threads are used when you want to have a series of transactions to a store where the order of those transactions matters. No guarantees are made about what order things process in for individual requests to a gall application and threads are intended to address this. Threads do not maintain their own state while inactive and are instead run only upon request. A good example is `ted/graph/create`
```
syzygyzer, line 21: > Threads are used when you want to have a series of transactions to a store where the order of those transactions matters.

Transactions in the context of databases are mappings between instances of database schemas. Here a store is a database schema, and its data at a moment in time is an instance of that schema. So it would make more sense to say "transactions on a store" (or maybe "of a store"?) rather than "transactions to a store", which makes it sound like a transformation between schemas rather than between schema instances.

> No guarantees are made about what order things process in for individual requests to a gall application and threads are intended to address this.

The grammar is a little confusing - maybe "No guarantees are made about the order in which individual requests to a gall applications are processed; threads were introduced to address this."

tutorials/hoon/graph.md
```diff
@@ -0,0 +1,153 @@
++++
+title = "Graph Store"
+weight = 3
+template = "doc.html"
++++
+
+# What is Graph Store?
+
+Graph store is a gall implementation of a tree-based database that has an affordance for cyclic structures by allowing nodes to reference other nodes. It is most comparable to modern NoSQL databases in structure, though it allows for schemas to be written in the form of marks. It is used as the backend for an increasing number of Landscape applications, and is considered to be best used for social data.
```
syzygyzer, line 9: > Graph store is a gall implementation of a tree-based database that has an affordance for cyclic structures by allowing nodes to reference other nodes.

"Graph store" should probably either be `graph-store` or Graph Store so that its a proper noun.

I don't actually know how `graph-store` works yet, but saying it is "tree-based" and allows cyclic structures is contradictory. By definition, a tree is a connected acyclic graph. So something is off here.

tutorials/hoon/graph.md
```diff
@@ -0,0 +1,153 @@
...
+
+`marked-graph` is a pair of `graph` and `(unit mark)`. The semantics of a graph are defined by the mark. This is just for validating the structure of the graph. 
+
+`graph` is a `mop`, a mold builder for order map, of an atom and a `node`.
+
+`node` is a pair of `post` and `children`.
+
+`post` is defined in `sur/post`. It contains:
+
+`author=ship`
```
syzygyzer, line 25: I think just putting the Hoon here would be nicer

```hoon
+$  post
  $:  author=ship
      =index
      time-sent=time
      contents=(list content)
      hash=(unit hash)
      =signatures
  ==
```

tutorials/hoon/graph.md
```diff
@@ -0,0 +1,153 @@
...
+
+When interacting with `graph-store` you will need to create an `update`
+
+`update` is a tagged union that currently only has one option
+
+`[%0 p=time q=update-0]`
+
+The reason this is a union is to serve as future proofing for version changes.
+
+An `update-0` can either be logged in the case of adding and removing nodes or signatures or non logged as in the case of adding or removing a graph. This logging refers to if this action adds an entry to the update log that is used to synchronize state between ships.
```
syzygyzer, line 48: > An `update-0` can either be logged in the case of adding and removing nodes or signatures or non logged as in the case of adding or removing a graph. 

Some punctuation
"An `update-0` can either be logged, as in the case of adding/removing nodes or signatures, or non-logged, as in the case of adding/removing a graph. "

tutorials/hoon/graph.md
```diff
@@ -0,0 +1,153 @@
...
+
+Below is an excerpt or `sur/graph-store.hoon` at the time of writing showing the definition of `logged-update-0` and `update-0` which should now be readable.
```
syzygyzer, line 50: Below is an excerpt or `sur/graph-store.hoon` at the time of writing, showing

tutorials/hoon/graph.md
```diff
@@ -0,0 +1,153 @@
...
+
+```
```
syzygyzer, line 52: add hoon tag

tutorials/hoon/graph.md
```diff
@@ -0,0 +1,153 @@
...
+      [%keys =resources]
+      [%tags tags=(set term)]
+      [%tag-queries =tag-queries]
+  ==
+```
+
+
+## Pokes
+
+Graph store can be poked with a `graph-update` which can:
```
syzygyzer, line 83: `graph-store` or Graph Store

> with a `graph-update` which can:

comma after `graph-update`

tutorials/hoon/graph.md
```diff
@@ -0,0 +1,153 @@
...
+ 
+ `[%remove-tag =term =resource]`
+ 
+ - archive and unarchive a graph
+ 
+ `[%archive-graph =resource]`
+ 
+ `[%unarchive-graph =resource]`
+
+## Scrys
```
syzygyzer, line 114: The plural of scry is scries, but since scry is a special word for us I'm not sure if that applies here

tutorials/hoon/graph.md
```diff
@@ -0,0 +1,153 @@
...
+ 
+- Fetch the update log by resource
+ `[%x %update-log @ @ ~]`
+ 
+- Fetch the time of the last entry of the update log
+ `﻿﻿[%x %peek-update-log @ @ ~]`
+
+## Marks
+
+There are marks in `mar/graph/validator` that are worth review. They are each short enough to read quickly. They are used to validate the shapes of various nouns to be stored in the graph-store.
```
syzygyzer, line 153: > worth review

worth reviewing

syzygyzer (changes requested):

tutorials/concepts/arch.md
baudtack: > The name of the model is "store/hook/thread"? Or is that what the model consists of?

Those are the parts idk if it has a name that doesn't suck.

baudtack: but it's sort of the same as like Model View Controller so this would be SHT (lol)

tutorials/hoon/graph.md
baudtack: You can have nodes reference other nodes but it's a tree like most nock structures. I don't know what to call that.

tutorials/hoon/graph.md
baudtack: Eh. let's go with scries for now.

tutorials/concepts/arch.md
tacryt-socryp: Perhaps "schema for database tables"? Trying to get across the idea that each store should maintain a single logical "concern"

tutorials/hoon/graph.md
tacryt-socryp: Graph store has virtual cycles (a node may reference any other node, or even multiple). But fundamentally it is storing an ordered map of nodes where each node may have children nodes, etc.

tutorials/concepts/arch.md
syzygyzer: Maybe just say that the model consists of stores, hooks, and threads, explained below.

tutorials/hoon/graph.md
syzygyzer: Yeah that puts you into a bit of a bind with terminology. In that case I'd either call it a "tree with additional virtual edges", or a "finite graph with a chosen spanning tree". I think the latter is more precise but also more confusing to someone without a background in graph theory, and either way some clarification would be needed, since the term "virtual edge" is overloaded in graph theory. So let's go with virtual edges and explain it like @tacryt-socryp

syzygyzer (approved):

PR #976, opened 2020-11-23
fix glossary links from breach page

syzygyzer:
Addresses https://github.com/urbit/docs/pull/948/files/18868f9d3b106093c5076f0a2d159739ed647850

----

#

PR #977, opened 2020-11-23
fix link again

syzygyzer:
See https://github.com/urbit/docs/pull/976

Also mentions how you can transfer without breaching

----

#


matildepark:
L20 needs to be `@/docs/glossary/proxies.md`.

matildepark:
Also this still refers to old, old, old urbit-help.

matildepark (approved):

PR #978, opened 2020-11-24
|code %reset

syzygyzer:
Includes instructions for resetting the code used to log in to Landscape.

Also replaces stale reference to urbit-help channel

----

#

### DIFF
--- a/tutorials/ship-troubleshooting.md
+++ b/tutorials/ship-troubleshooting.md
@@ -68,7 +68,18 @@ So for example, to silence all Ames packet-related errors, try `|knob %hole %hus
 
 A personal breach is when a ship tells all the other ships that have communicated with it to treat it as though the ship was just started for the first time again, since everyone on the network has forgotten about it.
 
-Personal breaches often fix connectivity issues, but should only be used as a last resort. To find out how to perform a personal breach, check out our [Guide to Breaches](../guide-to-breaches). Before taking such a drastic measure, try other methods in this guide. You can also ask for help on `~/~dopzod/urbit-help`, or, failing that, in the `#ship-starting-support` channel in our [Discord server](https://discord.gg/n9xhMdz).
+Personal breaches often fix connectivity issues, but should only be used as a
+last resort. To find out how to perform a personal breach, check out our [Guide
+to Breaches](../guide-to-breaches). Before taking such a drastic measure, try
+other methods in this guide. You can also ask for help on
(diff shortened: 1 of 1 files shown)


tutorials/ship-troubleshooting.md
```diff
@@ -68,7 +68,18 @@ So for example, to silence all Ames packet-related errors, try `|knob %hole %hus
...
+to Breaches](../guide-to-breaches). Before taking such a drastic measure, try
+other methods in this guide. You can also ask for help on
+in the Help channel in the Urbit Community group at `~bitbet-bolbel/urbit-community` or, failing that, in the `#ship-starting-support`
+channel in our [Discord server](https://discord.gg/n9xhMdz).
+
+### Reset `+code`
+
+You login to Landscape using the camel-case phrase obtained from dojo by
+entering `+code`. For security reasons you may wish to change this code
+regularly. You may do so by entering `|code %reset` into dojo.
```
Fang-, line 82: Bridge isn't mentioned here, so maybe we're fine leaving this out altogether, but a caveat to be aware of is that Bridge will no longer be able to derive your `+code` (and/or one-click log you in) after you've run `|code %reset`.

syzygyzer: Ah, I was not aware that +code was also used to log in to Bridge. Let me make sure I understand though - are you saying that once you change your code, you can no longer log in to Bridge this way at all? I definitely need to include that here if so.

Fang-: > +code was also used to log in to Bridge

Oh, that's not what I meant, and that's not the case.

I was saying that, as a convenience, Bridge can give you your +code (so you don't have to go into dojo like a nerd), but that this only works when you have not run `|code %reset` yet.

syzygyzer: Alright, in that case I am fine with leaving mention of Bridge out of this since if you're changing your code then you're already enough of a nerd to be using dojo. Is it even possible to change Bridge so that it can see a reset code? I'm guessing not since it really just interacts with Azimuth, and the new code isn't pushed to Ethereum.

baudtack: I think it is worth mentioning as an aside that bridge wont be able to derive your +code anymore. Guaranteed that will bite someone.  Guaranteed it will bite someone even if it's in here but then we can laugh at them.

baudtack (changes requested):

syzygyzer:
Alright, yeah it does make sense to mention the Bridge thing.

baudtack (approved):

PR #981, opened 2020-11-25
Eyre public API (draft)

drbeefsupreme:
I started on this early this year and got sidetracked and have not yet returned
to it. I'm putting up this pull request now so that it does not get accidentally
lost, but I'm not sure when I'll have time to come back to it. I vaguely recall
that Eyre either had undergone some major changes since I started on this, or
changes are planned, and that was why I set it aside, but I was not yet
organizing my notes well enough at the time for me to be sure.

Just to be clear, this is a draft, nobody should bother reviewing this yet.

----

#


Fang-:
(Aside from some arvo-wide changes this doesn't look terribly outdated.)

syzygyzer:
That's good to know. I got through checking most of it yesterday and only found one change in the semantics of the tasks so far (which might have just been a mistake or incomplete entry).

What arvo-wide changes? I checked the urbit repo yesterday for Eyre pull requests in the last year but there were so many that I didn't start digging into any of them.

Fang-:
Error handling changed, so the `%crud` section here is outdated. `%wegh` is just done through scry now (for `|mass`). The things that are `vane-task`s should probably get their own page/document instead of being discussed for each vane, anyway.

drbeefsupreme:
#1089

PR #984, opened 2020-12-02
moving arvo out of date warning

syzygyzer:
The first part of this document is not out of date and the warning is scary.
This moves the warning to the kernel section.

Per @jalehman 's suggestion

----

#


jalehman (approved):
Many thanks for the alacrity @syzygyzer!

PR #989, opened 2020-12-04
Hoon school - tapes use 24x the memory of cords

drbeefsupreme:
Just a one sentence edit based on a comment by @philipcmonk

----

#

### DIFF
--- a/tutorials/hoon/hoon-school/lists.md
+++ b/tutorials/hoon/hoon-school/lists.md
@@ -94,6 +94,9 @@ You can also use the words `cord` and `tape` for casting:
 'this is a cord'
 ```
 
+`tape`s use 24 times as much memory as `cord`s, and so in scenarios where
+performance is important it is preferable to use `cord`s.
+
 ## List Functions in the Hoon Standard Library
 
 Lists are a commonly used data structure. Accordingly, there are several functions in the [Hoon standard library](/docs/reference/library/) intended to make lists easier to use.


joemfb:
Yes, in u3. But the memory layout of a cell is unspecified, and other runtimes or memory models could cut that ratio substantially (ideally down to 1, but it might not be worth the trouble).

syzygyzer:
Hmmm any suggestion then? Maybe I could just say "uses up to 24x the memory" instead of a precise figure?

joemfb:
"In our runtime", "in, u3, our current runtime", ... Idk, we could honestly use some more precise vocabulary for such things.

But the underlying point is important, and shouldn't be waved away. A cell in arvo/hoon/nock is just the ordered association of two nouns, and a tape is just a cons-list of the bytes of a cord. How much memory a cell uses (6 32-bit words in u3) depends entirely on the memory model of the runtime -- it cannot be an intrinsic property within urbit.

baudtack:
I feel like this is sending the wrong message. You should use tapes when you're planning to manipulate the data. The space constraints aren't something you're going to need to worry about for the most part. While it may be true, talking about this in the lists discussion is going to make people think they shouldn't ever use tapes because they are less memory efficient.

syzygyzer:
> I feel like this is sending the wrong message. You should use tapes when you're planning to manipulate the data. The space constraints aren't something you're going to need to worry about for the most part. While it may be true, talking about this in the lists discussion is going to make people think they shouldn't ever use tapes because they are less memory efficient.

That's a really good point. I think from that context this discussion should possibly end up in the standard library docs instead, but if I were to do that I think it would be even more sensible to just have a discussion somewhere on how memory usage works in general (perhaps that already exists and I am forgetting).

I'll leave this open for now and think about it over the weekend.

baudtack:
Anymore thoughts on this?

drbeefsupreme:
I've come to the conclusion that including this wouldn't really contribute to understanding without a long diversion, probably a whole lesson, on how memory actually works, which doesn't exist. So I'm closing for now.

PR #990, opened 2020-12-11
updating code on advanced types

drbeefsupreme:
Was reading this page and noticed the code was out of date.

Went through the text as well to make sure I didn't mess up any references.
There was one part though that looked like it was already wrong even before -
starting at L200:

```
If a name is in the q.p map, it's a bridge. 
```
`tune` only had one `map`, and the line commented with `::bridges` is a `(list
hoon)`. Since a `%face` is a `p=$@(term tune) q=type`, I guessed that it meant
the `term` here, but I think I am still a little mistaken. So I need some
guidance here.

Tagging @jtobin since I ran into this while reading compiler documentation,
which we're supposed to work on together after the new year, which will
hopefully be accompanied by some improvements to the docs on the type system.

----

#

### DIFF
--- a/reference/hoon-expressions/advanced.md
+++ b/reference/hoon-expressions/advanced.md
@@ -18,27 +18,29 @@ large functions.  Don't worry about them for a little while.
           $@  $?  %noun
                   %void
               ==
-          $%  [$atom p=term q=(unit @)]
-              [$cell p=type q=type]
-              [$core p=type q=coil]
-              [$face p=$@(term tune) q=type]
-              [$fork p=(set type)]
-              [$hint p=(pair type note) q=type]
-              [$hold p=type q=hoon]
+          $%  [%atom p=term q=(unit @)]
+              [%cell p=type q=type]
+              [%core p=type q=coil]
+              [%face p=$@(term tune) q=type]
+              [%fork p=(set type)]
+              [%hint p=(pair type note) q=type]
+              [%hold p=type q=hoon]
           ==
 ::
-+$  coil  $:  p=[p=(unit term) q=poly r=vair]
-              q=type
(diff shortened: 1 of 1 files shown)


drbeefsupreme:
Any chance to look at this @jtobin ?

reference/hoon-expressions/advanced.md
```diff
@@ -188,13 +191,13 @@ of secret superpowers for hacking the namespace.  Remember that
...
 limb, we just search the type depth-first.
 
-If a name is in the `p.p` map, it's an alias.  (An alias is defined using the
+If a name is in the `p.q.p` `map`, it's an alias.  (An alias is defined using the
 `=*` rune.) The map contains a `(unit hoon)`; if the unit is full, the name
 resolves to that hoon (compiled against the `q` type).  If the unit is empty,
 the name is blocked / skipped (see [limb](@/docs/reference/hoon-expressions/limb/limb.md) for what
 this means).
 
-If a name is in the `q.p` map, it's a bridge.  (A bridge is defined using the
```
jtobin, line 197: Here `q.p` seems correct per my above comment, but as you point out, the result is not a map but a `(list hoon)`.  Perhaps `$tune` used to be implemented like this.

What happens now seems to be similar in spirit to what is described here, but the details are slightly different and more involved, and are probably unimportant for intuition.  See the `%tscm` case in `+play` and note the call to `+busk`.

jtobin (commented):
I think the docs are mostly correct here, per my in-line comments.  As I understand them stated, the original wings correspond to what I'd expect to see.  The details re: bridges differ, but seem to convey "what's going on" better than if we actually tried to describe exactly what happens with `+busk`.

PR #991, opened 2020-12-12
claz guide first draft

baudtack:

Issue #955, opened 2020-10-06
improve master ticket documentation


syzygyzer:
The documentation appears to fully describe what a master ticket is for, but people still end up confused. For example, there was someone in urbit-help just now that bought a planet for an ETH address they already owned, and then set up a master ticket and thought the planet had been stolen from them since it was transferred to a new address. I think we could be a bit more clear in our wording around master tickets, what they are for, and how they are not a requirement but rather a convenience.

syzygyzer:
After #999 I think what this still needs is a glossary entry and a link to that entry from the install page.

Issue #930, opened 2020-08-15
document advanced key custody methods and best practices


syzygyzer:
Per @galenwp's suggestion, we are commonly asked about how to do things like set up m of n signatures, such as m of n shamir keys, so it would be swell to document how to do this in some way.

PR #999, opened 2020-12-18
HD wallet generation, hardware wallets, hot and cold wallets

syzygyzer:
Addresses #998 as well as partially addresses #955 and #930. Will need to be
edited to link `claz` when #991 is merged.

One outstanding question: I am not certain that the `@q` `T` is actually the
master ticket. Is it actually `E`? When you receive an email invite, you get a
~four-word-camel-case phrase to put into Bridge which then generates a different
~four-word-camel-case phrase that is actually your master ticket. Based on my
reading of
https://github.com/urbit/proposals/blob/master/008-urbit-hd-wallet.md, it is
unclear to me whether the invitation ~four-word-camel-case phrase is actually
`T` and the master ticket is `E`, or if the invitation phrase isn't included in
this document at all. @Fang- ?

----

#

### DIFF
--- a/tutorials/concepts/azimuth.md
+++ b/tutorials/concepts/azimuth.md
@@ -101,3 +101,36 @@ ownership, spawn, management, and voting rights affiliated with your identities.
 
 For the technical implementation details, take a look at Azimuth's
 [Github repository](https://github.com/urbit/azimuth).
+
+## Hardware HD wallet
+
+You may also store your Urbit using a hardware HD wallet such as a Trezor or
+Ledger. This option is used instead of a master ticket - it is a totally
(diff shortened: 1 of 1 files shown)


Fang-:
> When you receive an email invite

The invite is a temporary wallet. So, technically, both those phrases are master tickets. (;  
But the first one is temporary (because the one who sent it also knows it), so probably best not to call it such. The second (for which you download the paper collateral etc) is permanent, and is your master ticket in the fullest sense of the word. (This is, also `T` in the diagram. `@q` simply means it's "`@p` notation, but the value isn't scrambled".)

baudtack (approved):

PR #1003, opened 2020-12-27
nock explanation: fixed some formatting in the code block where we su…

urcades:
…rface nock base definition


urcades:
This is literally just a re-arranging of the text ... it's nothing deep/big, but was infuriating enough to make PR!

PR #1008, opened 2021-01-19
Fixed example in parsing.md

risruc-habteb:
example used `lit` and `rit` where `pal` and `par` are proper now.
example previously had a syntax error on line 2:
`|^  (scan xprs tape)` corrected to `|^  (scan xprs expr)`


drbeefsupreme:
LGTM! Thanks!

PR #1009, opened 2021-01-20
tutorial: Deploy a ship to Azure Container Instances

eamsden:
I have put this tutorial through a start-to-finish test-run, deploying a comet to Azure Container Instances and accessing its Landscape by putting its FQDN in my browser.

@jalehman

### DIFF
--- a/tutorials/azure-container-instances.md
+++ b/tutorials/azure-container-instances.md
@@ -1,0 +1,120 @@
++++
+title = "Deploying to an Azure Container Instance"
+weight = 8
+template = "doc.html"
+aliases = ["docs/using/azure-container-instance"]
++++
+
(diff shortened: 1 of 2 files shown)


baudtack:
@vvisigoth do we have an azure account somewhere i can use to verify these instructions?

eamsden:
@baudtack if you create one with a fresh email you get $200 in AWS credits (I don't recall needing any payment information).

baudtack (approved):

jalehman:
@baudtack This should certainly be on the docs, but not here any more. This is usage-oriented content that should be on urbit.org under `/using`, ideally nested within `/using/running` somewhere.

drbeefsupreme:
@eamsden This ended up in the wrong spot because of the reorg. Can you open another PR on the urbit.org repo and throw it into the `using/running` folder? Also make sure to add it to `using/running/_index.md`. Thanks!

PR #1010, opened 2021-01-20
remove modulo and talk glossary entries

drbeefsupreme:
No longer exist. I checked for links to these pages and it seems they were unconnected.


baudtack (approved):
RIP Talk

PR #1015, opened 2021-01-24
Fix typo in dotket documentation.

clonex10100:
The dotket doc entry has a minor typo where `p` should be `q`


baudtack (approved):
Great! Thanks for the contrib!

PR #1023, opened 2021-02-20
Fix hoon school bank account typo

hosted-fornet:
There is an extra "we that" in one of the lines.
Remove to change line from
"...core we that that we examine below"
to
"...core that we examine below".


drbeefsupreme:
Thank you!

PR #1024, opened 2021-02-21
Fix hoon school type checking typo

hosted-fornet:
There is an extra "of" that either should be removed or needs other words added to it to be meaningful. In 80b739a, I remove it.


drbeefsupreme:
Thanks!

PR #1031, opened 2021-03-11
Update clay.md

conradconrad:
Looks like this was intended as a comment?


drbeefsupreme:
Yeah there are a lot of issues with `clay.md`. I've been working on updating it for the past couple of months when I find time. I'm going to close this and merge it with #1030 since that is the first major revision to the Clay docs I'm pushing.

PR #1034, opened 2021-03-15
auras reference

drbeefsupreme:
Added a page with a list of commonly used auras. Up til now, the only reference we had was buried in the Hoon school notes. This was lifted more or less directly from https://github.com/natareo/hooncard/blob/master/hooncard.pdf h/t to @sigilante

### DIFF
--- a/hoon/hoon-school/atoms-auras-and-simple-cell-types.md
+++ b/hoon/hoon-school/atoms-auras-and-simple-cell-types.md
@@ -109,6 +109,8 @@ Aura         Meaning                        Example Literal Syntax
 
 Some of these auras nest under others.  For example, `@u` is for all unsigned auras.  But there are other, more specific auras; `@ub` for unsigned binary numbers, `@ux` for unsigned hexadecimal numbers, etc.
 
+For a more complete list of auras, see [Auras](@/docs/hoon/reference/auras.md).
+
 ## Aura Inference in Hoon
 
 Let's do more examples in the Dojo using the `?` operator.  We'll focus on just the unsigned auras for now:
--- a/hoon/reference/auras.md
+++ b/hoon/reference/auras.md
@@ -52,5 +52,12 @@ two, starting from A.
         @uvJ    unsigned, 512-bit integer (frequently used for entropy)
 ```
 Auras are non-coercive, but conversions may have to go via the empty aura, e.g.
-`^-(@ud ^-(@ 'foo'))`.
-
+```
+> ^-(@ud ^-(@ 'foo'))
+7.303.014
+```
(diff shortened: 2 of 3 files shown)


hoon/hoon-school/atoms-auras-and-simple-cell-types.md
```diff
@@ -109,6 +109,8 @@ Aura         Meaning                        Example Literal Syntax
 
 Some of these auras nest under others.  For example, `@u` is for all unsigned auras.  But there are other, more specific auras; `@ub` for unsigned binary numbers, `@ux` for unsigned hexadecimal numbers, etc.
 
+For a more complete list of auras, see [Auras](@/docs/hoon/reference/auras.md).
+
```
ohAitch, line 113: I would say it is more *discussion* of auras, since the list itself is a copy of the code block above and thereby inherently no more or less complete.

hoon/reference/auras.md
```diff
@@ -0,0 +1,56 @@
...
+
+```
+        @ubD    signed single-byte (8-bit) decimal
+        @tD     8-bit ASCII text
+        @rhE    half-precision (16-bit) floating-point number
+        @uxG    unsigned 64-bit hexadecimal
+        @uvJ    unsigned, 512-bit integer (frequently used for entropy)
+```
+Auras are non-coercive, but conversions may have to go via the empty aura, e.g.
+`^-(@ud ^-(@ 'foo'))`.
```
ohAitch, line 55: Maybe add a note about irregular ^- form `` `@foo`bar`` implicitly doing this

hoon/hoon-school/atoms-auras-and-simple-cell-types.md
drbeefsupreme: It's not a copy - there are auras listed in this PR that are not in the Hoon school notes. It is an extension and thus more complete.

I did not edit the Hoon school notes to include the additional auras in this PR for pedagogical reasons - when someone is learning something new, its not necessarily a good idea to include all associated complexity right off the bat. But I would be willing to consider that that might not apply here, or that some of these additional auras should be listed in the Hoon school notes.

hoon/reference/auras.md
```diff
@@ -0,0 +1,63 @@
++++
+title = "Auras"
+weight = 70
+template = "doc.html"
++++
+
+```
```
jalehman, line 7: A brief bit of copy explaining what auras are would probably be nice. A backlink to the hoon school explanation would be a good idea too.

jalehman (approved):
One minor comment about helping contextualize the reader of the reference doc, but otherwise looks good. Thanks for adding this!

PR #1035, opened 2021-03-16
aura nesting fix, bunting

drbeefsupreme:
Thanks to ~sigryn-habrex for these suggestions.


baudtack (approved):
Not sure if it would be helpful to anyone but all bunts being 0 is the reason for loobeans, a thing that only makes any sense at all in the context of bunts.

PR #1014, opened 2021-01-21
Add anchor links management script and and set all tutorials to show anchor links by default

ynx0:
This PR is a supporting PR for to https://github.com/urbit/urbit.org/pull/551, **Add Anchor Links**.
It adds some boilerplate to allow for anchor links, links which appear next to all headings in a document (i.e. h1, h2, h3, ...) and allow the url to specific a heading to jump to.
This PR adds the setting to the all sections under `tutorials` and adds a script (`anchors.py`) to programatically manage anchor links, inspired by the [`weights.sh`](https://github.com/urbit/docs/blob/a076606755015af0c8b914fec66086fb4cae78ac/weights.sh) script.

This pr should be merged only after merging **Add Anchor Links**

PR #1042, opened 2021-03-20
Fix architecture.md to allow site to build

ynx0:
Discussion from #1014 
cc @drbeefsupreme

### DIFF
--- a/arvo/clay/architecture.md
+++ b/arvo/clay/architecture.md
@@ -1,10 +1,10 @@
-++
++++
 title = "Architecture"
 weight = 2
 template = "doc.html"
-++
++++
 
-Clay is the primary filesystem for the [Arvo](@/docs/Arvo/overview.md) operating system,
+Clay is the primary filesystem for the [Arvo](@/docs/arvo/overview.md) operating system,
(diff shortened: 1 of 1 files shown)


drbeefsupreme:
Thanks @ynx0 I think I actually made this mistake and didn't notice, then assumed it was your patch, I apologize! Thanks for fixing my mistake.

Now I am trying to merge your pull request again, but I think since the commits are already part of the tree, it says it will make no changes because of the reverts in the future... I am not quite sure what to do in this scenario. Perhaps if you create a new empty commit and then make the pull request, it will allow your changes to go through. But I am a git novice. @baudtack probably knows what to do

ynx0:
@drbeefsupreme just looked up what to do in this exact scenario and someone on [stackoverflow](https://stackoverflow.com/a/28497372) suggests to revert the revert as a valid strategy :stuck_out_tongue:.

ohAitch:
Yeah I suspect if you don't revert the revert and merge, the revert stands

drbeefsupreme:
That was it. Thanks! Sorry for the mess :)

Issue #1032, opened 2021-03-12
document ;<


drbeefsupreme:
It's used often and in many important contexts.

PR #1044, opened 2021-03-26
micgal entry

drbeefsupreme:
Adds an entry for `;<` to the rune docs. Addresses #1032

h/t to @juped and @sigilante whose discussion in the Hoon school chat I lifted
much of this from.

Possible outstanding issues:
 - Not sure that my example combining `+biff` and `+bind` is clear, given how
   `+bind` seems to treat gates that produce units as though they do not,
   leading to swapping their order in the sequence being commutative.
 - Many runes have an "expands to" section, but some of them do not. Does this
   rune expand? How can I tell? I got the desugaring from looking at the parser
   in `hoon.hoon`, but I don't know where to look for "expands to".

### DIFF
--- a/hoon/reference/rune/mic.md
+++ b/hoon/reference/rune/mic.md
@@ -80,11 +80,10 @@ implementation of monadic bind. This rune takes a gate `bind` which takes a mold
 ##### Desugaring
 
 ```hoon
-%+
-  (bind ^:mold)
-expr1
+%+  (bind mold)
+  expr1
 |=  mold
-  expr2
+expr2
 ```
 
 ##### Discussion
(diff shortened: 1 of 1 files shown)


juped:
biff is monadic bind for units, but 'bind' lifts functions to functions taking and producing units (it would be fmap in haskell) - it is not a monadic bind. i apologize for never fixing this stupid troll name.

`;<  mold  bind  exp1  exp2` expands to
```
%+  (bind mold)
  exp1
|=  mold
exp2
```

drbeefsupreme:
That's the same as the desugaring I got from `hoon.hoon`, minus the `^:`. I have the impression that there is some sort of difference since some rune entries have both an "expands to" section and a "desugaring" section like [miccol](https://urbit.org/docs/hoon/reference/rune/mic/#miccol). I've never bothered to inquire as to the difference between them until now. But the rune docs in general are very much non-standardized, so it could just be the result of neglect than anything important.

```hoon
        [%mcgl *]
      :^    %cnls
          :+  %cnhp
            q.gen
          [%ktcl p.gen]
        r.gen
      :+  %brts
        p.gen
      s.gen
```
I don't recall running into `^:` before this, but the [entry](https://urbit.org/docs/hoon/reference/rune/mic/#miccol) for it says that it doesn't actually produce a mold. So I'm not actually sure what to call it in this situation, other than a `^:`.

juped:
Most of the rune docs are outdated or otherwise false (and should be taken offline), including the one for ^: (irregular form ,), which in Hoon 140 just forces mold syntax. i.e., `%a` is a noun, `,%a` is a mold.

Pro tip: use open:ap to expand runes, i.e., `~(open ap (ream ';<  m  b  p  q'))`.

juped:
(Molds made with $ runes are now all crashmolds; molds used to fail softer by producing the bunt if the noun could not be correctly molded, and the outdated ^: doc up now is for a version of the rune whose function was turning a mold into a crashmold.)

PR #1045, opened 2021-03-28
Integrate thread docs

tinnus-napbus:
Adds userspace section and moves gall, landscape and threads under it.
The gall section under arvo now just has a short description and a link
to userspace/gall


baudtack (approved):
This is all pretty great!

PR #1050, opened 2021-03-31
remove sail/udon references

drbeefsupreme:
We don't use Sail/Udon anymore, so this tosses it into the memoryhole. This
means the Sail/Udon glossary page and the related mic runes.

Following #966


ohAitch:
Should it also be removed from the compiler, or is there a future for
serverside html generation / clientside wasm hoon?

tylershuster:
The login page is rendered in udon much to my chagrin￼

ohAitch:
And shouldn't be documented? ;)

(I'd add a quick summary in the eyre doc at least, which presumably was
written against the rune doc existing)

tylershuster:
No, I'm strongly against removing documentation of any extant systems. My chagrin is mostly over the *lack* of documentation. Try putting a json-encoded manifest in udon.

drbeefsupreme:
I have no strong opinion here. Sail/Udon was before my time, and I've only ever heard about how it isn't a thing anymore. Good to know that the Landscape login page uses it, that changes things somewhat. I was under the impression that it was used nowhere, and the fact that we still had anything for it in the compiler was just a matter of it being low priority.

Anybody in favor of removing it? If not I'll close this PR next week.

PR #1055, opened 2021-04-08
update clay data types

drbeefsupreme:
This brings `arvo/clay/data-types.md` up to date.

I think every entry was updated, though
some of them just had their code updated and the description was left unchanged.

A few entries were removed: `moot`, `cash`, `umph`, `upas`, and `mizu`. `mizu`
still exists in `lull.hoon` but it doesn't apppear to be used anywhere.

`$care` still needs work. It has 16 submodes and I've only been able to document
6 of them. As far as I've been able to tell, the rest are not documented by the
comments, and unraveling what they do from the code might end up being a major
chore so I'm hoping that @belisarius222 can shed light on these.

I only added a couple of new entries, so I'm sure there are more that I've
missed. For quick reference, here is the current list of entries in alphabetical order:

``` 
ankh arch blob care case dome cult desk dojo lobe miso moat mood nako raft rang
rant rave rede riff rind riot room rove rung soba tako unce nori update-state urge yaki
```

Anything sticking out as missing, unused, unimportant, etc?

I'm also a bit confused by `miso:clay`. This is for `ankh` deltas, so seems
pretty important. It has four options: `%del`, `%ins`, `%dif`, and `%mut`. Yet
it looks to be that `%dif` is unimplemented? `clay.hoon` 4040:4043:

```hoon
         %del  $(soba t.soba, deletes (~(put in deletes) p.i.soba)) 
         %ins  $(soba t.soba, changes (~(put by changes) [p p.q]:i.soba)) 
         %mut  $(soba t.soba, changes (~(put by changes) [p p.q]:i.soba)) 
         %dif  ~|(%dif-not-implemented !!) 
```

I have a hard time believing that Clay doesn't do diffs, so I must be
misunderstanding something. I think anyone reading the entry would be similarly
confused, so I want to address this before merging.

Lastly, I removed the public interface section at the end. It was very out of
date, and should be on a separate page.


arvo/clay/data-types.md
```diff
@@ -23,150 +24,180 @@ at least skimming, this so that you get a rough idea of how our state is
...
+                  new-cach=(map [=care =path] cach)     ::  new version
+              ==                                        ::
+              [%many track=? =moat lobes=(map path lobe)] ::  change range
+          ==                                            ::
 ```
 
-This represents a request for the state of the desk at a particular
-commit, specfied by `q`. `p` specifies what kind of information is
-desired, and `r` specifies the path we are requesting.
+Like a `rave` but with caches of current versions for `%next` and `%many`.
```
belisarius222, line 166: I wouldn't call it a cache, since I think we need that information for correctness.  "Cache" implies it could be deleted.

arvo/clay/data-types.md
```diff
@@ -23,150 +24,180 @@ at least skimming, this so that you get a rough idea of how our state is
...
 
 ```hoon
-    ++  moat  ,[p=case q=case r=path]                       ::  change range
+  +$  mood  [=care =case =path]                         ::  request in desk
 ```
 
-This represents a request for all changes between `p` and `q` on path
-`r`. You will be notified when a change is made to the node referenced
-by the path or to any of its children.
+This represents a request for the state of the `desk` at a particular
```
belisarius222, line 177: It's more generally a request for some data related to the state of the desk; not the state itself.

arvo/clay/data-types.md
```diff
@@ -23,150 +24,180 @@ at least skimming, this so that you get a rough idea of how our state is
...
 
-### `++care:clay`, clay submode
+### `$care:clay`, Clay submode
 
 ```hoon
-    ++  care  ?(%u %v %w %x %y %z)                          ::  clay submode
+  +$  care  ?(%a %b %c %d %e %f %p %r %s %t %u %v %w %x %y %z)  ::  clay submode
 ```
 
+TODO: The rest of these. Ask Ted if there is any docs on them because a lot of
```
belisarius222, line 197: I don't think there are any docs on them, but if there are some, they'd probably be in the Clay comments.

belisarius222: Btw, generally @philipcmonk knows more about Clay internals than I do, except for Ford, which I might know a bit more intimately.

belisarius222: A partial listing:
%a build hoon file at path
%b build mark by name (dynamically typed, i.e. a $dais datastructure)
%c build mark conversion gate by "from" and "to" mark names, dynamically typed (a $tube)
%d build mark, statically typed (vase of type-parameterized $nave core)
%e build mark conversion gate, statically typed
...
now from the end
...
%z hash of path contents (file, directory, or both)
%y directory listing
%x file contents

I don't recall the others offhand.

arvo/clay/data-types.md
```diff
@@ -186,28 +217,62 @@ if `q:ankh` is null), then this produces null.
...
 ```
 
-This is analogous to `++ankh:clay` except that the we have neither our
-contents nor the ankhs of our children. The other fields are exactly the
-same, so `p` is a hash of the associated ankh, `u.q`, if it exists, is a
-hash of the contents of this node, and the keys of `r` are the names of
-our children. `r` is a map to null rather than a set so that the
-ordering of the map will be equivalent to that of `r:ankh`, allowing
-efficient conversion.
+This is a single node in the filesystem. This may be file or a directory
```
belisarius222, line 230: I wouldn't describe this as a single node.  It includes all its subdirectories recursively -- see the `dir=(map @ta ankh)` type recursion.  `$arch` is indeed a single node, but `$ankh` is not.

belisarius222: I'd say an `$ankh` is a recursive type that can describe a whole tree of folders and files.

arvo/clay/data-types.md
```diff
@@ -216,91 +281,68 @@ that was at the head on date `p`, `%tas` refers to the commit labeled
...
+basically behaves exactly as you would expect.
 
 `lab` is a map of textual labels to numbered commits. Note that labels
 can only be applied to numbered commits. Labels must be unique across a
 desk.
 
-#### `++ankh`, filesystem node
+`mim` is a cache of the content in the directories that are mounted to Unix.
+Often, we convert to/from mime without anything really having changed; this lets
+us short-circuit that in some cases. Whenever you `%give` an `%ergo` `gift`, you
```
belisarius222, line 325: I don't really follow what you mean about %ergo.

arvo/clay/data-types.md
```diff
@@ -216,91 +281,68 @@ that was at the head on date `p`, `%tas` refers to the commit labeled
...
+must update this.
 
-```hoon
-    ++  ankh                                                ::  fs node (new)
-              $:  p=cash                                    ::  recursive hash
-                  q=(unit ,[p=cash q=*])                    ::  file
-                  r=(map ,@ta ankh)                         ::  folders
-              ==                                            ::
-```
+`fod` is the Ford cache, which keeps files, `mark`s, `cast`s, and `tube`s
```
belisarius222, line 328: I'd describe this in terms of build results rather than individual data structures.  The ford cache maintains a cache of the results of builds performed at this desk's current revision, including the full transitive closure of dependencies for each completed build.

arvo/clay/data-types.md
```diff
@@ -309,342 +351,308 @@ the desks we know about on their ship.
...
 
-#### `++mizu:clay`, merged state
+`%del` deletes the node.
 
-```hoon
-    ++  mizu  ,[p=@u q=(map ,@ud tako) r=rang]              ::  new state
-```
+`%ins` inserts a file given by `p`.
+
+`%dif` is currently unimplemented.
```
belisarius222, line 602: @philipcmonk is this right?

belisarius222 (changes requested):
Apologies for the slowness of this review.  I left some comments, but this is really only a partial review because I'm not that familiar with Clay's management of foreign desks and the new remote file retrieval system.

arvo/clay/data-types.md
philipcmonk: Apparently yes

arvo/clay/data-types.md
philipcmonk: Actually `%d` lists your desks, `%e` gives the nave, `%f` gives the statically typed mark conversion gate

`%u` now just checks whether the file exists, not what this page says.
`%p` produces the permissions for a directory
`%r` is like `%x` but produces the result as a vase instead of statically
`%s` has miscellaneous debug endpoints
`%t` produce list of descendant paths for a directory
`%v` get whole `dome` for a desk.  When used on a foreign desk this gets us "up-to-date" to the requested version.
`%w` get the version number and date of the requested path.  For example, `.^([@ud @da] %cw /=home`)` gives you the current version number and last revision date for the `%home` desk

arvo/clay/data-types.md
drbeefsupreme: I'm a bit confused as to what this really means. Marks implement diffs of files, but if diffs aren't implemented for ankhs (which I understand to be both, then what datatype has a notion of diff that marks implement? arch? Not sure that my question totally makes sense, but my initial impression of this (in the PR message) was that it somehow meant that Clay didn't do diffs at all, which definitely seemed wrong and must mean that my understanding of what ankhs are for is wrong.

drbeefsupreme: I guess this just means that there aren't diffs for changes in directory structure?

philipcmonk: `%dif` in a soba/miso is a way of specifying the new version of a file by just providing a diff on the old version, which would be implemented using `+pact` in the mark.  However, we still use `+diff` and `+pact` in marks for individual files for a few reasons.  One is that the object store can store blobs (files, as `+blob:clay`) directly or as a diff on another blob, which may be more efficient.  The other is that to perform a 3-way merge, you need to get the diff of each branch against the mergebase, and then when you merge those two diffs you get a 3rd diff, which you can `+pact` on the mergebase to get the new file.

> I guess this just means that there aren't diffs for changes in directory structure?

Right, we don't have a particular concept of changes in a directory structure.  We could define one in terms of additions and deletions, or we could encode renames as well.  Both 3-way merge and ted/diff implicitly use the former for now.  Git tries to derive renames heuristically.

arvo/clay/data-types.md
drbeefsupreme: I mostly just copied from this comment:

```
::  `mim` is a cache of the content in the directories that are mounted
::  to unix.  Often, we convert to/from mime without anything really
::  having changed; this lets us short-circuit that in some cases.
::  Whenever you give an `%ergo`, you must update this.
```

Looking around a bit more, I get the impression that `%ergo` moves automatically update `mim`? So this is more of a reminder to the coder to make sure `mim` gets updated when messing with `%ergo`'s logic than a user. But this is just based off of 5 minutes of perusing - I could be totally off.

arvo/clay/data-types.md
drbeefsupreme: How's this? I feel like this part basically needs its own separate tutorial since as its written its not apparent how to immediately put this stuff to use, but I think that's outside of the scope of this doc. Maybe something for @tinnus-napbus to look into

drbeefsupreme: https://github.com/urbit/docs/pull/1055/commits/c45d8c857a805b4895800f9dc76ea7ccc10cd4f1

arvo/clay/data-types.md
drbeefsupreme: https://github.com/urbit/docs/pull/1055/commits/4957fff63755abf87fd80b1e3ec77e79f54c0323

arvo/clay/data-types.md
drbeefsupreme: https://github.com/urbit/docs/pull/1055/commits/2e2d5757a653996d284afb35041d812710884af7

arvo/clay/data-types.md
tinnus-napbus: I've documented usage in a clay scry reference (see [https://github.com/tinnus-napbus/clay-docs/blob/main/scry.md](https://github.com/tinnus-napbus/clay-docs/blob/main/scry.md)) so these short summaries will be sufficient in the data types reference I think

arvo/clay/data-types.md
belisarius222: I'd just remove everything about `mim` except the sentence saying it's a cache of the Unix mount data.  The rest of that comment is pretty deep into the internals and I think will mostly seem like noise from the perspective of someone first learning about Clay.

arvo/clay/data-types.md
drbeefsupreme: Awesome, updated my descriptions a bit further. Will need to link to your page on clay scries once its done, but this makes me happy with the concise descriptions I have here.

arvo/clay/data-types.md
drbeefsupreme: https://github.com/urbit/docs/pull/1055/commits/3b8c67edb3c436a3be5a3405ed14283feba27e80

PR #1061, opened 2021-04-14
Revert "Support latest version of Zola"

drbeefsupreme:
Addresses https://github.com/urbit/urbit.org/issues/594

This needs to be addressed before this PR is merged again.

### DIFF
--- a/arvo/gall/overview.md
+++ b/arvo/gall/overview.md
@@ -2,6 +2,7 @@
 title = "Overview"
 weight = 1
 template = "doc.html"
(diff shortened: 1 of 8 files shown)


tylershuster:
Do we know definitely this is why search breaks? I have subsequently rolled out a change that will break the whole site if we go back to Zola 0.8.0 (any later is fine)

matildepark:
Search breaks because it can't find some prerequisite objects (elasticlunr and tippy) that were .js imports in the base.html template.

tylershuster:
Seems like adding those back would be easier than rolling back Zola

PR #1068, opened 2021-05-06
adds a+b -> [%a b] irregular form

drbeefsupreme:


drbeefsupreme:
Anything weird about this? I don't know whether its actually identical to `a/b`, one is deprecated, etc.

Fang-:
No this is fine. `a/b` is newer but has so far seen much less use than `a+b`. Listing them both should be fine for now, but we'll probably end up nuking one or the other eventually.

PR #1069, opened 2021-05-07
Update clay docs

tinnus-napbus:
adds docs on:
- scry reference
- reading and subscribing
- writing and modifying files
- managing mounts
- merging desks
- permissions
- foreign requests

also deletes the 3 old docs & updates the index and overview


tinnus-napbus:
btw @drbeefsupreme I had to do it all together in one PR cos of the links

drbeefsupreme (changes requested):
This is mostly only a pass for style. I don't know the details of what you're writing well enough to comment on accuracy - @philipcmonk will cover that.

Overall this is fantastic.

General comments (most are repeated in the suggestions, but these come up in multiple places):
 - Whenever you have a datatype that's used a few times, it should be an entry on `data-types.md` and link to it, rather than repeating the definition on multiple pages. Furthermore, any data types that are already on `data-types.md` should not have their definition repeated here - instead, you should link to the appropriate entry on `data-types.md`. For most documentation, I think the first mention of a term should be linked, and subsequent ones do not need to be. But for documentation where readers might not be reading the page from top to bottom, this might need to be adjusted. Use your best judgment.
 - There's a number of terms that ought to have backtics. I've suggested what to do for many of them, but since I can't find and replace as easily in a PR review a lot of them are still undone. Just follow the convention of the finished ones.
 - A lot of this is a mixture of tutorial/how-to/reference/explanation. We're trying to make all new docs of only one of those types. So for the docs about various `task`s, they need a bit of work. I would suggest re-arranging things so that there is a reference section that just has all `task`s and what they accept and return, along with the examples. But it shouldn't have the code for threads to run the examples on that page - that's inappropriate for a reference page. That should be its own how-to page that explains how to `pass` various `task`s to Clay and see the returns by hand. You already do this to an extent with the thread documentation you wrote in the past - so it might make sense to move that content there, or just link to it if its repeating any information.
 - In general we try to avoid duplicating information across the docs, to make it easier to find all the info about one topic in one spot, and so that when something changes you don't have to alter it in multiple places. There's a fair amount of duplicated information across this PR. This overlaps somewhat with the above points, but its worth mentioning on its own. The exception to this is Hoon school, which is supposed to be self-contained.

tinnus-napbus:
Per suggestions of @drbeefsupreme I have:

- restructed the tasks documents into two single documents - api reference & examples
- datastructures have been largely removed and instead linked to data types page
- brought style inline with recommendations in terms of backticking, vane capitalisation and such

arvo/clay/examples.md
```diff
@@ -0,0 +1,841 @@
...
+|%
+++  take-sign-loop
+  =/  m  (strand ,~)
+  ^-  form:m
+  %-  (main-loop:strandio ,~)
+  :~  |=  ~
+      ^-  form:m
+      ;<  res=(pair wire sign-arvo)
+        bind:m
+      ((handle:strandio ,(pair wire sign-arvo)) take-sign-arvo:strandio)
```
philipcmonk, line 254:
```suggestion
      ;<    res=(pair wire sign-arvo)
          bind:m
        ((handle:strandio ,(pair wire sign-arvo)) take-sign-arvo:strandio)
```

Should be indented like this so only the last child (the ~& and pure) are at the same level as the initial rune

arvo/clay/examples.md
```diff
@@ -0,0 +1,841 @@
...
+If we specify a non-existent mount point it will fail with an error printed to the dojo like:
+
+```
+> -send-task [%ogre %kids] 
+[%not-mounted %kids]
+```
+
+If we give it an unmounted `beam` it will not print an error but still won't work.
+
+## %dirk
```
philipcmonk, line 541: Maybe mention this corresponds to `|commit`

philipcmonk (approved):
Reviewed for content, this looks great.  Really like all the worked examples.

arvo/clay/examples.md
```diff
@@ -0,0 +1,841 @@
...
+   - [%warp](#warp) - Read a file on a foreign ship.
+   - [%merg](#merg) - Merge from a foreign `desk`.
+
+# Introduction
+
+This document contains a number of examples of interacting with Clay using its various `task`s. Sections correspond to the general details in the [API Reference](@/docs/arvo/clay/tasks.md) document.
+
+Most examples use one of these two threads which you can save to the `ted` directory of your `%home` `desk`:
+
+`send-task.hoon`
```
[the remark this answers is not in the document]
tinnus-napbus: thanks I've updated those ones to use `|pass`

arvo/clay/examples.md
tinnus-napbus: have added note about `|commit`

drbeefsupreme:
The changes look great - the only thing left I'd like to see is for `tasks.md`, for each `task` to have a subheadings that make it clear exactly what the structure of each `task` is, as well as the `gift` if it exists (or say it does not return a gift, but still have a heading for return).


```
### Accepts

[p=@ q=@]

### Returns

[ p=? q=@]
```

See e.g. https://urbit.org/docs/arvo/iris/iris-api/

drbeefsupreme:
Oh yeah I need to replace the images with ones hosted at urbit.org, I figured out why the upload was crashing, will make that change soon.

tinnus-napbus:
I've restructured the tasks doc with an accepts/returns structure as suggested

tinnus-napbus:
whoops forgot to move the task structure under the accepts section, have rectified

drbeefsupreme:
LGTM!

PR #1074, opened 2021-05-14
add page on life and rift

drbeefsupreme:
Addresses #980

Our description of life and rift has been incomplete and perhaps mildly misleading for some time. I've made a new page addressing this which I've put in the Azimuth section, though since these concepts also apply to moons and comets, I had also considered putting this page in the Ames section. It might make sense to rename the Azimuth section to Urbit ID to be more encompassing, but that can be a separate PR.

This is accompanied by changes to the using docs, which I'll put up a PR for in the urbit.org repo shortly.

### DIFF
--- a/azimuth/life-and-rift.md
+++ b/azimuth/life-and-rift.md
@@ -10,16 +10,17 @@ and _rift_. This numbering system partition messages according to the quantity
 of networking key changes and quantity of
 [breaches](@/using/id/guide-to-breaches.md), respectively. This is explained in
 more detail below. These values are utilized by [Ames](@/docs/arvo/ames/ames.md)
-and [Jael](@/docs/arvo/jael/jael-api.md) to ensure that you are communicating
-with a ship using its most recent set of keys.
+and [Jael](@/docs/arvo/jael/jael-api.md) to ensure that communication between
+ships is always done with the most recent set of networking keys, and that
+networking state is appropriately reset when a breach has occurred.
 
 Before it has been spawned, every ship begin with a `life` and `rift` of 0. For
 galaxies, stars, and planets, these values are stored in the Azimuth PKI, while
 for moons, these values are stored by their parent. Comets cannot change their
 networking keys, nor can they breach, and so their `life` and `rift` are always 0.
 
-You can check your current life and rift number by running the `+keys our`
-generator in dojo. You can inspect another ship's life and rift by running
+You can check your current `life` and `rift` number by running the `+keys our`
+generator in dojo. You can inspect another ship's `life` and `rift` by running
 `+keys ~sampel-palnet`.
 
 ## Life
--- a/azimuth/life-and-rift.md
+++ b/azimuth/life-and-rift.md
@@ -14,7 +14,7 @@ and [Jael](@/docs/arvo/jael/jael-api.md) to ensure that communication between
 ships is always done with the most recent set of networking keys, and that
 networking state is appropriately reset when a breach has occurred.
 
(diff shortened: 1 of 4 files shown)


drbeefsupreme:
Forgot to mention that this also removes mention of life and rift from the glossary. These are technical terms that most people shouldn't ever need to learn, so I don't think they're appropriate there.

azimuth/life-and-rift.md
```diff
@@ -0,0 +1,64 @@
...
+
+Thus under ordinary circumstances, a breach will increment both `life` and
+`rift` since a breach typically also involves changing the networking keys.
+However, it is possible to breach without changing the networking keys. As a
+breach resets networking keys to 0 unless new keys were specified, if the
+networking keys were already 0 then they have not changed, and so `rift` will
+increment but `life` will not. This is just a special case of breaching while
+maintaining the same networking keys. That is to say, if the new keys specified
+as part of a breach as the same as the old, again `rift` will increment while
+`life` will not.
```
Fang-, line 64: Again difficult/confusing and partially incorrect.

```suggestion
Thus under ordinary circumstances, a breach will increment both `life` and
`rift` since a breach typically also involves changing the networking keys.
However, it is possible to breach without changing the networking keys.
If the new keys specified
as part of a breach are the same as the old, `rift` will increment while
`life` will not.
```

I'd also call into question the degree to which these details about when the numbers increment are actually relevant to most users reading this doc. The common case is easy to reason about, the uncommon/weird case the user needs to intentionally trigger, and even then the behavior probably doesn't matter from their perspective?

glossary/breach.md
```diff
@@ -6,27 +6,22 @@ category = "arvo"
 +++
 
 Continuity on the [Ames](../ames) network occasionally needs to be broken at
-this early stage in order to correct a networking error or make major changes to
-Arvo. These infrequent events are known as
-**breaches** which either cause an individual ship to forget its network message
-history, called a **personal breach**, or cause the entire network to forget its
-history, called a **network breach**.
+this early stage in order to correct a networking error or to make major changes
+to Arvo. These infrequent events are known as **breaches** which either cause an
```
Fang-, line 10: This opening paragraph seems to be talking mainly about network breaches, but the paragraph on that says they won't happen anymore. Maybe shift focus to personal breaches right at the start, making network breaches more of a footnote to this section?

We could consider associating "breach" with personal breaches by default, reserving "network breach" for the historical case.

Fang- (commented):
Some inaccuracies and overcomplication in here.

baudtack (changes requested):
Seems good with suggested changes

azimuth/life-and-rift.md
drbeefsupreme: The reader I have in mind here is someone attempting to write code to interact with the Azimuth smart contracts or L2 stuff (like setting up atomic swaps), in which case the crunchy details and edge cases matter. Previously our docs on life/rift were aimed at general users but I don't think they need to know about those concepts at all.

glossary/breach.md
drbeefsupreme: https://github.com/urbit/docs/pull/1074/commits/7cf73850873edbb71ac171d3333aceb05478cb27

Fang-: I'd argue the commit is insufficient. The (local) network error case is fair game, but we've never asked/required people to breach to apply an upgrade, and "occasionally" currently feels more like "with increasing rarity".

PR #1077, opened 2021-05-21
revise comet glossary

drbeefsupreme:
The old glossary entry for comets was lacking, and didn't really reflect our
current position of encouraging the use of comets for people new to Urbit. This
revises it to reflect out position, spell out exactly what capabilities are lacking in comets, and
removes the part saying that they "will not be initially trusted by others",
which is too broad of a statement.


glossary/comet.md
```diff
@@ -6,9 +6,20 @@ template = "doc.html"
 category = "arvo"
 +++
 
-A **comet** is a kind of [ship](../ship) on the [Arvo](../arvo) network. They are disposable identities that will not be initially trusted by others on the network. One use might be to run bots as comets.
+A **comet** is a kind of [ship](../ship) on the [Arvo](../arvo) network. They
+are free identities for which we encourage new users to boot to try things out
```
baudtack, line 10: I strongly object to referring to comets this way. They are only "free" because they are not on the azimuth and they have many uses beyond just testing if you like urbit enough to buy a planet. They aren't intended to be a trial mode or anything like that.

glossary/comet.md
```diff
@@ -6,9 +6,20 @@ template = "doc.html"
...
+before deciding to purchase a [planet](..planet).
 
-Comets have no presence on [Azimuth](../azimuth) and thus have no associated Urbit identities.
+Comets have nearly all the same capabilities as a planet. The exceptions are:
+ - They may not spawn [moons](../moon).
+ - They may not [change their networking keys](../bridge).
```
baudtack, line 15: "can not" they aren't prevented from doing this by some authority, which this sounds like they are. Should explicitly mention here that it's because they are not on chain so there is no way to broadcast a trustable change in keys.

baudtack (changes requested):

drbeefsupreme:
@baudtack How's it look now?

baudtack (approved):

PR #1070, opened 2021-05-07
summary of cryptography in system overview

drbeefsupreme:
I've been working on documenting our cryptosystems. This was all on one page
found here
https://github.com/urbit/docs/blob/encryption/arvo/reference/cryptography.md but
I've split off the summary into its own page and filed it under System Overview, and over the next week or two will
be finishing up and moving the rest of it into appropriate sections.

I am 95% confident about the accuracy of this content, but some extra eyes on how
moon and comet keys are handled would be appreciated.

PR #1091, opened 2021-06-09
ames encryption

drbeefsupreme:
Followup to #1070

This PR contains 3 additional pages on cryptography in Urbit, focusing on
aspects relevant to Ames. The pages are specifically for Ames, zuse, and Vere.
The zuse page only discusses the cores directly relevant to Ames - so it does
not yet contain anything relevant to Ethereum like keccak. Same goes the Vere
page.

Vere is the part of the system I understand the least, so its pretty bare bones. I
say that all cryptography-related jets are in `/jets/e`, but this is really just a guess.

The next PR in this series will be for Azimuth/Ethereum related functions.

I've tagged @Fang- and @belisarius222 for review since I know you both
have worked on the relevant code here. Not sure who else may have.


system-overview/cryptography.md
```diff
@@ -114,3 +114,11 @@ infrastructure utilized by Urbit. `azimuth-tracker` obtains networking public
...
 Jael and utilized by Ames for end-to-end encrypted communication.
 
+### Additional documentation
+
+The following pages contained more detailed information about the cryptography
+utilized by each of the system components.
+
+ - [Ames](@/docs/arvo/ames/cryptography.md)
+ - [zuse](@/docs/arvo/reference/cryptography.md)
+ - [Vere](@/docs/vere/cryptography.md)
```
Fang-, line 124:
```suggestion
 - [Ames](@/docs/arvo/ames/cryptography.md)
 - [Zuse](@/docs/arvo/reference/cryptography.md)
 - [Vere](@/docs/vere/cryptography.md)
```

arvo/ames/cryptography.md
```diff
@@ -0,0 +1,146 @@
...
+The encrypted payload of each packet is a `$shut-packet`, which is the `+jam` of
+a cell with the `bone`, message number, and message fragment or ack (see
+[Ames](@/docs/ames/ames.md) for more information on packet structure). The
+message fragment is signed using the authentication key. It is encrypted using
+[`+en:crub:crypto`](@/docs/arvo/reference/cryptography.md#en) found in
+`sys/zuse.hoon`, which utilizes the 256-bit AES-SIV algorithm.
+
+## Diffie-Hellman key exchange {#key-exchange}
+
+For each foreign ship a given ship has met, `$ames-state` contains a
```
Fang-, line 45:
```suggestion
For each foreign ship a given ship has communicated with, `$ames-state` contains a
```

arvo/ames/cryptography.md
```diff
@@ -0,0 +1,146 @@
...
+    ++  nu  ^?                                          ::  reconstructors
+      |%  ++  pit  |~([a=@ b=@] ^?(..nu))               ::  from [width seed]
+          ++  nol  |~(a=ring ^?(..nu))                  ::  from ring
+          ++  com  |~(a=pass ^?(..nu))                  ::  from pass
+      --  ::nu                                          ::
+    --  ::acru                                          ::
+```
+
+As the `+acru` core is merely an interface, the details on how it is utilized
+may vary according to the cryptosuite implemented in it. We summarize what each
```
Fang-, line 113: Isn't the point of the interface that you can _utilize_ it in a consistent way, but that underlying _implementation_ may vary?

arvo/ames/cryptography.md
```diff
@@ -0,0 +1,146 @@
...
+how the specific cryptosuite utilized by Ames is implemented.
+
+#### `+as:acru`
+
+This core is used for the standard asymmetric cryptographic operations: encrypting
+(`+seal`), signing (`+sign`), authenticating (`+sure`), and decrypting (`+tear`).
+
+#### `+ex:acru`
+
+This core stores keys and their fingerprints. `+sec` is the secret key (which
```
Fang-, line 125:
```suggestion
This core is used to extract keys and their fingerprints. `+sec` is the secret key (which
```

Or something along these lines.

arvo/reference/cryptography.md
```diff
@@ -0,0 +1,222 @@
...
+Documentation for [Insecure Hashing](@/docs/hoon/reference/stdlib/2e.md) and the
+[SHA Hash Family](@/docs/reference/stdlib/3d.md) is found in the Hoon standard
+library reference.
+
+## Summary
+
+`zuse` contains several cryptosuites. The ones utilized by Ames are
+[`+ed:crypto`](#ed), [`+aes:crypto`](#aes), and [`+crub:crypto`](#crub), with
+the latter being the only one which is implementable as an [`+acru:ames`
+core](@/docs/arvo/ames/cryptography/#crypto-core).
```
Fang-, line 22:
```suggestion
`zuse` contains several cryptosuites. The ones utilized by Ames are
[`+ed:crypto`](#ed), [`+aes:crypto`](#aes), and [`+crub:crypto`](#crub), with
the latter being the only one which is implemented as an [`+acru:ames`](@/docs/arvo/ames/cryptography/#crypto-core)-shaped core.
```

arvo/reference/cryptography.md
```diff
@@ -0,0 +1,222 @@
...
+A `+crub:crypto` core's payload contains public encryption and authentication
+keys and optional secret encryption and authentication keys.
+```hoon
+  ++  crub  !:
+    ^-  acru
+    =|  [pub=[cry=@ sgn=@] sek=(unit [cry=@ sgn=@])]
+    |%
+    ...
+```
+`+crub` cores are typically created using one of the constructors in [`+nu:crub`](#nu).
```
Fang-, line 45:
```suggestion
`+crub` cores (because they follow the `+acru` interface) are typically created using one of the constructors in [`+nu:crub`](#nu).
```

Fang-: You could consider scrapping the explanations here entirely in favor of referring to the `+acru` docs. (And/or moving the explanations from here, over there.)

Fang- (commented):
Some humble clarifications and suggestions, but looks fine otherwise. I have no authority here though!

Fang-:
> I also put the `zuse` documentation under `/arvo/reference`, but maybe it should go in `/hoon/reference`. I've seen people say `zuse` to refer to the Hoon standard library in `hoon.hoon`, but is that right? Should `zuse` only refer to the contents of `zuse.hoon`? And what about `lull.hoon`, is it right to just say `zuse` when you mean `lull.hoon`...?

If I'm not mistaken, the hierarchy/dependency chain here is `hoon.hoon` -> `lull.hoon` -> `arvo.hoon` -> `zuse.hoon` -> rest of kernelspace. So, zuse isn't really in arvo proper (`arvo.hoon`), but it accessible to the vanes. Calling the things in `hoon.hoon` or `lull.hoon` "zuse" is certainly wrong.

arvo/ames/cryptography.md
drbeefsupreme: Right, I think what I had in mind was that you'd utilize different cryptosuites in different situations. Or maybe I just flubbed. Either way, you're right it is less confusing to say implemented here. https://github.com/urbit/docs/pull/1091/commits/fefe4c1695f63a65f79e7ac0a3946ff57d98553c

drbeefsupreme:
Thanks, that makes sense. I think I made a mistake with #1070 then calling `zuse` "the Hoon standard library" so I changed it to "part of the standard library". https://github.com/urbit/docs/pull/1091/commits/871d21871b03f0ea5f5a7210fd682f1b539bf110

There might be other parts of the docs that have this confusion, I'll need to go look, but that can be a separate PR.

arvo/reference/cryptography.md
drbeefsupreme: Since `+crub` lives in `zuse.hoon` I don't want to move it to the Ames section. I generally think it is good to avoid duplicating information in reference sections, but I'm not sure if it should "count" if one section is describing an interface and the other is describing an implementation of that interface. Though it becomes even more of an edge case when there is only one implementation of an interface that is ever used, like it is here. I'm inclined to leave it as-is, but I'm going to check out how other projects split up documentation for interfaces/implementations and see if its done similarly or otherwise.

Issue #944, opened 2020-09-08
+roll and +reel


drbeefsupreme:
Docs to accommodate whatever this ends up turning into

https://github.com/urbit/urbit/issues/3430

PR #1099, opened 2021-06-11
roll/reel revision and clarification

drbeefsupreme:
#944 and https://github.com/urbit/urbit/issues/3430

### DIFF
--- a/hoon/reference/stdlib/2b.md
+++ b/hoon/reference/stdlib/2b.md
@@ -564,7 +564,7 @@ the final value of the accumulator.
 (To "slam" means to call a gate and give it a sample/samples. In this instance,
 `a` is the list of samples that are given to the gate b.)
 
-The initial value of the accumulator is the bunt of `b`'s second argument. This can occasionally produce
+The initial value of the accumulator is the bunt of `b`'s second argument (`+<+`). This can occasionally produce
 undesired behavior (see examples). If you need more control over the initial
 value, try making use of `$_` and `|:`, or perhaps [`+spin`](#spin) or
 [`+spun`](#spun).
@@ -621,7 +621,7 @@ since that is the bunt of `@`:
 ```
 We can fix this with `|:`:
 ```
-> (reel `(list @)`~[1 2 3 4] |:([a=`@`1 b=`@`1] (mul a b)))
+> (reel `(list @)`~[1 2 3 4] |:([a=1 b=1] (mul a b)))
 24
 ```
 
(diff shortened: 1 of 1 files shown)


hoon/reference/stdlib/2b.md
```diff
@@ -564,6 +564,11 @@ the final value of the accumulator.
 (To "slam" means to call a gate and give it a sample/samples. In this instance,
 `a` is the list of samples that are given to the gate b.)
 
+The initial value of the accumulator is `+>+.$`. This can occasionally produce
```
Fang-, line 567:
```suggestion
The initial value of the accumulator is the bunt of `b`'s second argument. This can occasionally produce
```

Something like this would be much more clear. For the lark, you would've wanted `+<+`, in any case.

hoon/reference/stdlib/2b.md
```diff
@@ -564,6 +564,11 @@ the final value of the accumulator.
...
+unexpected behavior (see examples). If you need more control over the initial
```
Fang-, line 568:
```suggestion
undesired behavior (see examples). If you need more control over the initial
```

It's not really unexpected if you've read these docs and/or know what's going on. And it's the _undesired_ behavior you always want to affect, not necessarily the unexpected. ...But maybe that's too pedantic.

hoon/reference/stdlib/2b.md
```diff
@@ -600,6 +605,27 @@ The accumulator, which is a noun.
...
+```
+However, if you build a gate that uses `+mul` like so, the sample defaults to 0
+since that is the bunt of `@`:
+```
+> (reel `(list @)`~[1 2 3 4] |=([a=@ b=@] (mul a b)))
+0
+```
+We can fix this with `|:`:
+```
+> (reel `(list @)`~[1 2 3 4] |:([a=`@`1 b=`@`1] (mul a b)))
```
Fang-, line 624: Oh so `|:` is just a more explicit `|=  a=_1`? That's kinda neat. The `` `@` `` casts make this feel a little bit clunky, I think for this example it might be fine to not care about these technically becoming `@ud`. But your call!

hoon/reference/stdlib/2b.md
```diff
@@ -613,6 +639,11 @@ producing the final value of the accumulator.
 (To "slam" means to call a gate and give it a sample/samples. In this instance,
 `a` is the list of samples that are given to the gate b.)
 
+The initial value of the accumulator is `+>+.$`. This can occasionally produce
```
Fang-, line 642: As above, so below.

hoon/reference/stdlib/2b.md
drbeefsupreme: Good call

Issue #1064, opened 2021-04-26
Weird characters on irregular forms


tylershuster:
<img width="943" alt="image" src="https://user-images.githubusercontent.com/3999320/116163339-6e4c1e00-a6ac-11eb-915b-1e78bd82a8f5.png">

Some sort of botched HTML

PR #1100, opened 2021-06-16
irregular.md: add ^* and ^:, adds spacing, more verbose

drbeefsupreme:
This addresses #1016 and #1064

I replaced the terse "I:" and "R:" with "Regular:" and "Irregular:", and put the
entries onto separate lines.

I also added a couple missing entries, for `^*` and `^:`. I didn't even realize
that bunting with `*` was an irregular form since `^*` has never been on this
page. These could be separate PR's but they're all so minor I decided to do them together.

Also, I am under the impression that `^:` is no longer relevant since all molds
crash when you feed them a value of the wrong type. Nonetheless, it still
appears in the codebase, so we ought to have it documented while it lingers. I
can't seem to find the conversation I remember having about this though, so
please correct me if I am mistaken. If so, I should go edit the `ket.md` page to
say as much.

### DIFF
--- a/hoon/reference/irregular.md
+++ b/hoon/reference/irregular.md
@@ -12,48 +12,52 @@ docs and a short description of the rune. Both regular and irregular forms are g
 
 Want to `Ctrl-f` to find out the meaning of something weird you saw? Search for `\symbol`. ie `\?` or `\=`. It'll show you to the irregular forms that uses that symbol.
 
-## . dot (nock)
+## `.` dot (nock)
 
 Anything Nock can do, Hoon can do also.
 
-### .+ dotlus
+### `.+` dotlus
 
 [docs](@/docs/hoon/reference/rune/dot.md#dotlus) \\+
+
 `[%dtls p=atom]`: increment an atom with Nock 4.
 
 Regular: `.+(p)`
 
 Irregular: `+(p)`
 
-### .= dottis
+### `.=` dottis
 
 [docs](@/docs/hoon/reference/rune/dot.md#dottis) \\=
+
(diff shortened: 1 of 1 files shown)


hoon/reference/irregular.md
```diff
@@ -151,44 +162,67 @@ Hoon has the usual branches and logical tests.
...
+
+Irregular: `|(p)`
 
 ## ^ ket (cast)
 
 Lets us adjust types without violating type constraints.
 
+### ^: ketcol
+
+[docs](@/docs/hoon/reference/ket.md#ketcol) \\,
```
tylershuster, line 193: These lines that start with `[docs]` should have another space after them to avoid inlining with the next line.

tylershuster (changes requested):
Looking better, but still some funky formatting regarding #1064 

Now that I understand the problem a little better, perhaps this line "Want to Ctrl-f to find out the meaning of something weird you saw? Search for \symbol. ie \? or \=. It'll show you to the irregular forms that uses that symbol." is adding to the confusion. 

Why this shortcut? That is, why introduce the "backslash + rune" pattern? It seems to just add confusion.

And while you're in there, some red formatting a la https://urbit.org/docs/hoon/reference/rune/dot/#dotlus would be good for consistency's sake.

drbeefsupreme:
The backslash pattern, or something like it, is needed because it allows the Ctrl-F result to be unique, rather than giving every col (for instance) on the page.

hoon/reference/irregular.md
drbeefsupreme: Good catch

PR #1101, opened 2021-06-16
adds page describing filesystem hierarchy

drbeefsupreme:
Addresses #855

I give a brief description of what each folder is used for, and go into more
detail for `/sys`.

This page might be trying to be too many different things at once. It might
better belong in the `arvo/concepts` folder instead of `arvo/reference`. For
instance, I mention what dojo syntax to use to run generators and threads, as
well as Ford runes. I also describe how the subject is built up from the files
in `/sys`, which while interesting, might be more appropriately tacked onto
`/arvo/overview.md` rather than here.

@Fang- said the dependencies for the files in `/sys` is `hoon->lull->arvo->zuse`
but peeking around inside it looked like it was `hoon->arvo->lull->zuse`
https://github.com/urbit/docs/pull/1091#issuecomment-857961935 not sure if he
made a typo or I'm misunderstanding something.

### DIFF
--- a/arvo/reference/filesystem.md
+++ b/arvo/reference/filesystem.md
@@ -1,0 +1,71 @@
++++
+title = "Filesystem Hierarchy"
+weight = 20
+template = "doc.html"
++++
+
+Here we describe what each of the primary folders present in the `%home` `desk`
+are used for. This organization is merely a convention, and the exact location
+of any file does not affect its operation. That being said, some [Ford
+runes](@/docs/arvo/ford/ford.md) are designed with this structure in mind, and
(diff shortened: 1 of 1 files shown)


drbeefsupreme:
This also needs to wait for #1091 to merge since it links to it.

Fang-:
> @Fang- said the dependencies for the files in /sys is hoon->lull->arvo->zuse
but peeking around inside it looked like it was hoon->arvo->lull->zuse

Yeah you're right, I had misremembered. arvo->lull is correct.

baudtack (approved):

PR #1106, opened 2021-07-19
update jael docs

tinnus-napbus:
- update api ref
- add scry ref
- add data types
- add (brief) overview (maybe should expand?)
- add examples

@drbeefsupreme


arvo/jael/examples.md
```diff
@@ -0,0 +1,311 @@
...
+
+As you can see the `set` is now empty, so we know the `%nuke` succeeded and Jael will no longer send us pubkey updates for `~dopzod`. One thing to note about `%nuke` is that it *must* come from the same `duct` as the original subscription. You can't unsubscribe another app, ship, thread or what have you, so if we'd tried `%nuke`ing the subscription from a separate thread it wouldn't have worked.
+
+# `%rekey`
+
+Here we'll look at changing a ship's private keys by sending a `%rekey` `task` to Jael. 
+
+Here's a thread that will scry for a ship's current `life`, increment it, generate a random private key from some entropy, and send a `%rekey` `task` to jael to increase the ship's `life` and set the new private key.
+
+This example was run on an offline **copy** of a real ship booted with `urbit -L` because you can't change the `life` of a `urbit -F` fake ship. Setting a random private key this way on a real ship is obviously a bad idea so please be careful not to do that.
```
drbeefsupreme, line 188:
```suggestion
**This example was run on an offline _copy_ of a real ship booted with `urbit -L` because you can't change the `life` of a `urbit -F` fake ship. Setting a random private key this way on a real ship is obviously a bad idea so please be careful not to do that.**
```
This is dangerous enough that I think the whole paragraph should be bolded. I'd actually consider removing this example entirely. I'm not sure the educational value is worth the risk of someone needing to breach if they forget to use `-L`. Not sure if a breach would actually be needed, but either way it seems too easy to end up in a bad spot due to carelessness or a typo, for not much gain.

arvo/jael/scry.md
```diff
@@ -0,0 +1,222 @@
...
+  lyf=1
+    key
+  1.729.646.917.183.337.262.068.568.133.450.460.269.618.308.934.734.494.661.340.450.478.360.301.077.532.415.587.141.242.844.893.269.500.211.196.916.184.768.225.437.577.083.064.475.012.067.992.317.942.521.494.338
+  sig=~
+]
+```
+
+# `%deed`
+
+A `%deed` scry gets the `life`, pubkey and maybe a signature if the ship in question is a comet. It takes the target ship and life as its `path` like `/~sampel-palnet/1`and the type returned is a `[life pass (unit @ux)]`.
```
drbeefsupreme, line 127:
```suggestion
A `%deed` scry gets the `life`, pubkey and maybe a signature if the ship in question is a comet. It takes the target ship and life as its `path` like `/~sampel-palnet/1` and the type returned is a `[life pass (unit @ux)]`.
```

arvo/jael/scry.md
```diff
@@ -0,0 +1,222 @@
...
+## Example
+
+```
+> .^((list @p) %j /=saxo=/~sampel-palnet)
+~[~sampel-palnet ~talpur ~pur]
+```
+
+# `%subscriptions`
+
+A `%subscriptions` scry gets the current state of subscriptions to public key updates (typically initiated with a [%public-keys task](@/docs/arvo/jael/tasks.md#public-keys)). It takes a `life` as its `path` and returns a triple of the following:
```
drbeefsupreme, line 179:
```suggestion
A `%subscriptions` scry gets the current state of subscriptions to public key updates (typically initiated with a [%public-keys task](@/docs/arvo/jael/tasks.md#public-keys)). It takes a `life` as its `path` and returns a triple consisting of the following:
```

arvo/jael/tasks.md
```diff
@@ -0,0 +1,242 @@
...
+
+Jael `%give`s a `%init` `gift` to Unix.
+
+# `%listen`
+
+```hoon
+[%listen whos=(set ship) =source]  ::  set ethereum source
+```
+
+Set ethereum source.
```
drbeefsupreme, line 80:
```suggestion
Set Ethereum source.
```

drbeefsupreme (changes requested):
Looks great, only some minor suggestions

arvo/jael/data-types.md
```diff
@@ -0,0 +1,248 @@
++++
+title = "Data Types"
+weight = 4
+template = "doc.html"
++++
+
+Jael's type definitions in `lull.hoon` have a three substructures as well as the general types, so this reference is divided as follows:
```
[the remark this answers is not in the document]
tinnus-napbus: fair point I have reworded it

tinnus-napbus:
cool I've fixed the issues @drbeefsupreme raised

drbeefsupreme (approved):

drbeefsupreme:
We should get someone familiar with the technical details of Jael to review this as well. @Fang- ?

arvo/jael/tasks.md
```diff
@@ -0,0 +1,238 @@
...
+
+# `%public-keys`
+
+```hoon
+[%public-keys ships=(set ship)]
+```
+
+Subscribe to public key (and related) updates from Jael.
+
+An agent or thread can subscribe to be notified of public key updates, sponsorship changes and continuity breaches for the `set` of `ship`s specified in the `ships` field. The subscription will continue until Jael receives a [%nuke](#nuke) `task` to cancel it.
```
Fang-, line 166: > An agent or thread can ...

Not sure why these are mentioned explicitly. Technically any part of the system that can emit moves to arvo can do this. Ames uses this, for example.

arvo/jael/tasks.md
```diff
@@ -0,0 +1,238 @@
...
+
+# `%rekey`
+
+```hoon
+[%rekey =life =ring]
+```
+
+Update private keys.
+
+This is what is send to Eyre by `%hood` implicitly when you run `|rekey` after rekeying via Bridge (or via `|cycle-moon-keys` on your planet in the case of a moon). It will update your `life` (key revision number) and private keys. The `life` field is the new `life` (typically an increment of the current `life`) and the `ring` is a private key `@`.
```
Fang-, line 192:
```suggestion
This is what is sent to Jael by `%hood` when you run `|rekey`, as you must after setting new Azimuth keys. (In the case of a moon, `|cycle-moon-keys` sends this for you). It will update your `life` (key revision number) and private keys. The `life` field is the new `life` (typically an increment of the current `life`) and the `ring` is a private key `@`.
```

Or some similar change that's less awkwardly worded.

arvo/jael/tasks.md
```diff
@@ -0,0 +1,238 @@
...
+
+## Returns
+
+Eyre does not return any `gift` in response to a `%rekey` `task`.
```
Fang-, line 196:
```suggestion
Jael does not return any `gift` in response to a `%rekey` `task`.
```

There's a couple more instances, you'll want to ctrl-f for those.

arvo/jael/scry.md
```diff
@@ -0,0 +1,222 @@
++++
+title = "Scry Reference"
+weight = 3
+template = "doc.html"
++++
+
+Jael's scry endpoints never take a `care`. The particular endpoints are specified in the place of the desk in the path prefix, then additional arguments are specified in the path.
```
Fang-, line 7: Wondering if we should call this out as being considered old-style scry paths, considering all recent additions (in ames, dill, eyre) do use cares and expect an empty desk.

arvo/jael/scry.md
```diff
@@ -0,0 +1,222 @@
...
+## Example
+
+```
+> .^(@ud %j /=step=/(scot %p our))
+0
+```
+
+# `%code`
+
+A `%code` scry gets the current web login code. It takes your ship name as its `path` and the type returned is a `@p`.
```
Fang-, line 40: Looks like it doesn't actually care what ship gets passed in the path, just that there's one (any one) there. `/=code=/~zod` works just fine on my planet.

Tempted to just not comment on that here. (^:

arvo/jael/scry.md
```diff
@@ -0,0 +1,222 @@
...
+## Example
+
+```
+> .^(@p %j /=code=/(scot %p our))
+~lidlut-tabwed-pillex-ridrup
+```
+
+# `%life`
+
+A `%life` scry gets the current `life` (key revision number). It takes a ship as its `path` and the type returned is a `@ud`.
```
Fang-, line 51:
```suggestion
A `%life` scry gets the current `life` (key revision number) of a ship. It takes a ship as its `path` and the type returned is a `@ud`.
```

Same for `lyfe` below.

Fang-: Also, to distinguish this from `lyfe`, you might want to mention that this fails to produce a result if we do not know it (ie crash) whereas `lyfe` will always return a `unit`, but it might be empty if we don't know the answer.

Fang-: The above two comments apply to `rift` and `ryft` as well.

arvo/jael/scry.md
```diff
@@ -0,0 +1,222 @@
...
+> .^([life pass (unit @ux)] %j /=deed=/(scot %p our)/1)
+[ 1
+  2.649.818.598.466.464.524.534.996.841.661.372.043.001.125.438.577.869.575.450.096.229.472.239.369.342.882.589.479.483.389.735.480.760.982.635.242.955.515.616.049.650.771.700.602.823.946.406.713.457.849.302.626
+  ~
+]
+```
+
+# `%earl`
+
+A `%earl` scry deterministically derives a private key for a moon. Note the `|moon` generator doesn't use this and instead generates moon private keys non-deterministically. It takes the moon name and your ship's current `life` (not the moon's `life`) as its `path` like `/~doznec-dozzod-dozzod-dozzod/1`. The type returned is a [$seed](@/docs/arvo/jael/data-types.md#seed), the `life` of the moon in the `lyf` field of the `seed` will always be `1`.
```
Fang-, line 141: So is this just holdover from previous-generation moon infrastructure? cc @philipcmonk

arvo/jael/scry.md
```diff
@@ -0,0 +1,222 @@
...
+## Example
+
+```
+> .^(@p %j /=sein=/~sampel-palnet)
+~talpur
+```
+
+# `%saxo`
+
+A `%saxo` scry gets the sponsorship chain for the target ship (including the target ship itself). For example, a `%saxo` scry for a planet will return the planet, its star and the star's galaxy. This scry is used implicitly by `sein:title`. It takes the target ship as its `path` and the type returned is a `(list @p)`.
```
Fang-, line 168:
```suggestion
A `%saxo` scry gets the sponsorship chain for the target ship (including the target ship itself). For example, a `%saxo` scry for a planet will return the planet, its star and the star's galaxy. This scry is used implicitly by `saxo:title`. It takes the target ship as its `path` and the type returned is a `(list @p)`.
```

Fang- (changes requested):
Nice work as always! Some change requests and comments below.

arvo/jael/scry.md
philipcmonk: Yes

arvo/jael/tasks.md
tinnus-napbus: dammit I did this so many times, thought I'd caught them all

matildepark:

PR #1110, opened 2021-07-26
update iris docs

tinnus-napbus:
remove old iris docs and add:

- brief overview
- data types ref
- api ref
- example

@drbeefsupreme


matildepark:

